PR #18, opened 2019-12-26
Add changelog and Releasing procedure

stephanenicolas:
Hi @autonomousapps ,

here is a changelog, I tried my best to document the project from the release tab and the commit tab of GH. It's not ideal but at least we can get an idea of the project's progress.
For users, a well maintained changelog is really valuable, and for fellow contributors too.

Also, I started a release procedure document to make the project more industrial.
Can you please merge and finalize these 2 files ? 

Thx in advance,
 S.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,12 @@
+Release procedure for dependency-analysis-android-gradle-plugin
+
+1. Update the changelog
+1. Bump version number to next stable version (use semantic versioning: x.y.z)
+1. Update README if needed
+1. git commit -am "Prepare for release x.y.z." 
+1. Release on ...
+1. git tag -a x.y.z -m "Version x.y.z"
+1. Update build.kts to next snapshot version
+1. git commit -am "Prepare next development version."
+1. git push && git push --tags
+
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -0,0 +1,12 @@
+Release procedure for dependency-analysis-android-gradle-plugin
+
+1. Update the changelog
+1. Bump version number to next stable version (use semantic versioning: x.y.z)
+1. Update README if needed
+1. git commit -am "Prepare for release x.y.z." 
+1. Release on ...
+1. git tag -a x.y.z -m "Version x.y.z"
```
autonomousapps: Small question. Are you saying tags should always be `x.y.z`? I think I've been using `vx.y.z`.

RELEASING.md
```diff
@@ -0,0 +1,12 @@
+Release procedure for dependency-analysis-android-gradle-plugin
```
autonomousapps: I think this also needs to include "run the tests and ensure they pass." I will probably add a `dependsOn` to the release task that requires the tests to be green.

autonomousapps:
Updated locally and merged with commit d3e1857e602321ccf5447f7e2b12f02ce61785bd

PR #20, opened 2019-12-27
Add a small test DSL.

stephanenicolas:
My goal was to make it easier to have actual files in the tests themselves, so that it would be easier to define more diverse tests.
I introduced base project classes, without breaking the existing hierarchy of modules.

I searched in the gradle code if there were a few classes that we could use to have a real DSL to create projects. I didn't find one but a few things are pretty close, we should ask Luke when we meet him.
I am looking for something like : 
```kotlin
  whithBuildFile {
    plugins {
      id('java-library')
      id('org.jetbrains.kotlin.jvm')
    }
    dependencies {
      implementation platform('org.jetbrains.kotlin:kotlin-bom')
      implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
      api 'org.apache.commons:commons-math3:3.6.1'
      implementation 'com.google.guava:guava:28.0-jre'
    }  
  }
```
It would be even better if we could use the real DSL objects and mimic the fact that we write an actual script using the same exetensions as the DSL, but here, to generate a file.

### DIFF
--- a/src/functionalTest/kotlin/com/autonomousapps/utils/AndroidProject.kt
+++ b/src/functionalTest/kotlin/com/autonomousapps/utils/AndroidProject.kt
@@ -29,6 +29,57 @@ fun libraryFactory(projectDir: File, librarySpec: LibrarySpec): Module {
     }
 }
 
+abstract class BaseGradleProject(val projectDir: File): Module {
+    override val dir = projectDir.also { it.mkdirs() }
+    fun withBuildFile(content: String, name: String = "build.gradle") {
+        projectDir.resolve(name).also { it.parentFile.mkdirs() }.writeText(content.trimIndent())
+    }
+    fun withFile(relativePath: String, content: String ) {
+        projectDir.resolve(relativePath).also { it.parentFile.mkdirs() }.writeText(content.trimIndent())
+    }
+}
+
+abstract class AndroidGradleProject(projectDir: File): BaseGradleProject(projectDir)  {
+    val mainDir = projectDir.resolve("src/main").also { it.mkdirs() }
+    val mainJavaSrcDir = mainDir.resolve("java").also { it.mkdirs() }
+    val resDir = mainDir.resolve("res").also { it.mkdirs() }
+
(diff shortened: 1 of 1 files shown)


src/functionalTest/kotlin/com/autonomousapps/utils/AndroidProject.kt
```diff
@@ -29,6 +29,57 @@ fun libraryFactory(projectDir: File, librarySpec: LibrarySpec): Module {
...
+        projectDir.resolve(relativePath).also { it.parentFile.mkdirs() }.writeText(content.trimIndent())
+    }
+}
+
+abstract class AndroidGradleProject(projectDir: File): BaseGradleProject(projectDir)  {
+    val mainDir = projectDir.resolve("src/main").also { it.mkdirs() }
+    val mainJavaSrcDir = mainDir.resolve("java").also { it.mkdirs() }
+    val resDir = mainDir.resolve("res").also { it.mkdirs() }
+
+    fun withManifestFile(content: String, relativePath: String = "AndroidManifest.xml"  ) {
```
pavlospt: Shouldn't those be `= apply { }` to have a builder pattern?

stephanenicolas: I don't understand your comment @pavlospt. Can you be more precise please, I am still new to kotlin.

pavlospt:
```suggestion
    fun withManifestFile(content: String, relativePath: String = "AndroidManifest.xml"  )  = apply {
```
This will return the same instance of the class so that you can call more functions on that builder. Think of it like you would do in Java:

```java
Builder withManifestFile(...) { 
   // something here
   return this;
}
```

stephanenicolas: Thx @pavlospt, I agree it would be nicer but then we need to have a real builder.

PR #23, opened 2020-01-02
setup travis CI

stephanenicolas:
Still trying it. 

To setup travis : 
* go to https://travis-ci.com/getting_started
* click on the button Activate all repositories using github apps
* select your project (dependency-analysis-android-gradle-plugin)
* merge my PR
* from now on you should be able to see all PRs being built

I also add a build badge to the README in a seperate commit. It should work when travis is setup for your repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,54 @@
+language: android
+dist: trusty
+
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+
+jdk:
+  - oraclejdk8
+
+android:
+  components:
+    - platform-tools
+    - tools
+    - add-on
+    - extra-google-m2repository
+    - extra-android-m2repository
+    # for tests, 28 is needed.
+    # the components don't work out of the box on travis
+    # we install them manually in the before_install section
+    # - build-tools;28.0.3
+    # - platforms;android-29
+
+  licenses:
+    - 'android-sdk-preview-license-52d11cd2'
+    - 'android-sdk-license-.+'
+    - 'google-gdk-license-.+'
+
+branches:
+  except:
+    - gh-pages
+
+notifications:
+  email: false
+
+sudo: false
+
+before_install:
+  - touch $HOME/.android/repositories.cfg
+  - yes | sdkmanager "platforms;android-28"
+  - yes | sdkmanager "build-tools;28.0.3"
+
+script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,41 @@
...
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+
+jdk:
+  - oraclejdk8
+
+android:
+  components:
+    - build-tools-26.0.2
```
pavlospt: This seems odd. Why did you choose 26 for buold tools and android version?

stephanenicolas: Travis is stuck in time, I am trying to make it work. Tests require android SDK 28.

.travis.yml
```diff
@@ -0,0 +1,53 @@
...
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+
+jdk:
+  - oraclejdk8
```
autonomousapps: should we use a different jdk vendor?

.travis.yml
```diff
@@ -0,0 +1,53 @@
...
+  - touch $HOME/.android/repositories.cfg
+  - yes | sdkmanager "platforms;android-28"
+  - yes | sdkmanager "build-tools;28.0.3"
+
+script:
+  # use -d for debug
+  - ./gradlew clean check
+
+after_failure:
+  - cat build/reports/tests/functionalTest/classes/*
```
autonomousapps: what about the unit tests?

autonomousapps: We also use build scans, so we might consider customizing scans to include test results for easier viewing. (this is not a blocking comment.)

autonomousapps: Speaking of build scans, though, is there a way to make the scan link prominent in the CI results? Maybe it already is, but I'm assuming it would only be visible at the bottom of the console output right now.

.travis.yml
stephanenicolas: We don't need too. Oracle is still the standard provider of jdk8. And to be honest, it's so tricky to make it work that I would recommend to just keep this setting.

.travis.yml
stephanenicolas: I agree with the tests. I added a commit for it. 
For the build scan, why not but do we really need it ? Feel free to do it in a different PR, I think it would be better.

stephanenicolas:
comments are addressed @autonomousapps . I also rebased.

.travis.yml
autonomousapps: I'll respond re build scans once I see how easy it is to review test reports on Travis.

PR #29, opened 2020-01-04
Centralized the remaining output dirs into outputDirectories.kt

autonomousapps:


src/main/kotlin/com/autonomousapps/internal/outputDirectories.kt
```diff
@@ -31,3 +31,9 @@ fun getMisusedDependenciesHtmlPath(variantName: String) =
 fun getAbiAnalysisPath(variantName: String) = "${getVariantDirectory(variantName)}/abi.json"
 
 fun getAbiDumpPath(variantName: String) = "${getVariantDirectory(variantName)}/abi-dump.txt"
+
+// Root project aggregate reports. No need for variant-specific directory
+fun getMisusedDependenciesAggregatePath() = "$ROOT_DIR/misused-dependencies.txt"
+fun getMisusedDependenciesAggregatePrettyPath() = "$ROOT_DIR/misused-dependencies-pretty.txt"
+fun getAbiAggregatePath() = "$ROOT_DIR/abi.txt"
+fun getAbiAggregatePrettyPath() = "$ROOT_DIR/abi-pretty.txt"
```
pavlospt: @autonomousapps those could be `val`s ssince they do not accept any parameters that contributes to the output of their result :) wdyt?

autonomousapps: PRs welcome!

PR #62, opened 2020-02-26
Detect constant usage from Kotlin and Java dependencies, using ANTLR listeners

autonomousapps:
Partially resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/44, and fully resolves "can't detect constant usage" limitation, for both Kotlin and Java. The price (at present) is an over-eager algorithm that might report false negatives (it might fail to indicate a project is unused, even if it is). This would only happen if the unused-but-not-reported-as-such project had a very similar package structure to an actually unused project.

Future improvements to the constant detection algorithm are planned.

ANTLR listeners are used on the consuming side, to detect usage of imports of constants. On the producer side, ASM is used to detect constant (static final) members in the bytecode.


.gitignore
```diff
@@ -3,4 +3,6 @@ build
 out
 .idea
 *.iml
-*~
\ No newline at end of file
+*~
+# sometimes antlr barfs, and I don't want to add those files by accident
+src/main/gen/
```
ZacSweers: in general a good hygiene is to just add `gen`

autonomousapps:
Only fails on the smokeTest, which is expected, as I've had issues publishing to the gradle plugi portal the last few releases. That will be resolved before the next full release.

.gitignore
autonomousapps: Are you saying my .gitignore should have just `gen/`?

Issue #71, opened 2020-03-08
Dependency Analysis fails for pure java android app modules (no kotlin).


stephanenicolas:
Gradle: 6.2.1
AGP: 3.5.3 / 3.6.1
Dependency Analysis plugin : 0.22.0

**Describe the bug**
the `./gradlew buildHealth` command fails to configure a pure java (no kkotlin) android app.
```
A problem occurred configuring project ':app.
> Task with name 'compileDebugKotlin' not found in project ':app'.
```

**To Reproduce**

1. create an android app with no kotlin code
2. use the plugin

**Expected behavior**
It would work fine ;)

**Additional context**
Here is the full stack trace: 

```
* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':app.
        at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:80)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:73)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.access$600(LifecycleProjectEvaluator.java:53)
        at org.gradle.configuration.project.LifecycleProjectEvaluator$NotifyAfterEvaluate.run(LifecycleProjectEvaluator.java:199)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:402)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:394)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:165)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:250)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:158)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:92)
        at org.gradle.internal.operations.DelegatingBuildOperationExecutor.run(DelegatingBuildOperationExecutor.java:31)
[...]

PR #73, opened 2020-03-09
Do not fail to configure pure java android app modules

stephanenicolas:
Solves #71


autonomousapps:
This work was done on a different PR and is resolved.

PR #75, opened 2020-03-10
Fixed: configuration fails for pure Java Android app module.

autonomousapps:
Same as https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/73, but with a test and some incidental refactoring and test improvements.

Fixes https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/71.


src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -122,7 +122,7 @@ open class ClassListAnalysisTask @Inject constructor(
     }
 
     /**
-     * Class files generated by Kotlin source.
+     * Class files generated by Kotlin source. May be empty.
      */
     @PathSensitive(PathSensitivity.RELATIVE)
```
stephanenicolas: Both inputs should have the @Classpath annotation: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Classpath.html

autonomousapps: Good catch. I'll do a pass through all the tasks and double check normalization in a separate PR.

PR #76, opened 2020-03-10
Added support for setting subproject whitelist.

autonomousapps:
Projects not on the list will not be analyzed. If the list is empty, all subprojects are analyzed (which is the current behavior).

Example usage:
```kotlin
// Kotlin
dependencyAnalysis {
    projects.set(listOf(":entities"))
}
```

```groovy
// Groovy
dependencyAnalysis {
    projects.set([":entities"])
}

// or
// I'm actually not sure this works, but I think Groovy / Gradle magic adds an `=` 
// overload to `Property`s. 
dependencyAnalysis {
    projects = [":entities"]
}

```

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
@@ -4,7 +4,9 @@ package com.autonomousapps
 
 import org.gradle.api.Action
 import org.gradle.api.model.ObjectFactory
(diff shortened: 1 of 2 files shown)


autonomousapps:
@stephanenicolas this might interest you. I would welcome feedback. It doesn't go as far as _requiring_ you to manually apply the plugin to each of your projects, but it does give you finer control -- and maybe will let you actually complete a `buildHealth` build ;-)

PR #90, opened 2020-03-30
Fix version badge

pavlospt:
The `maven-metadata.xml` URL seemed to be outdated.


autonomousapps:
Thank you for this! The reason the badge isn't updating correctly is that I've changed the maven coordinates for the version published to the plugin portal to remove the `gradle.plugin` prefix, in order to be more "natural". When you use custom coordinates, a real human at Gradle has to vet your "new" plugin before publishing it. So I'm currently just awaiting that process to work itself out.

All that said, I still may prefer to update this badge from Maven Central. Let me think about it for a little while.

pavlospt:
@autonomousapps sure! I just happened to add the plug-in to a greenfield project and used the version in the badge which was 0.9.0 and couldn't find the mentioned tasks, so I thought of opening this PR 😄

autonomousapps:
Oh wow. I think that might actually be due to an outage on the plugin
portal today. Latest release is WAY ahead of that 😀

pavlospt:
Yeap indeed checked the releases tabs afterwards 😂

PR #103, opened 2020-04-09
Improve performance by using a simple in-memory cache.

autonomousapps:
This saves us from repeatedly re-analyzing the same bytecode.


autonomousapps:
Addresses part of https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/101

PR #113, opened 2020-04-17
Use a pretty report file for advices

stephanenicolas:
Adds a pretty report for each project. 
I want to do the same for the aggregate report task in the root project in a follow up PR.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
@@ -473,6 +473,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       }
 
       adviceReport.set(layout.buildDirectory.file(getAdvicePath(variantName)))
+      advicePrettyReport.set(layout.buildDirectory.file(getAdvicePrettyPath(variantName)))
+      adviceConsoleReport.set(layout.buildDirectory.file(getAdviceConsolePath(variantName)))
     }
 
     // Adds terminal artifacts to custom configurations to be consumed by root project for aggregate reports.
--- a/src/main/kotlin/com/autonomousapps/internal/outputDirectories.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/outputDirectories.kt
@@ -66,6 +66,12 @@ fun getAbiDumpPath(variantName: String) =
 
 fun getAdvicePath(variantName: String) = "${getVariantDirectory(variantName)}/advice.json"
 
+fun getAdvicePrettyPath(variantName: String) =
+  "${getVariantDirectory(variantName)}/advice-pretty.json"
+
+fun getAdviceConsolePath(variantName: String) =
+  "${getVariantDirectory(variantName)}/advice.txt"
+
 // Redundant plugin tasks (nb: no variants involved)
 fun getPluginAdvicePath() = "$ROOT_DIR/intermediates/advice-plugin.json"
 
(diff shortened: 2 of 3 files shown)


src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -473,6 +473,7 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       }
 
       adviceReport.set(layout.buildDirectory.file(getAdvicePath(variantName)))
+      advicePrettyReport.set(layout.buildDirectory.file(getAdvicePrettyPath(variantName)))
```
autonomousapps, line 476: 👍

src/main/kotlin/com/autonomousapps/internal/outputDirectories.kt
```diff
@@ -66,6 +66,9 @@ fun getAbiDumpPath(variantName: String) =
 
 fun getAdvicePath(variantName: String) = "${getVariantDirectory(variantName)}/advice.json"
 
+fun getAdvicePrettyPath(variantName: String) =
```
autonomousapps, line 69: 👍

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -85,6 +86,9 @@ abstract class AdviceTask : DefaultTask() {
   @get:OutputFile
   abstract val adviceReport: RegularFileProperty
 
+  @get:OutputFile
+  abstract val advicePrettyReport: RegularFileProperty
```
autonomousapps, line 90: 👍

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -93,6 +97,9 @@ abstract class AdviceTask : DefaultTask() {
     val adviceFile = adviceReport.get().asFile
     adviceFile.delete()
 
+    val advicePrettyFile = advicePrettyReport.get().asFile
```
autonomousapps, line 100: 👍

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -129,48 +136,53 @@ abstract class AdviceTask : DefaultTask() {
...
     if (didGiveAdvice) {
       chatter.chat("See machine-readable report at ${adviceFile.path}")
+      chatter.chat("See pretty report at ${advicePrettyFile.path}")
     } else {
-      chatter.chat("Looking good! No changes needed")
+      prettyReport.prettyAdvice("Looking good! No changes needed")
     }
 
     adviceFile.writeText(advices.toJson())
+    advicePrettyFile.writeText(prettyReport.toString())
```
autonomousapps, line 185: I think this will need a small change. It looks like you're writing out the console output to disk. This is ok and I have no problem with it. However, it breaks the current convention of using `-pretty.json`, which till now has only been used like this:
```
projectReportPrettyFile.writeText(unusedDirectDependencies.toPrettyString())
```
note the use of `toPrettyString()`, which is simply a json convention for printing with indentation and newlines, rather than just one long line. The output file for this also has the `.json` suffix, which isn't appropriate for plain english output like this.

If your goal is in fact to write the console output to disk, then I suggest three changes:
1. Change the filename extension to `.txt`
2. Change the outputDirectories method name to something like `getAdviceConsolePath()`
3. Change the file property name to `adviceConsole` and `adviceConsoleFile`.

I think this would be less surprising.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -191,4 +203,9 @@ abstract class AdviceTask : DefaultTask() {
     }
     filters
   }
+
+  fun StringBuilder.prettyAdvice(msg: String) {
```
autonomousapps, line 207: Should be private.

PR #115, opened 2020-04-17
Allow to rename dependencies via the DSL.

stephanenicolas:
This PR just applies the renaming map to the console output (file + displayed in terminal).

@autonomousapps I strongly feel for not making any copy of the file.
You will have the original console report file without using a renamer, and the renamer is always applied, but only to this task if it is present.

Also, I know you wanted to have a lambda instead of the map here. It is doable, but I believe it's less elegant and less convenient.
Input need to be serializable in our tasks. A map is directly serializable.
If we were to replace it with a lambda, it would work easily in kotlin, but not in gradle. In kotlin lambdas are serializable but not in groovy. In groovy though, the `Closure` type is serializable. If we would provide an overload of the method to define the transformer with both a lambda and a `Closure`, then things should work. It would mean the plugin would have to depend on the groovy runtime to use the `Closure` type. Gloops.

Though, besides being a bit harder to code, I don't think it would be so convenient for devs using the plugin. Most projects will use a map to define their dependencies and it's more convenient and efficient to build reverse map and pass it to the plugin. A transformer could easily have performance issues. 
And I am also unsure if devs would be able to use easily a transformer that would use an input of type dependency: `renamer { dependency -> ... }`, would they have to import it ? In groovy this could be painful also as there is no syntax support for knowing the members of the dependency class.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
@@ -100,8 +100,18 @@ open class DependencyAnalysisExtension(objects: ObjectFactory) : AbstractExtensi
   internal val dependencyRenamingMap: MapProperty<String, String>
     = objects.mapProperty(String::class.java, String::class.java)
 
-  fun setDependencyRenamingMap(map: Map<String, String>) {
-    dependencyRenamingMap.putAll(map)
+  /**
+   * Set a map of literal dependency declarations to semantic aliases. For example:
+   * ```
+   * dependencyAnalysis {
+   *   setDependencyRenamingMap(mapOf("commons-io:commons-io:2.6" to "commonsIo"))
+   * }
+   * ```
+   * This can be useful for projects that have extracted all dependency declarations as semantic
+   * maps.
+   */
+  fun setDependencyRenamingMap(renamer: Map<String, String>) {
+    dependencyRenamingMap.putAll(renamer)
     dependencyRenamingMap.disallowChanges()
   }
 
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
@@ -4,6 +4,7 @@ package com.autonomousapps
 
 import org.gradle.api.Action
 import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.SetProperty
 import org.gradle.kotlin.dsl.newInstance
@@ -94,6 +95,14 @@ open class DependencyAnalysisExtension(objects: ObjectFactory) : AbstractExtensi
 
   fun issues(action: Action<IssueHandler>) {
     action.execute(issueHandler)
+  }
+
+  internal val dependencyRenamingMap: MapProperty<String, String>
+    = objects.mapProperty(String::class.java, String::class.java)
+
+  fun setDependencyRenamingMap(map: Map<String, String>) {
+    dependencyRenamingMap.putAll(map)
+    dependencyRenamingMap.disallowChanges()
   }
 
   fun abi(action: Action<AbiHandler>) {
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -7,7 +7,10 @@ import org.gradle.api.GradleException
 /**
  * Only concerned with human-readable advice meant to be printed to the console.
  */
-internal class AdvicePrinter(private val consoleReport: ConsoleReport) {
+internal class AdvicePrinter(
+  private val consoleReport: ConsoleReport,
+  private val dependencyRenamingMap: Map<String, String>? = null
+) {
   /**
    * Returns "add-advice" (or null if none) for printing to console.
    */
@@ -114,6 +117,8 @@ internal class AdvicePrinter(private val consoleReport: ConsoleReport) {
     if (dependency.identifier.startsWith(":")) {
       "project(\"${dependency.identifier}\")"
     } else {
-      "\"${dependency.identifier}:${dependency.resolvedVersion}\""
+      val dependencyId = "${dependency.identifier}:${dependency.resolvedVersion}"
+      dependencyRenamingMap?.getOrDefault(dependencyId, null)
+        ?: "\"$dependencyId\""
     }
 }
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -118,7 +118,6 @@ internal class AdvicePrinter(
       "project(\"${dependency.identifier}\")"
     } else {
       val dependencyId = "${dependency.identifier}:${dependency.resolvedVersion}"
-      dependencyRenamingMap?.getOrDefault(dependencyId, null)
-        ?: "\"$dependencyId\""
+      dependencyRenamingMap?.getOrDefault(dependencyId, null) ?: "\"$dependencyId\""
     }
 }
--- a/src/main/kotlin/com/autonomousapps/tasks/AdvicePrinterTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/AdvicePrinterTask.kt
@@ -26,18 +26,18 @@ abstract class AdvicePrinterTask : DefaultTask() {
     description = "Displays advice on screen"
   }
 
-  @get:PathSensitive(PathSensitivity.RELATIVE)
+  @get:PathSensitive(PathSensitivity.NONE)
   @get:InputFile
   abstract val adviceConsoleReport: RegularFileProperty
+
+  @get:Input
+  abstract val dependencyRenamingMap: MapProperty<String, String>
 
   @get:Input
   abstract val chatty: Property<Boolean>
 
   @get:OutputFile
   abstract val adviceConsoleReportTxt: RegularFileProperty
-
-  @get:Input
-  abstract val dependencyRenamingMap: MapProperty<String, String>
 
   private val chatter by lazy { chatter(chatty.get()) }
 
--- a/src/main/kotlin/com/autonomousapps/tasks/AdvicePrinterTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/AdvicePrinterTask.kt
@@ -4,14 +4,15 @@ package com.autonomousapps.tasks
 
 import com.autonomousapps.TASK_GROUP_DEP
 import com.autonomousapps.internal.ConsoleReport
-import com.autonomousapps.internal.advice.*
-import com.autonomousapps.internal.utils.*
+import com.autonomousapps.internal.advice.AdvicePrinter
 import com.autonomousapps.internal.utils.chatter
+import com.autonomousapps.internal.utils.fromJson
+import com.autonomousapps.internal.utils.getAndDelete
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.*
-import java.lang.StringBuilder
 
 /**
  * Produces human-readable advice files and console report on how to modify a project's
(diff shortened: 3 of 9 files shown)


src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -6,7 +6,8 @@ import org.gradle.api.GradleException
 /**
  * Only concerned with human-readable advice meant to be printed to the console.
  */
-internal class AdvicePrinter(private val computedAdvice: ComputedAdvice) {
+internal class AdvicePrinter(private val computedAdvice: ComputedAdvice,
```
autonomousapps, line 9: I'm gonna nitpick stuff like this because I dislike this convention. If the line is too long, then it should be
```
class Thing(
  private val one,
  private val two
)
```

stephanenicolas: we should use ktlint...

autonomousapps: Maybe. I do have an `.editorconfig`, but I suppose it doesn't catch this.

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
```diff
@@ -63,6 +63,8 @@ open class DependencyAnalysisExtension(objects: ObjectFactory) {
   fun issues(action: Action<IssueHandler>) {
     action.execute(issueHandler)
   }
+
+  var dependencyRenamingMap: Map<String, String>? = null
```
autonomousapps, line 67: We should avoid nullable vars. Instead, use gradle's `MapProperty`. See https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_maps. Always prefer the Property/Provider API when possible.

stephanenicolas: I didn't know about the map property. Thx.
For the null, we could use an empty map instead, but the performance could be a little degraded. It should work though.

stephanenicolas:
As soon as we know how to proceed for the rest, I will make all the changes to finish the PR. 
Let's find an approach soon so I can work on it on Monday @autonomousapps , it would be a good timing for me.

stephanenicolas:
This PR will be reworked after https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/122 is merged.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -7,7 +7,8 @@ import org.gradle.api.GradleException
 /**
  * Only concerned with human-readable advice meant to be printed to the console.
  */
-internal class AdvicePrinter(private val consoleReport: ConsoleReport) {
+internal class AdvicePrinter(private val consoleReport: ConsoleReport,
```
autonomousapps, line 10: style nit

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
autonomousapps: Can this also be a `MapProperty`?

autonomousapps:
Also, rebase onto master. The CI build failed because a test ran out of memory. I made some changes just yesterday to mitigate that problem.

stephanenicolas:
@autonomousapps thx for the review, it's appreciated. On my end, I believe the feature to be complete after I merge. So if you wanna release, it's good on my end.

For the tests, I had made a small mistake on top of the memory issues. For the MapProperty, I am wondering if it would be better for users ? Wouldn't it be paintful to instantiate it in the extension ?

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
```diff
@@ -63,6 +64,10 @@ open class DependencyAnalysisExtension(objects: ObjectFactory) {
   fun issues(action: Action<IssueHandler>) {
     action.execute(issueHandler)
   }
+
+  var dependencyRenamingMap = objects.mapProperty(String::class.java, String::class.java).also {
```
autonomousapps, line 68: should be `val`.

To your question about ease of use, Groovy DSL users (who are more than 95% of all gradle users), can do
```groovy
dependencyAnalysis {
  dependencyRenamingMap = [...]
}
```
(Gradle magic makes that work)
Kotlin DSL users must do:
```kotlin
dependencyAnalysis {
  dependencyRenamingMap.putAll(mapOf(...))
}
```
I think that's fine.

src/main/kotlin/com/autonomousapps/tasks/AdvicePrinterTask.kt
```diff
@@ -35,6 +36,10 @@ abstract class AdvicePrinterTask : DefaultTask() {
   @get:OutputFile
   abstract val adviceConsoleReportTxt: RegularFileProperty
 
+  @get:Input
+  @get:Optional
```
autonomousapps, line 40: I don't think it's optional. You set this input unconditionally. Right?

autonomousapps:
Also lots of test failures: https://scans.gradle.com/s/tnivc3kicc3jg/tests/failed

stephanenicolas:
Yes, I couldn't work on that yesterday, but I am now.

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
stephanenicolas: I find the `putAll` not so elegant. It looks like there is a state already and you want to add new content. 

Do you mind if I would keep the var ?

stephanenicolas:
@autonomousapps I couldn't apply the requested changes: 

When I use a map property in the extension, it looks like there is a bug in early gradle versions  (before 6.3) that crashes the tests. I couldn't pinpoint the exact problem because tests take a long time to run (I tweaked the quick tests to run early gradle versions).
I think the issue is to use a null convention value for the `MapProperty`.

Though, I wanted to use this null value which seems better to me than an empty map, especially as the advice printer uses this map to display all dependencies, I was thinking that accessing an empty map was not efficient.

Because I wanted to keep the null value, and make the tests work for early gradle versions:
* I removed the map property from the extension
* I use a nullable var map in the extension, with a null default
* I do always assign the property as an input to the printer task, but now, it is assigned to null by default, which is only possible when the input is marked as optional.

Actually, I know it was not your initial plan, but I like the semantic from a user perspective: the optional map, the use of `=` more than `putAll`. I think the API is better, but are you ok with this choice ?

autonomousapps:
I like gradle-managed properties over `var`s because there's an ordering issue.

If the tasks -- for some reason -- get configured eagerly, they'll pull the map value from the extension _before_ it has been set by the user. So it will be null, even if you set it to something non-null. This violates the principle of least surprise. This is why I've been using properties exclusively to connect the extension to tasks, even when the developer UX seems less elegant.

We know there are many poorly written plugins out there that can cause tasks to be eagerly realized. We can't just ignore that fact. So, properties are the way to go.

That said, you can still provide users a nicer UX with properties backing. Something like:

**plugin code**
```kotlin
internal val map: MapProperty<String, String> = objects.mapProperty().also {
  it.convention(null) // I think this is redundant, actually
}
fun setMap(userMap: Map<String, String>) {
  map.putAll(userMap)
  map.disallowChanges()
}
```
**user code, groovy**
```groovy
dependencyAnalysis {
  map = myMap // I think it works like this in groovy...?
}
```

**user code, kotlin**
```kotlin
dependencyAnalysis {
  map = myMap // not sure if this works, but if not:
  setMap(myMap)
}
```
caveat: coding in a github comment

stephanenicolas:
Thx for the explanations @autonomousapps . 
I made the change, it should pass CI. 

I could get rid of the issue for gradle < 6.3 by removing the convention call. 
I removed the `@optional` annotation. 
I added the easy `fun` to the DSL, I tested it and it works well.

stephanenicolas:
I had forgotten a change where I rename it:
```
-  fun setMap(map: Map<String, String>) {
+  fun setDependencyRenamingMap(map: Map<String, String>) {
```
Are you ok with the name. It's a bit verbose but clear..

autonomousapps:
Yes, that name is fine. Verbose == clear, in this case.

autonomousapps:
I'm not sure this works? I just tried it out on another project, with:
```kotlin
dependencyAnalysis {
    setDependencyRenamingMap(mapOf("org.jetbrains.kotlin:kotlin-stdlib-jdk8" to "kotlin8"))
}
```
I also tried inverting the order. In both cases, the printed advice still used `"org.jetbrains.kotlin:kotlin-stdlib-jdk8"`.

This leads me to my next point, which is that this is a major new feature which _has_ to have functional tests. I'd put them in a whole new "Spec" class. Call it `RenamingSpec`?

autonomousapps:
It would be useful to add one or more tests for this new feature.

Please rebase onto master and then take a look at `FacadeFilterSpec`, which uses a new approach to writing test fixtures, for jvm (non-android) projects.* 

The entry point for creating a new test fixture is `com.autonomousapps.fixtures.jvm.JvmProject.Builder` in `functionalTest/groovy`. Apologies for all the Groovy, but since we're using Spock I just decided it made sense to write more of the fixture code in groovy to reduce compatibility issues.

I think it would be useful to write functional tests against jvm projects for two reasons:
1. They should be easier to write and faster to run, and our test suite already takes quite a while
2. We have relatively little coverage of jvm projects (vs android) and it would be good to rectify that

*now that I'm writing this, I realize that this could probably be easily extended to Android. Something I'll take a look at later

ZacSweers:
I'd rather a service loader or a custom task that this plugin could pass args to and read a remapped result from. That would allow a lot more flexibility

stephanenicolas:
@autonomousapps as you will in the new test, the version number is also part of the renaming map entries:

```
def renamingMap = new HashMap()
renamingMap.put('commons-io:commons-io:2.6', 'commons-io')
dependencyAnalysis.dependencyRenamingMap = renamingMap
```

autonomousapps:
@ZacSweers I wanted to follow-up on your comment. @stephanenicolas and I discussed it offline, and I explained that you meant something like
```kotlin
dependencyAnalysis {
  registerIdentifierMappingTask(provider: TaskProvider<out MappingTaskInterfaceYouProvide>)
}
abstract class MappingTaskInterface : DefaultTask() {
  @get:OutputFile
  abstract val remapped: MapProperty<String, String>
}
```
or, to speak to the bit about "service loaders", that the plugin could look for a service loader on the classpath and wire this up automatically.

While I am not opposed to something like this, I think it's at least several steps away along an iterative path that starts with what we already have in this PR. The cognitive overhead of a simple map is very low by comparison. There are also some preexisting issues (written down in my brain...) that would limit the utility of this approach, and which should be resolved before we tried something like this. Finally, I think it might be possible to have these two approaches living side by side -- eventually.

ZacSweers:
I think it would be sensible to make the logic implemented via task and then just make a simple helper that allows folks to use the map-based approach

```
registerIdentifierMappingTask(MappingTaskInterface.of(yourMapHere))
```

autonomousapps:
I just tried this out on a personal project, and the console output looks nice. One thing I found surprising was that this only seems to transform `advice-console.txt`. Is this intentional? Why not transform all `advice*` output?

What is our ultimate goal here?

stephanenicolas:
For me the goal was to have a user readable report that would use the mappings that we also use internally: `deps.foo = "org.foo:bar:1.0.0"`
And I only needed the console report to respect this mapping. I am using the tool and it works well for me. 

I am open to other requirements, but I don't know which one you have in mind. Also, I must say I would like to see this PR merged and released, it would be a little bit easier.

autonomousapps:
I could accept that only if it was understood that it was just the first increment. I think that for non-author users, they would expect that the mapping would have a more _comprehensive_ impact on the output. 

I think all of the `advice*.[json|txt]` files should be mapped. Either copied to new files with the mapped names or written in place with the mapped names. And this should be at both the subproject level and the root, aggregate-report level.

stephanenicolas:
I am with you @autonomousapps , it would make more sense that all data generated by the plugin could also handle this renaming map. But I wonder how to do that elegantly. How to have all tasks's output being duplicated and renamed. 

Maybe we should have a wrapper "service" around all output that would always duplicate the files in presence of a renaming map. But yes, for now, I would really love to have the PR in: it's not only a useful tool for me (and hopefully/probably others) but it is also the second one of a larger refactor to make it possible. It's been quite some work that I would like to see materialized.

autonomousapps:
I have rebased this onto master locally and made some minor changes. I'll push them to your branch and merge some time today.

Now that I've explored the code some more, it's clear to me that the renaming map should be input to a new task that runs after `AdviceTask` and `AdvicePrinterTask` that takes the old output and writes new files with new, remapped content (or _maybe_ rewrite the files in place). I think that is by far the cleanest approach. I think that's non-negotiable.

Nevertheless, this is mergeable (and thanks for the test!), and I hope it unblocks you. We'll iterate soon.

PR #122, opened 2020-04-21
Model console report and dissociate display task for humans

stephanenicolas:
The PR is not polished yet.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
@@ -464,6 +464,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       output.set(outputPaths.pluginKaptAdvicePath)
     }
 
+    val advicePrinterTask = tasks.register<AdvicePrinterTask>("advicePrinter$variantTaskName")
+
     // Combine "misused dependencies" and abi reports into a single piece of advice for how to alter one's
     // dependencies
     val adviceTask = tasks.register<AdviceTask>("advice$variantTaskName") {
@@ -494,6 +496,14 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       adviceReport.set(outputPaths.advicePath)
       advicePrettyReport.set(outputPaths.advicePrettyPath)
       adviceConsoleReport.set(outputPaths.adviceConsolePath)
+      adviceConsolePrettyReport.set(outputPaths.adviceConsolePrettyPath)
+      finalizedBy(advicePrinterTask)
+    }
+
+    advicePrinterTask.configure {
+      adviceConsoleReport.set(adviceTask.flatMap { it.adviceConsoleReport })
+      chatty.set(getExtension().chatty)
+      adviceConsoleReportTxt.set(outputPaths.adviceConsoleTxtPath)
     }
 
     // Adds terminal artifacts to custom configurations to be consumed by root project for aggregate reports.
--- a/src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
@@ -6,6 +6,7 @@ import com.autonomousapps.Behavior
 import com.autonomousapps.TASK_GROUP_DEP
 import com.autonomousapps.advice.Dependency
 import com.autonomousapps.internal.*
+import com.autonomousapps.internal.ConsoleReport
 import com.autonomousapps.internal.advice.*
 import com.autonomousapps.internal.utils.*
 import com.autonomousapps.internal.utils.chatter
@@ -13,7 +14,6 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.*
-import java.lang.StringBuilder
 
 /**
  * Produces human- and machine-readable advice on how to modify a project's dependencies in order to have a healthy
@@ -90,6 +90,9 @@ abstract class AdviceTask : DefaultTask() {
   @get:OutputFile
   abstract val adviceConsoleReport: RegularFileProperty
 
+  @get:OutputFile
+  abstract val adviceConsolePrettyReport: RegularFileProperty
+
   private val chatter by lazy { chatter(chatty.get()) }
 
   @TaskAction
@@ -98,6 +101,7 @@ abstract class AdviceTask : DefaultTask() {
     val adviceFile = adviceReport.getAndDelete()
     val advicePrettyFile = advicePrettyReport.getAndDelete()
     val adviceConsoleFile = adviceConsoleReport.getAndDelete()
+    val adviceConsolePrettyFile = adviceConsolePrettyReport.getAndDelete()
 
     // Inputs
     val allComponents = allComponentsReport.get().asFile.readText().fromJsonList<Component>()
@@ -132,58 +136,17 @@ abstract class AdviceTask : DefaultTask() {
 
     val computedAdvice = advisor.compute(filterSpecBuilder())
     val advices = computedAdvice.getAdvices()
-    val advicePrinter = AdvicePrinter(computedAdvice)
-
-    var didGiveAdvice = false
-    val consoleReport = StringBuilder()
-
-    if (!computedAdvice.filterRemove) {
-      advicePrinter.getRemoveAdvice()?.let {
-        consoleReport.add("Unused dependencies which should be removed:\n$it\n")
-        didGiveAdvice = true
-      }
-    }
-
-    if (!computedAdvice.filterAdd) {
-      advicePrinter.getAddAdvice()?.let {
-        consoleReport.add("Transitively used dependencies that should " +
-          "be declared directly as indicated:\n$it\n")
-        didGiveAdvice = true
-      }
-    }
-
-    if (!computedAdvice.filterChange) {
-      advicePrinter.getChangeAdvice()?.let {
-        consoleReport.add("Existing dependencies which should be modified " +
-          "to be as indicated:\n$it\n")
-        didGiveAdvice = true
-      }
-    }
-
-    if (!computedAdvice.filterCompileOnly) {
-      advicePrinter.getCompileOnlyAdvice()?.let {
-        consoleReport.add("Dependencies which could be compile-only:\n$it\n")
-        didGiveAdvice = true
-      }
-    }
-
-    // TODO add filter
-    advicePrinter.getRemoveProcAdvice()?.let {
-      consoleReport.add("Unused annotation processors that should be removed:\n$it\n")
-      didGiveAdvice = true
-    }
-
-    if (didGiveAdvice) {
-      chatter.chat("See machine-readable report at ${adviceFile.path}")
-      chatter.chat("See pretty report at ${advicePrettyFile.path}")
-      chatter.chat("See console report at ${adviceConsoleFile.path}")
-    } else {
-      consoleReport.add("Looking good! No changes needed")
-    }
+    val consoleReport = ConsoleReport.from(computedAdvice)
 
     adviceFile.writeText(advices.toJson())
     advicePrettyFile.writeText(advices.toPrettyString())
-    adviceConsoleFile.writeText(consoleReport.toString())
+    adviceConsoleFile.writeText(consoleReport.toJson())
+    adviceConsolePrettyFile.writeText(consoleReport.toPrettyString())
+
+    chatter.chat("See machine-readable report at ${adviceFile.path}")
+    chatter.chat("See pretty report at ${advicePrettyFile.path}")
+    chatter.chat("See machine-readable console report at ${adviceConsoleFile.path}")
+    chatter.chat("See pretty console report at ${adviceConsolePrettyFile.path}")
   }
 
   private fun filterSpecBuilder() = FilterSpecBuilder().apply {
@@ -203,10 +166,5 @@ abstract class AdviceTask : DefaultTask() {
       filters.add(ViewBindingFilter())
     }
     filters
-  }
-
-  private fun StringBuilder.add(msg: String) {
-    chatter.chat(msg)
-    this.append(msg)
   }
 }
--- a/src/test/kotlin/com/autonomousapps/internal/advice/AdvisorTest.kt
+++ b/src/test/kotlin/com/autonomousapps/internal/advice/AdvisorTest.kt
@@ -6,6 +6,7 @@ import com.autonomousapps.advice.Advice
 import com.autonomousapps.advice.Dependency
 import com.autonomousapps.internal.*
 import com.autonomousapps.internal.utils.filterToOrderedSet
+import com.autonomousapps.tasks.AdviceTask
 import org.junit.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
@@ -52,7 +53,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute()
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNotNull(changeAdvice, "Change advice should not be null")
@@ -86,7 +88,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { incorrectConfigurationsBehavior = Ignore })
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNull(changeAdvice, "Change advice should be null")
@@ -115,7 +118,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { anyBehavior = Ignore })
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNull(changeAdvice, "Change advice should be null")
@@ -148,7 +152,8 @@ class AdvisorTest {
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply {
       incorrectConfigurationsBehavior = Warn(setOf(library5.identifier))
     })
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNotNull(changeAdvice, "Change advice should not be null")
@@ -177,7 +182,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { anyBehavior = Warn(setOf(library5.identifier)) })
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNotNull(changeAdvice, "Change advice should not be null")
@@ -212,7 +218,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder())
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val changeAdvice = AdvicePrinter(consoleReport).getChangeAdvice()
 
     // Then
     assertNull(changeAdvice, "There should be nothing to change")
@@ -245,7 +252,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder())
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNotNull(addAdvice, "Add advice should not be null")
@@ -277,7 +285,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { usedTransitivesBehavior = Ignore })
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNull(addAdvice, "Add advice should be null")
@@ -305,7 +314,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { anyBehavior = Ignore })
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNull(addAdvice, "Add advice should not be null")
@@ -335,7 +345,8 @@ class AdvisorTest {
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply {
       usedTransitivesBehavior = Warn(setOf(project1.identifier))
     })
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNotNull(addAdvice, "Add advice should not be null")
@@ -366,7 +377,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder().apply { anyBehavior = Warn(setOf(project1.identifier)) })
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNotNull(addAdvice, "Add advice should not be null")
@@ -402,7 +414,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute(FilterSpecBuilder())
-    val addAdvice = computedAdvice.advicePrinter().getAddAdvice()
+    var consoleReport = ConsoleReport.from(computedAdvice)
+    val addAdvice = AdvicePrinter(consoleReport).getAddAdvice()
 
     // Then
     assertNotNull(addAdvice, "Add advice should not be null")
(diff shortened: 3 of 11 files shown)


src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -457,6 +457,12 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       output.set(outputPaths.pluginKaptAdvicePath)
     }
 
+    val adviceTransformTask = tasks.register<AdviceTransformTask>("adviceTransform$variantTaskName") {
```
stephanenicolas, line 460: I would have preferred to declare it bedore `AdviceTask` but I couldn't. 
This task here will never be called and the only way I could find to have it configured lazily is to make `AdviceTask` being finalized by it. And this forces the task `AdviceTransformTask` to be declared before.
I write this comment out of curiosity, if you know a better way to do it @autonomousapps , I would be interested to learn about it.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,85 @@ abstract class AdviceTask : DefaultTask() {
...
+
+    fun isEmpty() = addToApiAdvice.isEmpty() &&
+      removeAdvice.isEmpty() &&
+      changeToApiAdvice.isEmpty() &&
+      changeToImplAdvice.isEmpty() &&
+      compileOnlyDependencies.isEmpty() &&
+      unusedProcsAdvice.isEmpty()
+
+    companion object {
+      fun build(computedAdvice: ComputedAdvice): ConsoleReport {
```
stephanenicolas, line 176: There might be a better way to do this in kotlin. A builder ? My kotlin fluency...

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,85 @@ abstract class AdviceTask : DefaultTask() {
...
+        if (!computedAdvice.filterChange && changeAdvices.isNotEmpty()) {
+          changeToApiAdvice = computedAdvice.changeToApiAdvice
+          changeToImplAdvice = computedAdvice.changeToImplAdvice
+        }
+
+        if (!computedAdvice.filterCompileOnly && computedAdvice.compileOnlyDependencies.isNotEmpty()) {
+          compileOnlyDependencies = computedAdvice.compileOnlyDependencies
+        }
+
+        // TODO add filter
```
stephanenicolas, line 205: If the TODO is just about making a filter in the same way as others, without anything specific, I can do that.

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.autonomousapps.internal.utils.chatter
+import com.autonomousapps.tasks.AdviceTask.ConsoleReport
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.*
+import java.lang.StringBuilder
+
+/**
+ * Produces human- and machine-readable advice on how to modify a project's dependencies in order to have a healthy
```
stephanenicolas, line 17: TODO change comment, we just produce the human-readable file.

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
```diff
@@ -0,0 +1,98 @@
...
+ * build.
+ */
+@CacheableTask
+abstract class AdviceTransformTask : DefaultTask() {
```
stephanenicolas, line 21: Do you have a better name @autonomousapps ? `AdviceConsoleTask` ?

autonomousapps:
Please rebase this onto master and force-push. I'm not using travis anymore, as it was timing out. CI is now done via Github Actions. (I have no confidence that travis's green light is legit.)

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
autonomousapps: I had to do something similar here:
```kotlin
    // Configured below
    val failOrWarn = tasks.register<FailOrWarnTask>("failOrWarn")

    val adviceReport = tasks.register<AdviceAggregateReportTask>("adviceReport") {
      dependsOn(adviceReportsConf, advicePluginsConf)

      adviceReports = adviceReportsConf
      advicePluginReports = advicePluginsConf
      chatty.set(getExtension().chatty)

      projectReport.set(outputPaths.adviceAggregatePath)
      projectReportPretty.set(outputPaths.adviceAggregatePrettyPath)

      finalizedBy(failOrWarn)
    }

    failOrWarn.configure { ... }
```

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -486,7 +492,9 @@ class DependencyAnalysisPlugin : Plugin<Project> {
 
       adviceReport.set(outputPaths.advicePath)
       advicePrettyReport.set(outputPaths.advicePrettyPath)
-      adviceConsoleReport.set(outputPaths.adviceConsolePath)
+      adviceConsoleReport.set(adviceTransformTask.flatMap { it.adviceConsoleReport })
```
autonomousapps, line 495: I don't think this works. I think this would force gradle to create a task graph where `adviceTransformTask` had to run _before_ `adviceTask`. This of course would be impossible given the `finalizedBy`.

I think I understand your earlier concern now. What you want is to just register the transform task above:
```kotlin
val adviceTransformTask = tasks.register<AdviceTransformTask>("adviceTransform$variantTaskName")
```
and then below `adviceTask`
```kotlin
adviceTransformTask.configure { ... }
```

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
```diff
@@ -78,12 +78,8 @@ inline fun <reified K, reified V> String.fromJsonMapList(): Map<K, List<V>> {
   return adapter.fromJson(this)!!
 }
 
-inline fun <reified T> List<T>.toPrettyString(withNulls: Boolean = false): String {
-  return getJsonListAdapter<T>(withNulls).indent("  ").toJson(this)
-}
-
-inline fun <reified T> Set<T>.toPrettyString(withNulls: Boolean = false): String {
-  return getJsonSetAdapter<T>(withNulls).indent("  ").toJson(this)
+inline fun <reified T> T.toPrettyString(withNulls: Boolean = false): String {
```
autonomousapps, line 81: does this actually work? I thought we had to be explicit about it being a `Collection<T>`

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,85 @@ abstract class AdviceTask : DefaultTask() {
...
-        consoleReport.add("Dependencies which could be compile-only:\n$it\n")
-        didGiveAdvice = true
+  internal data class ConsoleReport(
+    val addToApiAdvice: Set<Advice> = emptySet(),
+    val addToImplAdvice: Set<Advice> = emptySet(),
+    val removeAdvice: Set<Advice> = emptySet(),
+    val changeToApiAdvice: Set<Advice> = emptySet(),
+    val changeToImplAdvice: Set<Advice> = emptySet(),
+    val compileOnlyDependencies: Set<Dependency> = emptySet(),
+    val unusedProcsAdvice: Set<Advice> = emptySet()) {
```
autonomousapps, line 166: nit: `) {` should be on new line

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
autonomousapps: Yes, it's about making a filter in the same way as the other filters (trickling down from `DependencyAnalysisExtension`, but I wouldn't bother with that in this PR. It seems increasingly likely I'm going to rewrite how filtering / customization works.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,85 @@ abstract class AdviceTask : DefaultTask() {
...
+        if (!computedAdvice.filterCompileOnly && computedAdvice.compileOnlyDependencies.isNotEmpty()) {
+          compileOnlyDependencies = computedAdvice.compileOnlyDependencies
+        }
+
+        // TODO add filter
+        if (computedAdvice.unusedProcsAdvice.isNotEmpty()) {
+          unusedProcsAdvice = computedAdvice.unusedProcsAdvice
+        }
+
+        return ConsoleReport(addToApiAdvice,
```
autonomousapps, line 210: nit: `addToApiAdvice` should be on next line, and terminal `)` should also be on its own line.

Yes I know, we need a linter.

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
```diff
@@ -0,0 +1,98 @@
...
+
+  init {
+    group = TASK_GROUP_DEP
+    description = "Displays advice on screen"
+  }
+
+  @get:PathSensitive(PathSensitivity.RELATIVE)
```
autonomousapps, line 28: I know I use this in many places, but I think I'm wrong. I strongly suspect we can use `PathSensitivity.NONE`. So, this isn't a blocker, but if you get a chance to test NONE vs RELATIVE, that would be interesting.

Longer discussion: the documentation on this is very poor, but after a discussion with someone on the gradle community slack, I believe I understand it better. What `RELATIVE` means is that, if the path to the file input changes, then the input is considered changed -- even if the contents are the same. Since we don't _actually_ care about the path but only the content, I think `NONE` is appropriate. I have an open issue to go over all task inputs to check for this kind of thing, but I haven't had time yet. So this is more of an FYI / information-sharing comment.

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
```diff
@@ -0,0 +1,98 @@
...
+
+  private val chatter by lazy { chatter(chatty.get()) }
+
+  @TaskAction
+  fun action() {
+    // Output
+    val adviceConsoleReportTxtFile = adviceConsoleReportTxt.getAndDelete()
+
+    // Inputs
+    val adviceConsoleReportFile = adviceConsoleReport.asFile.get()
```
autonomousapps, line 46: huh, interesting. Not a big deal, but I normally invert the order:
```kotlin
adviceConsoleReport.get().asFile
```

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    val consoleReport = adviceConsoleReportFile
```
autonomousapps, line 48: I added a companion utility to `getAndDelete()`, in the same file, that looks like:
```kotlin
internal inline fun <reified T> RegularFileProperty.fromJsonSet(): Set<T> {
  return get().asFile.readText().fromJsonSet()
}
```
you could write companions to that that aren't `Set`s.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
stephanenicolas: Let see how it goes to get the renamer feature done. I create a ticket for it.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
stephanenicolas: Let me double checked but yes, I think it did for me.

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
stephanenicolas: ok

src/main/kotlin/com/autonomousapps/tasks/AdviceTransformTask.kt
stephanenicolas: ok

stephanenicolas: The doc is definitely poor of on this side.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
autonomousapps: Good find; thanks.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
autonomousapps: I think the method should be named `from`, not `build`. You're basically converting an instance of `ComputedAdvice` into an instance of `ConsoleReport`, and I think the right verb for that is "from".

I also found it surprising to see a bunch of `var`s. My instinct with Kotlin code is to use `val`s to a great extremity :) Having said that, to use vals instead of vars here, these would need to be mutable sets and you'd use `addAll()`, and there would be efficiency implications. Having said _that_, I don't know that this area here is an actual hotspot, so it may not matter.

If the choice is between `var` or a mutable set, I prefer the `var`. We can imagine using val and `by lazy { ... }`, but that might be harder to read and comprehend, and I prefer easier comprehension.

src/test/kotlin/com/autonomousapps/internal/advice/AdvisorTest.kt
```diff
@@ -52,7 +54,8 @@ class AdvisorTest {
       unusedProcs = emptySet()
     )
     val computedAdvice = advisor.compute()
-    val changeAdvice = computedAdvice.advicePrinter().getChangeAdvice()
+    var consoleReport = build(computedAdvice)
```
autonomousapps, line 57: I think I like it better if you don't import the "static" function. Reading in github, it looks like a local function, and I prefer it "namespaced".

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
stephanenicolas: So, I just rename, right ?

autonomousapps: Yes. Sorry, I may be being too verbose, but I'm basically thinking aloud so we can have a kind of async conversation in case there is more to react to.

I can, uh, try to _reduce_ that in the future.

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -457,6 +457,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       output.set(outputPaths.pluginKaptAdvicePath)
     }
 
+    val adviceTransformTask = tasks.register<AdviceConsoleTask>("adviceTransform$variantTaskName")
```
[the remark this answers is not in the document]
stephanenicolas: done. I renamed the task too.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,88 @@ abstract class AdviceTask : DefaultTask() {
...
+    adviceConsolePrettyFile.writeText(consoleReport.toPrettyString())
 
-    if (!computedAdvice.filterAdd) {
-      advicePrinter.getAddAdvice()?.let {
-        consoleReport.add("Transitively used dependencies that should " +
-          "be declared directly as indicated:\n$it\n")
-        didGiveAdvice = true
-      }
-    }
+    chatter.chat("See machine-readable report at ${adviceFile.path}")
```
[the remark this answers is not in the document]
stephanenicolas: These specifically should have only been done in the `AdvicePrinterTask`. I changed that.
The files are always output in the `AdviceTask`, we  print the location of the files on screen if we are chatty.
In the printer task, the `if` gates the message to be displayed. Let me know how you like this.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,88 @@ abstract class AdviceTask : DefaultTask() {
...
+    } else {
+      chatter.chat("See console report at ${adviceConsoleFile.path}")
     }
+  }
 
-    if (!computedAdvice.filterCompileOnly) {
-      advicePrinter.getCompileOnlyAdvice()?.let {
-        consoleReport.add("Dependencies which could be compile-only:\n$it\n")
-        didGiveAdvice = true
+  internal data class ConsoleReport(
```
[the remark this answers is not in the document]
stephanenicolas: There was no `.internal.models`, so I assumed you meant `.internal.advice` as other advice related classes were already present.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -132,58 +137,88 @@ abstract class AdviceTask : DefaultTask() {
...
+    val addToApiAdvice: Set<Advice> = emptySet(),
```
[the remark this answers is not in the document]
stephanenicolas: In the new commit I am gonna push, I will need your help to test compileOnly components, the rest is tested and working: 
`com/autonomousapps/internal/advice/ConsoleReportTest.kt:239`

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -457,6 +457,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       output.set(outputPaths.pluginKaptAdvicePath)
     }
 
+    val adviceTransformTask = tasks.register<AdvicePrinterTask>("advicePrinter$variantTaskName")
```
autonomousapps, line 460: may as well change the variable name as well

src/main/kotlin/com/autonomousapps/internal/advice/ConsoleReport.kt
```diff
@@ -0,0 +1,70 @@
+package com.autonomousapps.internal.advice
```
autonomousapps, line 1: wrong package. I meant either
1. `com.autonomousapps.advice` for something that is public-facing; or
2. `com.autonomousapps.internal.models.kt` if it's only for internal use.

Note that the second one refers to a _file_ named `models.kt`, which has a bunch of `data class` es in it.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
autonomousapps: I ended up merging this bit onto master separately, so you may have a minor merge conflict here when you rebase.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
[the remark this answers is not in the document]
autonomousapps: It looks better. Some of the output seems redundant, but I don't really care right now (I think we can make a readability pass at some point in the future).

I am a bit confused by `advice-console.txt` and `advice-console-pretty.json`. Those files have similar names, but their contents are radically different. One looks basically like a json-prettified version of a `List<Advice>` (with names, so more like a Map I guess), while the other is a literal record of the console output.

In an attempt to move this forward, here's a concrete proposal. Let me know what you think.
1. Rename `advice-console-pretty.json` to `advice-simplified.json`
2. Print just the `Dependency` objects, rather than the full `Advice` object. As you've already noted, the extra info contained by `Advice` becomes redundant when you present it as a map like this.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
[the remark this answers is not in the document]
autonomousapps: Not exactly. I commented on the new file.

src/test/kotlin/com/autonomousapps/internal/advice/ConsoleReportTest.kt
```diff
@@ -0,0 +1,305 @@
...
+    classes = emptySet()
+  )
+  /* ****************************************
+   * Change advice (incorrect configurations)
+   * ****************************************
+   */
+
+  @Test fun `isEmpty should return true when empty`() {
+    // When
+    var consoleReport = createReport()
```
autonomousapps, line 27: all of these `var`s can, and should, be `val`s. Always default to `val`. Kotlin Rule #1. Only use `var` if you can't, for some reason, use `val`.

src/test/kotlin/com/autonomousapps/internal/advice/ConsoleReportTest.kt
```diff
@@ -0,0 +1,305 @@
...
+
+    // When
+    var consoleReport = ConsoleReport.from(computedAdvice = computedAdvice)
+
+    // Then
+    val expected = setOf(
+      Advice.compileOnly(orgDotSomethingV1, "compileOnly")
+    )
+    val actualApi = consoleReport.compileOnlyDependencies
+    assertEquals(expected, actualApi, "Expected $expected\nActual   $actualApi\n")
```
autonomousapps, line 261: This must fail, because `expected` is a `Set<Advice>`, while `actualApi` is a `Set<Dependency>`. My IDE is warning me about the type mismatch -- doesn't yours?

src/test/kotlin/com/autonomousapps/internal/advice/ConsoleReportTest.kt
```diff
@@ -0,0 +1,305 @@
...
+    var consoleReport = ConsoleReport.from(computedAdvice = computedAdvice)
+
+    // Then
+    val adviceApi = consoleReport.changeToApiAdvice
+    assertTrue(adviceApi.isEmpty(), "Expected $adviceApi to be empty")
+    val adviceImpl = consoleReport.changeToImplAdvice
+    assertTrue(adviceImpl.isEmpty(), "Expected $adviceImpl to be empty")
+  }
+
+  private fun createReport(addToApiAdvice: Set<Advice> = emptySet(),
```
autonomousapps, line 289: please fix stye. Each param on its own line, including the first. That will also fix the indentation. This much whitespace hurts my brain :)

stephanenicolas:
> In the new commit I am gonna push, I will need your help to test compileOnly components, the rest is tested and working:
com/autonomousapps/internal/advice/ConsoleReportTest.kt:239

src/main/kotlin/com/autonomousapps/internal/advice/ConsoleReport.kt
stephanenicolas: I moved it to the `com.autonomousapps.internal.models.kt` file. I really hesitated, just because the data is serialized but at the same time it doesn't make much sense to consume this report as it can be regenerated from the advice data directly.

stephanenicolas:
> In an attempt to move this forward, here's a concrete proposal. Let me know what you think. 
> 1. Rename advice-console-pretty.json to advice-simplified.json
> 2. Print just the Dependency objects, rather than the full Advice object. As you've already noted, the extra info contained by Advice becomes redundant when you present it as a map like this.

I don't think it clarifies the situation much more. What confuses me is the `advice-simplified.json` name. 

Indeed, the ideal situation would be to make the `ComputedAdvice` class serializable in json. I don't know if it's possible as I don't know enough kotlin and moshi.

We would like to have:
A) the AdviceTask that saves a `ComputedAdvice` on disk + pretty json (`advice.json`).
B) the AdvicePrinterTask that would load the `ComputedAdvice` json, issue the report both on display and on disk (`console-report.txt`)and optionally a report + a pretty json (console-report.json).

I think this would be the more clear. But it doesn't look like we can do this because we have more than data in `ComputedAdvice`, the main issue being the functional representation of filters instead of data to create this functional representation.
Ideally we would have a very rich representation of the lists of advice and the filters defined by the users that would be pure data, and be able to persist it on disk.
I am not sure these 2 concepts should be merged, and I think we should also better distinguish the user filters from the various internal category filters.
But all of that is quite some refactoring !

So, in a few words, I don't know how to make another solution clear, concepts are blending between advice, computed advice, filters, and reports and the code shows it. Let me know what change you want me to make, at this stage I would focus on merging the PR.

I agree to have lists of dependencies only in the `ConsoleReport` class. For the file names, I am not sure of what is better. Let me know and will make the change.

src/test/kotlin/com/autonomousapps/internal/advice/ConsoleReportTest.kt
stephanenicolas: I didn't notice :D.

stephanenicolas: The test was failing for this, but I got a logic problem here that I don't know how to workaround. Can you point to a test of `compileOnly` so that there is a non empty list of advices being computed for this config ?

I don't know what value `componentV1` should have.

src/main/kotlin/com/autonomousapps/internal/advice/ConsoleReport.kt
```diff
@@ -1,16 +1,15 @@
...
 
 internal data class ConsoleReport(
-  val addToApiAdvice: Set<Advice>,
-  val addToImplAdvice: Set<Advice>,
-  val removeAdvice: Set<Advice>,
-  val changeToApiAdvice: Set<Advice>,
-  val changeToImplAdvice: Set<Advice>,
-  val compileOnlyDependencies: Set<Dependency>,
-  val unusedProcsAdvice: Set<Advice>
+  val addToApiAdvice: List<Dependency>,
```
autonomousapps, line 6: is there a reason you changed from set to list?

src/main/kotlin/com/autonomousapps/internal/advice/ConsoleReport.kt
```diff
@@ -23,37 +22,37 @@ internal data class ConsoleReport(
...
+      var addToImplAdvice = emptyList<Dependency>()
+      var removeAdvice = emptyList<Dependency>()
+      var changeToApiAdvice = emptyList<Dependency>()
+      var changeToImplAdvice = emptyList<Dependency>()
+      var compileOnlyDependencies = emptyList<Dependency>()
+      var unusedProcsAdvice = emptyList<Dependency>()
 
       if (!computedAdvice.filterRemove && computedAdvice.removeAdvice.isNotEmpty()) {
-        removeAdvice = computedAdvice.removeAdvice
+        removeAdvice = computedAdvice.removeAdvice.map { it.dependency }
```
autonomousapps, line 34: fyi there's a `mapToSet` extension function available to simplify creating sets.

stephanenicolas:
Can you get a better stacktrace please for the error ?

src/main/kotlin/com/autonomousapps/internal/advice/ConsoleReport.kt
stephanenicolas: There was a reason, but I forgot wich one. Do you prefer sets ?

stephanenicolas:
And can you help with the unused compile only candidates test, please ?

autonomousapps:
@stephanenicolas 
> Can you get a better stacktrace please for the error ?

there's a build scan link in the quoted text. That has the full stacktrace.

> There was a reason, but I forgot wich one. Do you prefer sets ?

Yes.

> And can you help with the unused compile only candidates test, please ?

Did you already fix the issue where you were doing an equality assertion on two sets of different types?

autonomousapps:
I found the problem with your test. You're violating an invariant (and maybe we should guard against that). You had:
```kotlin
  private val orgDotSomethingV1 = Dependency("org.something", "1.0")
```
but you need 
```kotlin
  private val orgDotSomethingV1 = Dependency("org.something", "1.0", "implementation")
```
You've put this dependency into a component and said it's "not transitive". However, if a component/dependency is not-transitive, then it _must_ have a non-null configuration.

The compileOnly advice will _not_ suggest adding a dependency to `compileOnly` if it's transitive. At the Dependency level (cf Component), it checks this by checking if the configurationName is null. If null, the dependency is transitive.

autonomousapps:
In particular, this was resulting in an empty set:
```kotlin
  val compileOnlyDependencies: Set<Dependency> = compileOnlyCandidates
    // We want to exclude transitives here. In other words, don't advise people to declare used-transitive components.
    .filterToOrderedSet { !it.isTransitive }
    .mapToOrderedSet { it.dependency }
    // This filter returns false -- as it should! -- if the configurationName is null
    .filterToOrderedSet { it.configurationName?.endsWith("compileOnly") == false}
    .filterToOrderedSet(filterSpec.compileOnlyAdviceFilter)
```

autonomousapps:
FYI, I validated that my change fixed the `:test` task, but `functionalTest` fails. 

You can run a subset of the functional tests (latest gradle + agp, rather than the full matrix) like so
```
./gradlew functionalTest -DfuncTest.quick
```

autonomousapps:
FYI, after I merge I'm just going to delete `AdvisorTest` as not worth maintaining. We already have comprehensive coverage from the functional / end-to-end tests, and they're more meaningful with less hard-to-read stub usage.

PR #137, opened 2020-04-29
Service loader support.

autonomousapps:


autonomousapps:
All the test failures are relating to running out of metaspace.

Issue #126, opened 2020-04-22
Add a filter for unusedProcsAdvice


stephanenicolas:

Paladinium:
We would like to fail the build on unused dependencies:

`dependencyAnalysis {
	issues {
		onUnusedDependencies { fail("org.projectlombok:lombok") }
		onUsedTransitiveDependencies { ignore() }
		onIncorrectConfiguration { ignore() }
	}
}`

Even though we added an exception, the build still fails:

> Unused annotation processors that should be removed:
> - implementation("org.projectlombok:lombok:1.18.12")

@stephanenicolas I assume your change would provide a solution for this, right?

autonomousapps:
@Paladinium thanks for the comment. Yes, the change @stephanenicolas is working on should resolve this. It will add a new block to the `issues {}` block, name tbd.

Paladinium:
Thanks, @autonomousapps . I highly appreciate your and @stephanenicolas work and I am looking forward to seeing it in the next release!

autonomousapps:
This is now resolved. https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/143

PR #143, opened 2020-05-02
Add a filter for unused annotation processors.

stephanenicolas:
Solves #126 . 

There was also a little issue in a different test, I hope you're ok that I fix it here.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/advice/Advice.kt
+++ b/src/main/kotlin/com/autonomousapps/advice/Advice.kt
@@ -87,13 +87,24 @@ data class Advice(
   fun isAdd() = fromConfiguration == null && !isCompileOnly()
 
   /**
-   * An advice is "remove-advice" if it is declared and not used, AND is not `compileOnly`.
+   * An advice is "remove-advice" if it is declared and not used, AND is not `compileOnly`,
+   * AND is not `processor`.
    */
-  fun isRemove() = toConfiguration == null && !isCompileOnly()
+  fun isRemove() = toConfiguration == null && !isCompileOnly() && !isProcessor()
 
   /**
    * An advice is "change-advice" if it is declared and used (but is on the wrong configuration),
    * AND is not `compileOnly`.
    */
   fun isChange() = fromConfiguration != null && toConfiguration != null && !isCompileOnly()
+
+  /**
+   * An advice is "processors-advice" if it is declared on a k/apt
+   * or annotationProcessor configuration.
+   */
+  fun isProcessor() =
+    toConfiguration == null && fromConfiguration?.let {
+      it.endsWith("kapt", ignoreCase = true) ||
+        it.endsWith("annotationProcessor", ignoreCase = true)
+    } == true
 }
--- a/src/main/kotlin/com/autonomousapps/internal/advice/filter/FilterSpec.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/filter/FilterSpec.kt
@@ -13,6 +13,7 @@ internal class FilterSpecBuilder {
   var usedTransitivesBehavior: Behavior = Warn()
   var incorrectConfigurationsBehavior: Behavior = Warn()
   var compileOnlyBehavior: Behavior = Warn()
+  var unusedProcsBehavior: Behavior = Warn()
 
   // Filters
   var universalFilter: CompositeFilter = CompositeFilter()
@@ -29,6 +30,7 @@ internal class FilterSpecBuilder {
       usedTransitivesBehavior = usedTransitivesBehavior,
       incorrectConfigurationsBehavior = incorrectConfigurationsBehavior,
       compileOnlyBehavior = compileOnlyBehavior,
+      unusedProcsBehavior = unusedProcsBehavior,
       universalFilter = universalFilter,
       facadeFilter = facadeFilter
     )
@@ -44,6 +46,7 @@ internal class FilterSpecBuilder {
  * - [usedTransitivesBehavior] applied to add-dependencies advice; supplied by user.
  * - [incorrectConfigurationsBehavior] applied change-advice; supplied by user.
  * - [compileOnlyBehavior] applied to compileOnly-advice; supplied by user.
+ * - [unusedProcsBehavior] applied to unusedProcs-advice; supplied by user.
  */
 internal class FilterSpec(
   private val universalFilter: CompositeFilter = CompositeFilter(),
@@ -52,7 +55,8 @@ internal class FilterSpec(
   private val unusedDependenciesBehavior: Behavior = Warn(),
   private val usedTransitivesBehavior: Behavior = Warn(),
   private val incorrectConfigurationsBehavior: Behavior = Warn(),
-  private val compileOnlyBehavior: Behavior = Warn()
+  private val compileOnlyBehavior: Behavior = Warn(),
+  private val unusedProcsBehavior: Behavior = Warn()
 ) {
 
   private val shouldIgnoreAll = anyBehavior is Ignore
@@ -60,6 +64,7 @@ internal class FilterSpec(
   val filterAdd = shouldIgnoreAll || usedTransitivesBehavior is Ignore
   val filterChange = shouldIgnoreAll || incorrectConfigurationsBehavior is Ignore
   val filterCompileOnly = shouldIgnoreAll || compileOnlyBehavior is Ignore
+  val filterUnusedProcs = shouldIgnoreAll || unusedProcsBehavior is Ignore
 
   val removeAdviceFilter: (HasDependency) -> Boolean = { hasDependency ->
     val dependency = hasDependency.dependency
@@ -107,6 +112,20 @@ internal class FilterSpec(
       // If we're ignoring everything or just ignoring all compileOnly dependencies, then do that
       false
     } else if (anyBehavior.filter.plus(compileOnlyBehavior.filter).contains(dependency.identifier)) {
+      // If we're ignoring some specific dependencies, then do that
+      false
+    } else {
+      // If the dependency is universally ignored, do that
+      universalFilter.predicate(dependency)
+    }
+  }
+
+  val unusedProcsAdviceFilter: (HasDependency) -> Boolean = { hasDependency ->
+    val dependency = hasDependency.dependency
+    if (anyBehavior is Ignore || unusedProcsBehavior is Ignore) {
+      // If we're ignoring everything or just ignoring all unused procs dependencies, then do that
+      false
+    } else if (anyBehavior.filter.plus(unusedProcsBehavior.filter).contains(dependency.identifier)) {
       // If we're ignoring some specific dependencies, then do that
       false
     } else {
--- a/src/main/kotlin/com/autonomousapps/internal/models.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/models.kt
@@ -306,10 +306,10 @@ data class Method internal constructor(val types: Set<String>) {
 }
 
 data class AnnotationProcessor(
-  val dependency: Dependency,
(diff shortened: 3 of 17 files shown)


src/test/kotlin/com/autonomousapps/internal/ConsoleReportTest.kt
```diff
@@ -287,10 +289,56 @@ class ConsoleReportTest {
     val consoleReport = ConsoleReport.from(computedAdvice = computedAdvice)
 
     // Then
-    val adviceApi = consoleReport.changeToApiAdvice
-    assertTrue(adviceApi.isEmpty(), "Expected $adviceApi to be empty")
-    val adviceImpl = consoleReport.changeToImplAdvice
-    assertTrue(adviceImpl.isEmpty(), "Expected $adviceImpl to be empty")
+    val adviceCompileOnly = consoleReport.compileOnlyDependencies
```
stephanenicolas, line 292: Here is the other test that I fixed.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
```diff
@@ -87,13 +98,23 @@ data class Advice(
...
    */
   fun isChange() = fromConfiguration != null && toConfiguration != null && !isCompileOnly()
+
+  /**
+   * An advice is "processors-advice" if it is declared on a k/apt
+   * or annotationProcessor configuration.
+   */
+  fun isProcessor() = fromConfiguration != null &&
+    toConfiguration == null &&
+    (fromConfiguration.endsWith("apt", ignoreCase = true)
```
autonomousapps, line 118: I appreciate this attempt to be more general (apt vs kapt), but actually we only check for kapt currently. I've made no attempt to be compatible with the old `apt`. And I'm still not convinced we need to differentiate processor-advice from remove-advice like this.

src/main/kotlin/com/autonomousapps/internal/advice/filter/FilterSpec.kt
```diff
@@ -114,4 +119,17 @@ internal class FilterSpec(
...
+  val unusedProcsAdviceFilter: (Dependency) -> Boolean = { dependency ->
+    if (anyBehavior is Ignore || unusedProcsBehavior is Ignore) {
+      // If we're ignoring everything or just ignoring all compileOnly dependencies, then do that
+      false
+    } else if (anyBehavior.filter.plus(unusedProcsBehavior.filter).contains(dependency.identifier)) {
+      // If we're ignoring some specific dependencies, then do that
+      false
+    } else {
+      // If the dependency is universally ignored, do that
+      universalFilter.predicate(dependency)
```
autonomousapps, line 132: Latest master has changed the convention somewhat. These two lines are now generally:
```kotlin
// If the dependency is universally ignored, or is a facade we care about, ignore it
universalFilter.predicate(dependency) && facadeFilter.predicate(hasDependency)
```
although it seems highly unlike an annotation processor would trigger the facade filter. :thinking:

src/main/kotlin/com/autonomousapps/tasks/FailOrWarnTask.kt
```diff
@@ -37,13 +37,17 @@ abstract class FailOrWarnTask : DefaultTask() {
...
+  abstract val failOnUnusedProcs: Property<Behavior>
+
   @TaskAction
   fun action() {
     val adviceReports = advice.get().asFile.readText().fromJsonSet<BuildHealth>()
 
     val anyUnusedDependencies = adviceReports.any { (_, advices, _) -> advices.any { it.isRemove() } }
     val anyUsedTransitives = adviceReports.any { (_, advices, _) -> advices.any { it.isAdd() } }
     val anyIncorrectConfigurations = adviceReports.any { (_, advices, _) -> advices.any { it.isChange() } }
+    val anyUnusedProcs = adviceReports.any { (_, advices, _) -> advices.any { it.isChange() } }
```
autonomousapps, line 50: do you mean to use `it.isChange()` at the end?

stephanenicolas: No, it's a merge issue.
Thx

src/main/kotlin/com/autonomousapps/advice/Advice.kt
```diff
@@ -69,6 +69,17 @@ data class Advice(
       dependency = dependency,
       fromConfiguration = dependency.configurationName, toConfiguration = toConfiguration
     )
+
+    fun processor(dependency: Dependency) = Advice(
```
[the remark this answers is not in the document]
stephanenicolas: i/ got to review this but I am not sure to see how everything is going to work. Are you saying they are completely distinct sets of advices all the time and there is not risk to have advices being misinterpreted ?

src/main/kotlin/com/autonomousapps/internal/advice/filter/FilterSpec.kt
stephanenicolas: Let's keep the facade. Who knows ?

src/main/kotlin/com/autonomousapps/tasks/FailOrWarnTask.kt
stephanenicolas: I don't understand here how you wanna do without introducing a `isProcessor`.

src/main/kotlin/com/autonomousapps/internal/models.kt
```diff
@@ -301,10 +301,10 @@ data class Method internal constructor(val types: Set<String>) {
 }
 
 data class AnnotationProcessor(
-  val dependency: Dependency,
+  override val dependency: Dependency,
```
stephanenicolas, line 304: @autonomousapps is it the right way to do it ? `AnnotationProcessor` now implements `HasDependency`.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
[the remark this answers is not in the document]
autonomousapps: Yes, I consider it non-dangerous. Remove is remove. We distinguish dependency from annotation processor based on a set of well-known configuration names. The only other way is to add a new property to `Advice`, and I already feel like that class is bloated enough. Adding a new factory can only be redundant without a new property (which I'm against).

src/main/kotlin/com/autonomousapps/tasks/FailOrWarnTask.kt
autonomousapps: `isProcessor()` seems ok. It's the new factories for creating `Advice` instances I'm opposed to.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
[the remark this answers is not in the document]
autonomousapps: Looking at your code again, and I strongly believe we should remove this new method and just keep using the original 
```kotlin
@JvmStatic
fun remove(dependency: Dependency) = Advice(
  dependency = dependency,
  fromConfiguration = dependency.configurationName, toConfiguration = null
)
```
they're identical.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
```diff
@@ -87,13 +92,23 @@ data class Advice(
   fun isAdd() = fromConfiguration == null && !isCompileOnly()
 
   /**
-   * An advice is "remove-advice" if it is declared and not used, AND is not `compileOnly`.
+   * An advice is "remove-advice" if it is declared and not used, AND is not `compileOnly`,
+   * AND is not `processor`.
    */
-  fun isRemove() = toConfiguration == null && !isCompileOnly()
+  fun isRemove() = toConfiguration == null && !isCompileOnly() &&!isProcessor()
```
autonomousapps, line 98: This is fine. I think you're right and we need some easy way to distinguish dependency-related remove advice from processor-related remove advice. Having an `isProcessor()` method makes sense.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
```diff
@@ -87,13 +92,23 @@ data class Advice(
...
    * An advice is "change-advice" if it is declared and used (but is on the wrong configuration),
    * AND is not `compileOnly`.
    */
   fun isChange() = fromConfiguration != null && toConfiguration != null && !isCompileOnly()
+
+  /**
+   * An advice is "processors-advice" if it is declared on a k/apt
+   * or annotationProcessor configuration.
+   */
+  fun isProcessor() = fromConfiguration != null &&
```
autonomousapps, line 110: I think that `fromConfiguration != null` is redundant, given lines 112-113, where you test that `fromConfiguration` ends with kapt or annotationProcessor. You might need to do this:
```kotlin
fromConfiguration?.endsWith("kapt", ignoreCase = true) == true
```

src/main/kotlin/com/autonomousapps/internal/advice/filter/FilterSpec.kt
```diff
@@ -114,4 +119,18 @@ internal class FilterSpec(
...
+    val dependency = hasDependency.dependency
+    if (anyBehavior is Ignore || unusedProcsBehavior is Ignore) {
+      // If we're ignoring everything or just ignoring all unused procs dependencies, then do that
+      false
+    } else if (anyBehavior.filter.plus(unusedProcsBehavior.filter).contains(dependency.identifier)) {
+      // If we're ignoring some specific dependencies, then do that
+      false
+    } else {
+      // If the dependency is universally ignored, do that
+      universalFilter.predicate(dependency) && facadeFilter.predicate(hasDependency)
```
autonomousapps, line 133: I appreciate that you did this (using `facadeFilter`), but I think we should remove that. For now, at least. I can't think of a scenario where an annotation processor would reasonably be considered a "facade", as I term it. Therefore I'd rather _not_ do this and see if anyone complains and provides a good use-case.

src/main/kotlin/com/autonomousapps/internal/models.kt
autonomousapps: Yes, assuming you need to use this somewhere that requires `HasDependency`, this is the right way to do it.

I don't even mind if this stays when you remove the usage that required it (facadeFilter) -- assuming you agree with me there.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
```diff
@@ -180,6 +183,7 @@ abstract class AdviceTask : DefaultTask() {
     unusedDependenciesBehavior = failOnUnusedDependencies.get()
     usedTransitivesBehavior = failOnUsedTransitiveDependencies.get()
     incorrectConfigurationsBehavior = failOnIncorrectConfiguration.get()
+    unusedProcsBehavior= failOnUnusedProcs.get()
```
autonomousapps, line 186: nit: missing a space before the `=`.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
stephanenicolas: I didn't like the 2 `?.`, so I used a let. Let me know if it's ok for you.

src/main/kotlin/com/autonomousapps/tasks/AdviceTask.kt
stephanenicolas: we badly need ktlint..

stephanenicolas:
I am rebasing and addressing comments. I wanna get another build on CI to debug the failing tests.

stephanenicolas:
@autonomousapps when we get a green build, you should update the [wiki](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Customizing-plugin-behavior)

PR #152, opened 2020-05-08
Publish a snapshot on each PR.

autonomousapps:


stephanenicolas:
It's a very good idea. But how are we gonna deal with version numbers ? Is it gonna be like `0.40-branch-SNAPSHOT` ?

autonomousapps:
I thought I'd keep it simple, at least to start. Snapshots aren't expected
to be identical each time you retrieve them, and Gradle even treats them
special, so I thought I'd just publish with a snapshot suffix and that
would be good enough.

stephanenicolas:
I think it would be better to do this when master builds. 
For PRs, you need to distinguish the branches.

autonomousapps:
Even if we only published on master, we'd still have to deal with the version number issue. Branch / no branch doesn't matter. Either way, we're overwriting the snapshot over and over again. And I think this is fine, at least to start. Snapshots are volatile by nature.

autonomousapps:
Although now that I think about it more, I think you're right we should only publish from master. Otherwise different PRs will just overwrite each other, which is weird. Thanks for pushing.

PR #158, opened 2020-05-11
Support ABI exclusions

### LINKED ISSUE #48: Don't recommend adding direct dependencies if they're only used by generated code?

ZacSweers:
This is intended to resolve #48 by adding a new API to extensions for `abi` handling and `exclusions` configuration.

The API in practice looks like this:

```kotlin
abi {
  exclusions {
    // Convenience helpers
    ignoreSubPackage("internal")
    ignoreInternalPackages()
    ignoreGeneratedCode()

    // Raw, regexp-based APIs
    excludeAnnotations(".*\\.Generated")
    excludeClasses(".*\\.internal\\..*")
  }
}
```

I've left toe-holds in for directory based parsing (i.e. `build/`), but that will take more work as the current code looks only at class files and not where they originated from.

No tests yet as I want to put this up for review of the design before moving further.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/analyzer/DependencyAnalyzer.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/analyzer/DependencyAnalyzer.kt
@@ -71,7 +71,8 @@ internal interface DependencyAnalyzer<T : ClassAnalysisTask> {
    * no meaningful ABI.
    */
   fun registerAbiAnalysisTask(
-    dependencyReportTask: TaskProvider<DependencyReportTask>
+    dependencyReportTask: TaskProvider<DependencyReportTask>,
+    abiExclusions: Provider<String>
   ): TaskProvider<AbiAnalysisTask>? = null
 }
 
--- a/src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
@@ -7,6 +7,7 @@
 
 package com.autonomousapps.internal.kotlin
 
+import com.autonomousapps.internal.AbiExclusions
 import com.autonomousapps.internal.asm.ClassReader
 import com.autonomousapps.internal.asm.Opcodes
 import com.autonomousapps.internal.asm.tree.ClassNode
@@ -57,8 +58,21 @@ fun getBinaryAPI(classStreams: Sequence<InputStream>, visibilityFilter: (String)
             it.isEffectivelyPublic(classAccess, mVisibility)
           }
 
-          ClassBinarySignature(name, superName, outerClassName, supertypes, memberSignatures, classAccess,
-              isEffectivelyPublic(mVisibility), metadata.isFileOrMultipartFacade() || isDefaultImpls(metadata)
+          val annotations = (visibleAnnotations.orEmpty() + invisibleAnnotations.orEmpty())
+            .map { it.desc.replace("/", ".") }
+
+          ClassBinarySignature(
+            name = name,
+            superName = superName,
+            outerName = outerClassName,
+            supertypes = supertypes,
+            memberSignatures = memberSignatures,
+            access = classAccess,
+            isEffectivelyPublic = isEffectivelyPublic(mVisibility),
+            isNotUsedWhenEmpty = metadata.isFileOrMultipartFacade() || isDefaultImpls(metadata),
+            annotations = annotations,
+            // TODO toe-hold for filtering by directory
+            sourceFileLocation = null
           )
         }
       }
@@ -67,12 +81,18 @@ fun getBinaryAPI(classStreams: Sequence<InputStream>, visibilityFilter: (String)
 }
 
 
-fun List<ClassBinarySignature>.filterOutNonPublic(nonPublicPackages: List<String> = emptyList()): List<ClassBinarySignature> {
-  val nonPublicPaths = nonPublicPackages.map { it.replace('.', '/') + '/' }
+internal fun List<ClassBinarySignature>.filterOutNonPublic(
+  exclusions: AbiExclusions = AbiExclusions.NONE
+): List<ClassBinarySignature> {
   val classByName = associateBy { it.name }
 
-  fun ClassBinarySignature.isInNonPublicPackage() =
-      nonPublicPaths.any { name.startsWith(it) }
+  // Library note - this function (plus the exclusions parameter above) are modified from the original
+  // Kotlin sources this was borrowed from.
+  fun ClassBinarySignature.isExcluded(): Boolean {
+    return (sourceFileLocation?.let(exclusions::excludesPath) ?: false) ||
+      exclusions.excludesClass(canonicalName) ||
+      annotations.any(exclusions::excludesAnnotation)
+  }
 
   fun ClassBinarySignature.isPublicAndAccessible(): Boolean =
       isEffectivelyPublic &&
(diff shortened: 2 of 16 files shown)


src/main/kotlin/com/autonomousapps/internal/analyzer/DependencyAnalyzer.kt
```diff
@@ -70,7 +70,8 @@ internal interface DependencyAnalyzer<T : ClassAnalysisTask> {
    * This is a no-op for com.android.application projects, since they have no meaningful ABI.
    */
   fun registerAbiAnalysisTask(
-    dependencyReportTask: TaskProvider<DependencyReportTask>
+    dependencyReportTask: TaskProvider<DependencyReportTask>,
+    abiExclusions: Provider<String>
```
autonomousapps, line 74: I just added support for the java `application` plugin, and so this has changed somewhat on master. FYI.

src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
```diff
@@ -57,8 +58,20 @@ fun getBinaryAPI(classStreams: Sequence<InputStream>, visibilityFilter: (String)
             it.isEffectivelyPublic(classAccess, mVisibility)
           }
 
-          ClassBinarySignature(name, superName, outerClassName, supertypes, memberSignatures, classAccess,
-              isEffectivelyPublic(mVisibility), metadata.isFileOrMultipartFacade() || isDefaultImpls(metadata)
+          val annotations = visibleAnnotations.orEmpty().map { it.desc.replace("/", ".") }
+
+          ClassBinarySignature(
```
autonomousapps, line 63: So, the reason I didn't format this file to project standards is that it's basically copy-pasted from JetBrains (as you can see from the top of the file). So I'm a little hesitant to make too many changes and lose the ability to update from the original. Admittedly an annoying situation.

I don't know a better way to do this, so I guess I just like talking to myself.

src/main/kotlin/com/autonomousapps/internal/models.kt
```diff
@@ -416,3 +417,29 @@ internal data class ConsoleReport(
     }
   }
 }
+
+data class AbiExclusions(
```
autonomousapps, line 421: should be `internal`.

src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
```diff
@@ -25,6 +25,10 @@ internal inline fun <T, R> Iterable<T>.mapToSet(transform: (T) -> R): HashSet<R>
   return mapTo(HashSet(collectionSizeOrDefault(10)), transform)
 }
 
+internal inline fun <T, R> Iterable<T>.mapToLinkedHashSet(transform: (T) -> R): HashSet<R> {
```
autonomousapps, line 28: Do you think we should change the pre-existing `maptToSet()` function to use a `LinkedHashSet` instead of adding a new method? Or, why do you prefer a linked hash set specifically here?

src/main/kotlin/com/autonomousapps/tasks/AbiAnalysisTask.kt
```diff
@@ -66,6 +74,9 @@ abstract class AbiAnalysisWorkAction : WorkAction<AbiAnalysisParameters> {
     // Inputs
     val jarFile = parameters.jar.get().asFile
     val dependencies = parameters.dependencies.get().asFile.readText().fromJsonList<Component>()
+    val exclusions = parameters.exclusions.takeIf { it.isPresent }?.run {
```
autonomousapps, line 77: I have a preference for
```kotlin
val exclusions = parameters.exclusions.orNull?.run {
  fromJson<AbiExclusions>()
} ?: AbiExclusions.NONE
```
to match the existing pattern. It's also a _tiny_ bit less verbose.

src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
ZacSweers: Hmm, I could just leave a comment explaining it? I like region comments too to isolate fork changes

src/main/kotlin/com/autonomousapps/internal/models.kt
ZacSweers: I did at first but it was a bit more invasive than I expected due to being used in constructor params later. I'll do it in an isolated commit and can revert if we want

src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
ZacSweers: I'd be down for that, as that would match kotlin's conventional behavior as well (`toSet()` guarantees preserved order)

ZacSweers: As for why LinkedHashSet - https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

src/main/kotlin/com/autonomousapps/tasks/AbiAnalysisTask.kt
ZacSweers: TIL orNull!

ZacSweers: 0e501e0

src/main/kotlin/com/autonomousapps/internal/models.kt
ZacSweers: Not as bad as I thought actually 354351c

src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
ZacSweers: a3c0509

src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
ZacSweers: a6ece6b

src/main/kotlin/com/autonomousapps/internal/analyzer/DependencyAnalyzer.kt
ZacSweers: Fixed in my rebase 👍

ZacSweers:
Addressed the inline comments, going to work on adding a test(s) now

ZacSweers:
I wrote a pretty good unit test for the ABI checking. I can still write a functional test, but at this point it would be just to verify gradle plumbing rather than functionally correct exclusion behavior. Can take a crack if you think it's worth it, let me know

autonomousapps:
I want to talk a bit about the impact this has on generated advice.

Without setting any exclusions and running buildHealth on a personal project, I will see this in the abi-dump.txt file:
```
public abstract interface class com/seattleshelter/core/api/PetsService {
	public abstract fun getCatById (J)Lio/reactivex/Single;
	public abstract fun getCatsList ()Lio/reactivex/Single;
        ...etc...
}
```
and furthermore, the advice for this project suggests adding RxJava as an `api` dependency (as we would expect).

If I add the following:
```kotlin
abi {
  exclusions {
    excludeClasses("com\\.seattleshelter\\.core\\.api\\.PetsService")
  }
}
```
then abi dump does not contain that interface any longer, and the advice no longer suggests adding RxJava2 as an `api` dependency, since it's no longer part of the published api, as it was only used by the now-excluded class.

Is this what you expected? If not, could you be precise with what your goal is?

One interesting thing about this is that this plugin deals primarily in resolved dependencies. This new DSL operates at a different level (that of classes), which has an indirect impact on the dependency advice.

ZacSweers:
Right, in this case that's expected. The idea is that you can limit exposed APIs for things that are either optional or public-only-becuse-tey-have-to-be

PR #171, opened 2020-05-18
Refactor test suite

autonomousapps:
Typically massive PR where I decide it would be "neat" to do something and keep going till I've lost a weekend.


autonomousapps:
Build failed

Issue #172, opened 2020-05-18
Add a filter for compileOnly


stephanenicolas:
Similar to  #126, but for compileOnly related issues.

It looks like we, at least miss an extension entry point and some glue code in:
`com.autonomousapps.tasks.AdviceTask#filterSpecBuilder`.

autonomousapps:
This is now resolved.

PR #175, opened 2020-05-19
Filter for compile only dependencies via DSL

stephanenicolas:
Solves issue #172.
Ignore/warn/fails on errors related to dependencies that "could be compile only".
Adds an issue handler method to the DSL:
```
onAdvisedCompileOnly {
  fail(<"foo">)
}
```

@autonomousapps , we need to update the wiki after this is merged.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/android/AdviceSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/AdviceSpec.groovy
@@ -18,23 +18,27 @@ final class AdviceSpec extends AbstractAndroidSpec {
       dependencyAnalysis {
         issues {
           onAny {
-              severity 'fail'
-              exclude '$KOTLIN_STDLIB_JDK7_ID'
+            severity 'fail'
+            exclude '$KOTLIN_STDLIB_JDK7_ID'
           }
           onUnusedDependencies {
-              severity 'fail'
-              exclude ':lib_android'
+            severity 'fail'
+            exclude ':lib_android'
           }
           onUsedTransitiveDependencies {
-              severity 'fail'
-              exclude '$CORE_ID'
+            severity 'fail'
+            exclude '$CORE_ID'
           }
           onIncorrectConfiguration {
-              severity 'fail'
-              exclude '$COMMONS_COLLECTIONS_ID'
+            severity 'fail'
+            exclude '$COMMONS_COLLECTIONS_ID'
+          }
+          onCompileOnly {
+            severity 'fail'
+            exclude '$ANDROIDX_ANNOTATIONS_ID'
           }
           onUnusedAnnotationProcessors {
-              exclude 'fail'
+            severity 'fail'
           }
         }
         setFacadeGroups()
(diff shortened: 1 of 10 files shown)


src/functionalTest/groovy/com/autonomousapps/android/AdviceSpec.groovy
```diff
@@ -32,6 +32,9 @@ final class AdviceSpec extends AbstractAndroidSpec {
           onUnusedAnnotationProcessors {
               fail()
           }
+          onAdvisedCompileOnly {
```
autonomousapps, line 35: I think `onCompileOnly` is good enough.

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
```diff
@@ -199,6 +199,7 @@ open class IssueHandler @Inject constructor(objects: ObjectFactory) {
   internal val usedTransitiveDependenciesIssue = objects.newInstance(Issue::class.java)
   internal val incorrectConfigurationIssue = objects.newInstance(Issue::class.java)
   internal val unusedAnnotationProcessorsIssue = objects.newInstance(Issue::class.java)
+  internal val advisedCompileOnlyIssue = objects.newInstance(Issue::class.java)
```
autonomousapps, line 202: `compileOnlyIssue`

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -357,6 +357,7 @@ class DependencyAnalysisPlugin : Plugin<Project> {
         failOnUsedTransitiveDependencies.set(usedTransitiveDependenciesIssue.behavior)
         failOnIncorrectConfiguration.set(incorrectConfigurationIssue.behavior)
         failOnUnusedProcs.set(unusedAnnotationProcessorsIssue.behavior)
+        failOnAdvisedCompileOnly.set(advisedCompileOnlyIssue.behavior)
```
autonomousapps, line 360: `failOnCompileOnly` :)

src/main/kotlin/com/autonomousapps/advice/Advice.kt
```diff
@@ -69,6 +69,11 @@ data class Advice(
       dependency = dependency,
       fromConfiguration = dependency.configurationName, toConfiguration = toConfiguration
     )
+
+    @JvmStatic
+    fun ofCompileOnly(dependency: Dependency) = ofCompileOnly(
```
autonomousapps, line 74: Is this really needed? I guess it's a bit more convenient, but I don't want to bake in the assumption that we only advice `compileOnly` vs `debugCompileOnly`, etc. So I think it's just easier in the long-run to keep using `ofChange()`

src/functionalTest/groovy/com/autonomousapps/android/AdviceSpec.groovy
autonomousapps: And as far as the test is concerned, I think that you want two compileOnly dependencies on the classpath. Then you can have this:
```groovy
onCompileOnly {
  fail("...") // one of the compileOnly identifiers
}
```
and finally, the test assertion should be that the list of compileOnly deps that are in the advice should not include the one being here filtered.

src/main/kotlin/com/autonomousapps/advice/Advice.kt
stephanenicolas: To be clear, you want me to remove both:
```
   @JvmStatic
    fun ofCompileOnly(dependency: Dependency, toConfiguration: String) = Advice(
      dependency = dependency,
      fromConfiguration = dependency.configurationName, toConfiguration = toConfiguration
    )

    @JvmStatic
    fun ofCompileOnly(dependency: Dependency) = ofCompileOnly(
      dependency = dependency, toConfiguration = "compileOnly"
    )
```
Right ?

autonomousapps: yes

src/functionalTest/groovy/com/autonomousapps/android/AdviceSpec.groovy
```diff
@@ -32,6 +32,9 @@ final class AdviceSpec extends AbstractAndroidSpec {
           onUnusedAnnotationProcessors {
               fail()
           }
+          onCompileOnly {
+              fail()
```
autonomousapps, line 36: this is missing a parameter value for "ignoring" some compileOnly dep, to fully test the new extension.

PR #182, opened 2020-05-27
Coalesce all variant-specific dependency advice with plugin advice at the project level.

autonomousapps:
Pretty sure there's more old cruft I can delete.


autonomousapps:
Build failed

PR #183, opened 2020-05-28
Added registerPostProcessingTask() to the DSL.

autonomousapps:


src/main/kotlin/com/autonomousapps/AbstractPostProcessingTask.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.autonomousapps.internal.utils.fromJson
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+
+abstract class AbstractPostProcessingTask : DefaultTask() {
+
+  @get:PathSensitive(PathSensitivity.NONE)
```
ZacSweers, line 15: Why none?

autonomousapps:
Only the content of the file matters. The path to it and name of it are
irrelevant.

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -677,6 +677,20 @@ class DependencyAnalysisPlugin : Plugin<Project> {
         outputPretty.set(paths.aggregateAdvicePrettyPath)
       }
 
+    // Is there a post-processing task? If so, run it
+    afterEvaluate {
+      val postProcessingTask = if (this == rootProject) {
```
sboishtyan, line 682: Could we replace `afterEvaluate` if we used Gradle property or provider for postProcessingTask?

autonomousapps: I would have liked to do that, but I don't know if a `TaskProvider<Something>` can be held in a property.

There's also the issue that I do this wiring at configuration time and not task execution time, so I have to "wait" (via `afterEvaluate`) for the value to be set on the extension before I can query it.

PR #201, opened 2020-06-14
Support star imports for finding annotation processors.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/188. (except that it doesn't handle when the jar task has been disabled)


autonomousapps:
Build failed

PR #206, opened 2020-06-24
Disable plugin if invoked from Android Studio.

autonomousapps:


src/main/kotlin/com/autonomousapps/tasks/AdviceSubprojectAggregationTask.kt
```diff
@@ -216,3 +220,8 @@ abstract class AdviceSubprojectAggregationTask : DefaultTask() {
       "\"${dependency.identifier}:${dependency.resolvedVersion}\""
     }
 }
+
+private fun shouldNotBeSilent(): Boolean {
+  val silent = System.getProperty("dependency.analysis.silent", "false")
```
sboishtyan, line 225: Thank you

Issue #207, opened 2020-06-26
DAAGP Should support null unit test (and android tests) variants


stephanenicolas:
Repro on DAAGP 0.50.0.

Since AGP 4.0, it is possible for an android module to disable completely (filter) its tests variants when it has no such tests. this enables to speed up AGP, especially during the sync with AS.

Here is an example: https://github.com/android/gradle-recipes/blob/master/BuildSrc/testVariantFilteringOnBuildType/buildSrc/src/main/kotlin/CustomPlugin.kt

When using DAAGP on a module with unit test variants disabled, we got the stack trace:
```
Caused by: java.lang.IllegalStateException: unitTestVariant must not be null
        at com.autonomousapps.DependencyAnalysisPlugin$configureAndroidLibProject$1$1.execute(DependencyAnalysisPlugin.kt:204)
        at com.autonomousapps.DependencyAnalysisPlugin$configureAndroidLibProject$1$1.execute(DependencyAnalysisPlugin.kt:44)
```

PR #208, opened 2020-06-26
Solves #207

stephanenicolas:
AGP 4.0+ allow test variant filtering.
See example here:
https://github.com/android/gradle-recipes/blob/master/BuildSrc/testVariantFilteringOnBuildType/buildSrc/src/main/kotlin/CustomPlugin.kt

The plugin should support that some modules might have a null variant.


stephanenicolas:
PR was tested on a real project and seems to work fine.

PR #214, opened 2020-06-29
Fixed issue with incorrectly assigning the test variant to a dependency.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/211.

Basically was relying on set logic when I needed a merged map.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/DependencyMisuseTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/DependencyMisuseTask.kt
@@ -135,7 +135,7 @@ internal class MisusedDependencyDetector(
     val unusedDeps = mutableListOf<Dependency>()
     val usedTransitiveComponents = mutableSetOf<TransitiveComponent>()
     val usedDirectClasses = mutableSetOf<String>()
-    val usedDependencies = mutableSetOf<VariantDependency>()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/autonomousapps/tasks/DependencyMisuseTask.kt
```diff
@@ -135,7 +135,7 @@ internal class MisusedDependencyDetector(
     val unusedDeps = mutableListOf<Dependency>()
     val usedTransitiveComponents = mutableSetOf<TransitiveComponent>()
     val usedDirectClasses = mutableSetOf<String>()
-    val usedDependencies = mutableSetOf<VariantDependency>()
+    val usedDependencies = mutableMapOf<Dependency, MutableSet<String>>()
```
sboishtyan, line 138: So the problem was because of incorrect merging main and test sources?

autonomousapps: More or less. I don't have the intermediary files any more, but it was something like this:
```
dependency X provides class A that is only used in variant "test"
dependency X provides class B used by variants "main", test"
```
and the old `Set` contained both of those elements, in that order (i.e., some of your classes from that one dependency is only used by test source, while other classes are used in both test and main sources). By changing the collection type from a set to a map, there will never be more than one entry for "dependency X".

PR #225, opened 2020-07-20
Reduced the number of times we read the same bytestream (from a jar) from 3 to 2.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/216.

The DependencyReportTask now produced a even more gargantuan 'Component' with data needed to also find constants, so that the latter does not need the additional ByteArray.

I expect this should measurably reduce the number of allocations and therefore memory consumption. It should also mean less CPU usage, since we don't have to visit the same bytecode twice.

It may be possible to reduce this further, but that can be done (if ever) separately.


autonomousapps:
Build failed https://gradle.com/s/vhh22w5sgc7ww

PR #233, opened 2020-07-23
Use more lazy APIs for tasks and files

stephanenicolas:
In this PR I try to:
* use more task providers APIs instead of using `get()` on them. Even though the target classes have probably been realized already at this point, it is a bit cleaner.
* use more lazy file collections instead of files to limit CPU usage during the configuration phase.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
@@ -447,7 +447,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       finalizedBy(failOrWarn)
 
       doLast {
-        logger.debug("Advice report (aggregated)  : ${adviceReport.get().projectReport.get().asFile.path}")
+        val reportPath = adviceReport.flatMap { it.projectReport.map { it.asFile.path } }
+        logger.debug("Advice report (aggregated)  : $reportPath")
       }
     }
 
--- a/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
@@ -12,6 +12,7 @@ import org.gradle.api.Project
 import org.gradle.api.UnknownDomainObjectException
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.attributes.Attribute
+import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileTree
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.SourceSet
@@ -19,7 +20,6 @@ import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.bundling.Jar
 import org.gradle.kotlin.dsl.get
 import org.gradle.kotlin.dsl.register
-import java.io.File
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet as JbKotlinSourceSet
 
 internal abstract class JvmAnalyzer(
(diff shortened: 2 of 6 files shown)


src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
```diff
@@ -51,41 +51,43 @@ internal abstract class JvmAnalyzer(
...
-  private val variantFiles: Set<VariantFile> by lazy(mode = LazyThreadSafetyMode.NONE) {
+  private val variantFiles: Provider<Set<VariantFile>> by lazy(mode = LazyThreadSafetyMode.NONE) {
     val mainVariantFiles = mainSourceSet.asFiles(project).toVariantFiles("main")
-    val testVariantFiles = testSourceSet?.asFiles(project)?.toVariantFiles("test") ?: emptySet()
+    val testVariantFiles = testSourceSet?.asFiles(project)?.toVariantFiles("test")
+      ?: project.providers.provider { emptySet() }
 
     // return
-    mainVariantFiles + testVariantFiles
+    project.providers.provider {mainVariantFiles.get() + testVariantFiles.get()}
```
stephanenicolas, line 60: There is probably something better to do here.

src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
```diff
@@ -51,41 +51,43 @@ internal abstract class JvmAnalyzer(
...
-      // remove file extension from end
-      .mapNotNull {
-        val index = it.lastIndexOf(".")
-        if (index != -1) {
-          it.substring(0, index)
-        } else {
-          // This could happen if the directory were empty, (eg `src/main/java/` with nothing in it)
-          null
+  private fun FileCollection.toVariantFiles(name: String): Provider<Set<VariantFile>> {
+    return elements.map { set ->
```
autonomousapps, line 64: note to self: look at FileCollection API.

src/main/kotlin/com/autonomousapps/tasks/AbiAnalysisTask.kt
```diff
@@ -127,7 +127,7 @@ abstract class AbiAnalysisWorkAction : WorkAction<AbiAnalysisParameters> {
   }
 
   private fun allClassFiles(): Set<File> =
-    parameters.javaClasses.asFileTree.files.plus(parameters.kotlinClasses.asFileTree.files)
+    parameters.javaClasses.plus(parameters.kotlinClasses)
```
autonomousapps, line 130: 🙇‍♂️

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -71,13 +73,15 @@ abstract class ClassAnalysisTask(private val objects: ObjectFactory) : DefaultTa
...
   }
+}
 
-  @Internal
-  protected fun getTestFiles(): Set<File> {
-    val testJavaClasses = testJavaClassesDir.orNull?.asFileTree?.files ?: emptySet()
-    val testKtClasses = testKotlinClassesDir.orNull?.asFileTree?.files ?: emptySet()
-    return testJavaClasses + testKtClasses
-  }
+fun getTestFiles(
```
autonomousapps, line 78: why is this
1. not `protected` and
2. not `@Internal`
?

autonomousapps: note to self: why does it have parameters now?

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -99,18 +103,19 @@ abstract class JarAnalysisTask @Inject constructor(
 
   @TaskAction fun action() {
     // Output
-    val reportFile = output.getAndDelete()
-    val reportPrettyFile = outputPretty.getAndDelete()
+    val reportFile = output
```
autonomousapps, line 106: the reason I delete at the very start is in case there's some error that occurs during task execution. I want to start each task action by deleting its prior output (if it exists). I want to avoid the potential for erroneous or partial output.

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -99,18 +103,19 @@ abstract class JarAnalysisTask @Inject constructor(
...
     workerExecutor.noIsolation().submit(JarAnalysisWorkAction::class.java) {
       variantFiles.set(this@JarAnalysisTask.variantFiles)
       jar = jarFile
-      kaptJavaSource = kaptJavaStubs.files
-      layouts = layoutFiles.files
-      testFiles = getTestFiles()
+      kaptJavaSource = kaptJavaStubs
+      layouts = layoutFiles
+      testJavaClassesDir = this@JarAnalysisTask.testJavaClassesDir
+      testKotlinClassesDir = this@JarAnalysisTask.testKotlinClassesDir
```
autonomousapps, line 118: why do we need to send them in separately?

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -119,31 +124,33 @@ abstract class JarAnalysisTask @Inject constructor(
...
+  var report: RegularFileProperty
+  var reportPretty: RegularFileProperty
 }
 
 abstract class JarAnalysisWorkAction : WorkAction<JarAnalysisParameters> {
 
   private val logger = getLogger<JarAnalysisTask>()
 
   override fun execute() {
+    val testFiles = getTestFiles(parameters.testJavaClassesDir, parameters.testKotlinClassesDir)
```
autonomousapps, line 141: why is it important to wait until we're in the work action to do this? And if it _is_ more correct to do it here, probably make the function available somewhere else (besides the task class)?

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -119,31 +124,33 @@ abstract class JarAnalysisTask @Inject constructor(
...
-      kaptJavaSource = parameters.kaptJavaSource
+      jarFile = parameters.jar.get().asFile,
+      layouts = parameters.layouts.files,
+      testFiles = testFiles,
+      kaptJavaSource = parameters.kaptJavaSource.files
     ).analyze()
 
-    parameters.report.writeText(classNames.toJson())
-    parameters.reportPretty.writeText(classNames.toPrettyString())
+    parameters.report.getAndDelete().writeText(classNames.toJson())
```
autonomousapps, line 150: as noted above, I believe it is important to delete at the very beginning of the execution. I would also accept doing it at the beginning of the worker action (for simple tasks that only delegate to workers and doing nothing else). But I feel strongly it shouldn't happen at the _end_.

src/main/kotlin/com/autonomousapps/tasks/classAnalysisTasks.kt
```diff
@@ -177,54 +184,66 @@ abstract class ClassListAnalysisTask @Inject constructor(
 
   @TaskAction fun action() {
     // Output
-    val reportFile = output.getAndDelete()
-    val reportPrettyFile = outputPretty.getAndDelete()
+    val reportFile = output
+    val reportPrettyFile = outputPretty
+
+    val inputClassFiles = project.objects.setProperty(FileSystemLocation::class.java)
```
autonomousapps, line 190: what's up with this? I think we shouldn't do this because, at the very least, calling `projects` definitely breaks the configuration cache contract. If we need a SetProperty as an input, we can do it at the class level.

autonomousapps:
Build scan for failed build https://gradle.com/s/fufuxc3l37sx4

src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -447,7 +447,8 @@ class DependencyAnalysisPlugin : Plugin<Project> {
       finalizedBy(failOrWarn)
 
       doLast {
-        logger.debug("Advice report (aggregated)  : ${adviceReport.get().projectReport.get().asFile.path}")
+        val reportPath = adviceReport.flatMap { it.projectReport.map { it.asFile.path } }
```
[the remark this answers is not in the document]
vlsi: In cases like this it makes sense to add `if (logger.isDebugEnabled())` or something like that to avoid computations when debug is disabled (which is the most common case)

PR #238, opened 2020-07-28
Interrogate consumer proguard files for rules.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/227.

TODO:
1. Add a test
2. Rename `ProguardClasses` to something not dumb
3. Resolve TODOs.

and as a follow-up:

4. Consider finding a way to make the "metadata" more discoverable. For instance, it would be nice to know _why_ a dependency was not marked unused. We could say "this dependency provides consumer proguard rules and so, out of an abundance of caution, we are saying keep it." Another complementary option would be to provide some kind of DSL configuration.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -84,7 +84,7 @@ dependencies {
     because("Auto-wiring into Kotlin projects")
   }
 
-  testImplementation("org.spockframework:spock-core:2.0-M3-groovy-2.5") {//1.3-groovy-2.5//2.0-M3-groovy-2.5//2.0-M3-groovy-3.0
+  testImplementation("org.spockframework:spock-core:1.3-groovy-2.5") {//1.3-groovy-2.5//2.0-M3-groovy-2.5//2.0-M3-groovy-3.0
     exclude(module = "groovy-all")
     because("For Spock tests")
   }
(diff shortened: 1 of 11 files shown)


autonomousapps:
Build failed https://gradle.com/s/omx63tw6vj2rk

autonomousapps:
Example failed test: https://scans.gradle.com/s/omx63tw6vj2rk/tests/:functionalTest/com.autonomousapps.android.AnnotationProcessorSpec/dagger%20is%20used%20with%20annotationProcessor%20on%20method%20(Gradle%206.5.1%20AGP%204.1.0-beta01)?anchor=e30&focused-exception-line=0-5#1

This is not unexpected. A consequence of whitelisting all android libs that supply proguard rules is now a lot of things get filtered out, even if they're truly unused. I don't like it so I need to consider how to finesse the analysis. Maybe look for usages of reflection?

autonomousapps:
My idea with looking for reflective use has grown more complex. The actual reflection is done in the `CameraX` class, which is provided by the `androidx.camera:core` dependency, which is a transitive dependency of `androidx.camera:camera-camera2`. So I'd have to look in not only the direct dependencies, but the transitive dependencies, and correlate them back up the graph.

PR #246, opened 2020-08-07
Remove aggregate advice from archives configuration.

autonomousapps:
This ensure it doesn't get hooked into the assemble task, which is bad news.

Finally resolves the issue whereby users executing `assemble` (or some variants thereof) would also have this plugin's artifacts assembled, which was normally not desirable. Per Gradle, this current automagic behavior is deprecated and will eventually be removed (along with the old "archives" configuration).


autonomousapps:
Build failed

PR #247, opened 2020-08-09
Don't analyze module-info.class files.

autonomousapps:
Maybe resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/239.


autonomousapps:
Job canceled for mysterious reasons.

PR #249, opened 2020-08-11
Lookup Kotlin classes compiled by the multiplatform plugin

martinbonnin:
For KMP, the KotlinCompile task is named `compileKotlinJvm` instead of `compileKotlin`. 

On `apollo-android:apollo-api`, this changes the output from:

```
- null("org.jetbrains.kotlin:kotlin-stdlib:1.3.72")
Transitively used dependencies that should be declared directly as indicated:
- null("org.jetbrains.kotlin:kotlin-stdlib:1.3.72"
```

to 

```
> Task :apollo-api:aggregateAdvice
Transitively used dependencies that should be declared directly as indicated:
- api("org.jetbrains.kotlin:kotlin-stdlib:1.3.72")
- api("com.squareup.okio:okio:2.7.0")
```

### DIFF
--- a/testkit/build.gradle.kts
+++ b/testkit/build.gradle.kts
@@ -66,7 +66,7 @@ publishing {
 tasks.withType<Sign>().configureEach {
   onlyIf {
     val isNotSnapshot = !version.toString().endsWith("SNAPSHOT")
-    isNotSnapshot //&& gradle.taskGraph.hasTask(publishToMavenCentral.get())
+    isNotSnapshot && (findProperty("signing.keyId") != null)
   }
 }
 
--- a/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
@@ -172,6 +172,7 @@ internal abstract class JvmAnalyzer(
   protected fun javaCompileTask() = project.tasks.namedOrNull("compileJava")
 
   protected fun kotlinCompileTask() = project.tasks.namedOrNull("compileKotlin")
+    ?: project.tasks.namedOrNull("compileKotlinJvm") // for multiplatform projects
 
   protected fun getJarTask(): TaskProvider<Jar> = project.tasks.named(mainSourceSet.jarTaskName, Jar::class.java)
 


testkit/build.gradle.kts
```diff
@@ -66,7 +66,7 @@ publishing {
 tasks.withType<Sign>().configureEach {
   onlyIf {
     val isNotSnapshot = !version.toString().endsWith("SNAPSHOT")
-    isNotSnapshot //&& gradle.taskGraph.hasTask(publishToMavenCentral.get())
+    isNotSnapshot && (findProperty("signing.keyId") != null)
```
martinbonnin, line 69: This is so that I could call `./gradlew publishToMavenLocal` without having to sign artifacts but I can remove it if needed.

autonomousapps: My inclination is that all non-snapshot builds should be signed. Another way to resolve your problem would have been to change the version [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/testkit/build.gradle.kts#L15) to have the `-SNAPSHOT` suffix.

I'll think on this a bit. There's probably something I can do with the script configuration to make it easier for others to contribute (and it may or may not be your proposal here).

martinbonnin: I considered changing the version but I don't like having local changes, I always end up comiting them in source control...
Another option is for me to configure my signing info and sign with my key. For snapshots staying on my laptop, it shouldn't harm. I can do that as well.

autonomousapps:
This is a problem that should be solved. I strive to make it relatively simple for people to contribute. I even have a guide to help!
GitHub <notifications@github.com> wrote:
“@martinbonnin commented on this pull request.”


“In testkit/build.gradle.kts:”

> @@ -66,7 +66,7 @@ publishing {
 tasks.withType<Sign>().configureEach {
   onlyIf {
     val isNotSnapshot = !version.toString().endsWith("SNAPSHOT")
-    isNotSnapshot //&& gradle.taskGraph.hasTask(publishToMavenCentral.get())
+    isNotSnapshot && (findProperty("signing.keyId") != null)
“I considered changing the version but I don't like having local changes, I always end up comiting them in source control...
 Another option is for me to configure my signing info and sign with my key. For snapshots staying on my laptop, it shouldn't harm. I can do that as well.”

autonomousapps:
Sorry for the delay on this. I do want to get it across the finish line, but I've been distracted by other things.

autonomousapps:
Ugh, github actions. I have no visibility into why it's failing. Will have to run locally.

martinbonnin:
> Ugh, github actions. I have no visibility into why it's failing. Will have to run locally.

Do you mean it would be nice to add tests for multiplatform cases? I can try to add some, they should run an an ubuntu VM/

Without specific tests, the Actions seemed to pass on this branch: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/runs/1039954701?check_suite_focus=true

autonomousapps:
It would be good to have a test, yes. But what's annoying me right now is that CI is failing for no clear reason. My test suite is pretty intense and I've had issues with GHA for a while now. Could be OOMing. 
GitHub <notifications@github.com> wrote:
““Ugh, github actions. I have no visibility into why it's failing. Will have to run locally.”

Do you mean it would be nice to add tests for multiplatform cases? I can try to add some, they should run an an ubuntu VM/

Without specific tests, the Actions seemed to pass on this branch: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/runs/1039954701?check_suite_focus=true”

autonomousapps:
Yay, CI is green.

@martinbonnin could you provide a minimal reproducer sample somewhere so I can add it to my functional test suite?

autonomousapps:
I have rebased this locally and merged to `main`. Closing. Thanks for the contribution! (would still like to add a test)

PR #259, opened 2020-08-28
Detect AARs with bundled native libs.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/252.

This is not quite done yet (I'd like to add a test).


autonomousapps:
Build failed

PR #261, opened 2020-08-31
Emit warning when a dependency is on multiple configurations.

autonomousapps:
Partially resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/257. This is likely to change based on response to [this question](https://gradle-community.slack.com/archives/CAH4ZP3GX/p1598847778029900) in the gradle slack.

nb I suspect I can make the algorithm a tad more efficient.


autonomousapps:
This now fully resolves the linked issue. Still not happy with the algorithm (too many loops).

PR #285, opened 2020-09-19
Trying to resolve issue 241.

autonomousapps:
Resolves (?) https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/241

Need to ensure I have a test that actually uses the subextension.


autonomousapps:
I tested locally and this in fact resolves the issue.

PR #294, opened 2020-10-02
Initial pass at making the variant ABI more accessible.

autonomousapps:
Intended to resolve https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/237.

This can be tested as follows:
```groovy
// Create a new task with type `AbstractPostProcessingTask`
def postDebug = tasks.register("postProcessDebug", AbstractAbiPostProcessingTask) {
  doLast {
    String abiDump = abiDump()
    println(abiDump)
  }
}

dependencyAnalysis {
  registerAbiPostProcessingTask(postDebug, "debug")
}
```
and then `./gradlew lib:postProcess`. You should see your ABI dump get printed to console.

This is very similar to the existing facility to consume the advice output, described in the wiki [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Post-processing).

### DIFF
--- a/src/main/kotlin/com/autonomousapps/AbstractAbiPostProcessingTask.kt
+++ b/src/main/kotlin/com/autonomousapps/AbstractAbiPostProcessingTask.kt
@@ -1,0 +1,22 @@
+package com.autonomousapps
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.PathSensitive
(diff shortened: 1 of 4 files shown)


autonomousapps:
Whoops, I just realized that this will default to giving you the ABI of whatever the _last_ variant to be processed was. Unlike the advice stuff, this does not attempt to coalesce all variants into a single report. So, the above code snippet will work, but it's ambiguous as to which variant is being analyzed. Will update the PR later to be more intelligent about that.

PR #311, opened 2020-10-15
Find lint jars. (But don't do much with them, yet.)

autonomousapps:
Partially resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/303.

So far this PR only 
1. Finds android libraries that include lint jars
2. Includes this information in certain reports (including `reason`)

It does not yet use it for computing advice. To do that, I think I will need to also check if the library is either
1. being used in some other capacity (for example Timber, which is a library + a lint jar)
2. is _only_ a lint jar (such as the little robots RxLintRegistry)


autonomousapps:
Build failed https://gradle.com/s/brgwbkteiwhcu

autonomousapps:
Build failed https://gradle.com/s/546t42jsxitxc

PR #312, opened 2020-10-19
Don't recommend removing lint-only jars

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/303


autonomousapps:
Grumble. My initial pass at this fails because rxlint has an empty R class file
```
$ javap -c nl.littlerobots.rxlint.R
public final class nl.littlerobots.rxlint.R {
}
```
so I can't just look for jars with no class files, I also have to be sure the class files are themselves empty.

autonomousapps:
Ok, I think this works now. Let's see what CI says!

autonomousapps:
Build failed https://gradle.com/s/om7b4rp2j42xo

PR #388, opened 2021-05-03
Issue 386. Annotate project path as task input property.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/386 (I hope).


src/main/kotlin/com/autonomousapps/tasks/AdviceSubprojectAggregationTask.kt
```diff
@@ -22,12 +22,13 @@ abstract class AdviceSubprojectAggregationTask : DefaultTask() {
...
   }
 
-  private val projectPath = project.path
-
   /*
    * Inputs
    */
 
+  @get:Input
+  val projectPath = project.path
```
autonomousapps, line 30: This change is the actual fix.

PR #413, opened 2021-06-13
Fix inputs for Option arguments

eugene-krivobokov:
`@Option` arguments weren't considered as inputs for tasks. This resulted in wrong up-to-date or cache results.


autonomousapps:
Thanks for the PR! Can't believe I missed that (really: I would have thought that Gradle's `validatePlugins` task would catch it). Hm, I wonder if they need to be "beans" now; that is, have a `get` prefix.

autonomousapps:
CI passes, verified it locally. Thanks again!

PR #426, opened 2021-06-21
Fix: don't fail in configuration when an Android unit test variant is disabled.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/423.


autonomousapps:
Build failed https://gradle.com/s/igjnpvtaar4vc

PR #429, opened 2021-06-24
Java bytecode analysis will filter to analyze only jar files

Astro03:
I'm currently working in a repo that leaks non-jar file types (i.e. xml) onto
the classpath. The other filetypes causes the dependency analysis tool to
throw ZipExceptions since the tool tries to open each file as a ZipFile.

Test:
1. Import dependency analysis into my repo
2. Run "./gradlew buildHealth" -> fails
3. Build dependency analysis tool with these changes (make snapshot)
4. Import snapshot of dependency analysis into my repo
5. Run "./gradlew buildHealth" -> passes and shows valid results

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/JarAnalyzer.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/JarAnalyzer.kt
@@ -12,6 +12,7 @@ import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.artifacts.result.DependencyResult
 import org.gradle.api.artifacts.result.ResolvedDependencyResult
 import org.gradle.api.logging.Logger
+import java.util.zip.ZipException
 import java.util.zip.ZipFile
 
 /**
@@ -28,7 +29,9 @@ internal class JarAnalyzer(
 
   fun components(): List<Component> {
     computeTransitivity()
-    return artifacts.asComponents()
+    return artifacts
+      .filter { it.file.name.endsWith(".jar") }
+      .asComponents()
   }
 
   private fun computeTransitivity() {
--- a/src/main/kotlin/com/autonomousapps/tasks/FindServiceLoadersTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindServiceLoadersTask.kt
@@ -49,9 +49,9 @@ abstract class FindServiceLoadersTask : DefaultTask() {
 
   @TaskAction fun action() {
     val outputFile = output.getAndDelete()
-    val serviceLoaders: Set<ServiceLoader> = artifacts.flatMapToSet {
-      findServiceLoaders(it)
-    }
+    val serviceLoaders: Set<ServiceLoader> = artifacts
+      .filter { it.file.name.endsWith(".jar") }
+      .flatMapToSet { findServiceLoaders(it) }
 
     outputFile.writeText(serviceLoaders.toJson())
   }
--- a/src/main/kotlin/com/autonomousapps/tasks/InlineMemberExtractionTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/InlineMemberExtractionTask.kt
@@ -127,8 +127,9 @@ internal class InlineDependenciesFinder(
 
   // from the upstream bytecode. Therefore "candidates" (not necessarily used)
   private fun findInlineImportCandidates(): Set<ComponentWithInlineMembers> {
-    return artifacts
-      .map { artifact ->
+    return artifacts.filter {
+        it.file.name.endsWith(".jar")
+      }.map { artifact ->
         artifact to InlineMemberFinder(inMemoryCacheProvider, logger, ZipFile(artifact.file)).find().toSortedSet()
       }.filterNot { (_, imports) ->
         imports.isEmpty()


autonomousapps:
Thanks for the contribution! This seems like a reasonable improvement given your use-case.

autonomousapps:
Build passes, which is great. I'm wondering if this is worth writing a regression test for. Could you provide a minimal reproducer, or tell me how to do that? How _do_ you end up with an XML on the classpath, anyway? I don't expect you to necessarily be able to write the test in the highly idiosyncratic test suite, but you'd also be welcome to try, and I could provide advice if you're interested.

Astro03:
Awesome. I'll try to get to the unit test today and would welcome any sort of advice you were willing to share. 

Any tips on the functional tests? I can't seem to get them to work on my machine and the wiki wasn't very useful for my case. 385 test succeeded, 280 tests fail, with 1 skipped.  I installed Android Studio and added the ANDROID_HOME  and PATH. 
```
"Running test: Test buildHealth can be executed (Gradle 6.1.1 AGP 3.5.4)(com.autonomousapps.android.OtherAndroidSpec)

OtherAndroidSpec > buildHealth can be executed (#gradleVersion AGP #agpVersion) > com.autonomousapps.android.OtherAndroidSpec.buildHealth can be executed (Gradle 6.1.1 AGP 3.5.4) FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure at OtherAndroidSpec.groovy:34
```

As for the xml files on the classpath, it's legacy code and I cannot change it since we have custom gradle tasks which use it.

autonomousapps:
This improvement will have to be tested by the functional test suite, I think, not the unit test suite.

Can you provide a build scan for your test failure? I've put some effort into making the test suite runnable in different environments.

One trick for making the functional tests a bit faster when running locally is to pass `-DfuncTest.quick` to your gradle invocation. That will run only a subset of the tests. Still takes a while, but considerably faster than the full suite.

> As for the xml files on the classpath, it's legacy code and I cannot change it since we have custom gradle tasks which use it.

I'm not suggesting you change your code. I'm just asking if you can tell me how to reproduce it so we could collaborate on writing a test that verifies this new behavior.

Astro03:
Here is a [buildscan](https://gradle.com/s/5dayc5vuam5kw). 

I hope I didn't miss anything. 

I'll look into how to update the classpath with an xml file and get back to you. thx

Astro03:
To add an xml to the compile classpath

```
// I added this to a module's build.gradle
dependencies {
    compile files('astro.xml')
}
```

Poojith:
> How do you end up with an XML on the classpath, anyway? 

Just to give you some context on why these XML files end up on the classpath - we have factory subprojects (responsible for loading and creating dependent objects on our internal Spring version) that generate these XML files to compile property values (part of our config system).

autonomousapps:
@Astro03 you can see the reason your functional tests failed [here](https://scans.gradle.com/s/5dayc5vuam5kw/tests/:functionalTest/com.autonomousapps.android.AnnotationProcessorSpec/autovalue%20is%20used%20with%20kapt%20(Gradle%206.9%20AGP%203.5.4)?anchor=e30&focused-exception-line=0-14&top-execution=1). tl;dr apparently you haven't done Android development on that machine :) You need to install the SDK tools. I'll make a point to update the readme and/or wiki for the future, but for now...

You have three options:
1. [Download](https://developer.android.com/studio/) and install Android Studio, and then launch the sdk manager and install the missing build tools and platform SDKs.
2. [Download](https://developer.android.com/studio/) the CLI tools and run `sdkmanager` from the command line. (scroll down to **Command line tools only**)
3. Don't bother running the Android tests and save yourself the trouble. To run only the JVM tests, do the following
```bash
./gradlew funcTest --tests com.autonomousapps.jvm.*
```

Regarding advice on how to write a test, take a look at `SecurityProviderSpec` as a simple example of a JVM spec. It runs an artificial project defined at `SecurityProviderProject`. In that project, there's a block like this:

```groovy
def builder = newGradleProjectBuilder()
builder.withSubproject('proj') { s ->
  s.sources = sources
  s.withBuildScript { bs ->
    bs.plugins = plugins
    bs.dependencies = dependencies
  }

  // add this for your use-case
  s.withFile('path/to/foo.xml', """\
      <!-- Add your XML content here -->
    """.stripIndent()
  )
}

...
```
where you would have to add `s.withFile(...)`. This tells the test suite to write a file with the given content to the given path.

Next, you'll need to create a dependency on your XML by updating `bs.dependencies = dependencies` to include something like `implementation files('astro.xml')`. `bs.dependencies` is a `List<Dependency>`, and I never contemplated having file dependencies like that in my test suite, so you'll have to update the code to handle that.

**Important** do _not_ use the `compile` configuration, as that does not exist in Gradle 7, and my plugin is Gradle 7-compatible.

One other piece of advice. If you can't figure out how to get the artificial build script to look how you want it, or you want to double-check that your XML got written as expected, go to `AbstractJvmSpec` and comment out the cleanup code. Do this:
```groovy
def cleanup() {
  if (gradleProject != null && shouldClean) {
    //clean(gradleProject.rootDir)
  }
}
```
Then, you can manually inspect the generated files at `build/functionalTest/...`

I appreciate your willingness to take this on. Let me know if you run into any issues.

Astro03:
@autonomousapps  Thanks for the confirmation on the suspicions. 

and thank ou for letting me know how to  run the tests.. this is awesome. 

I'll prob get to it next week sometime. thanks again

Astro03:
@autonomousapps 

I'm actually not sure how to repro the problem I am seeing at work.  

I created a normal Gradle project and added the xml file to the compile configuration. Your plugin works. I also tried adding xml files to other projects and your plugin also worked.

I'm not sure how to reproduce this issue I am seeing then. This might be caused by one of the plugins we use. 

Any idea on how I can reproduce getting non-jar items in the "components"  or any of the other places I made code changes here?

autonomousapps:
Thanks for trying. I'll take a look and see if I can repro. I thought it would be as simple as https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/429#issuecomment-867946582.

Even if I fail to repro, I'm inclined to merge as the changes seem good and don't cause any harm.

autonomousapps:
I also cannot replicate this failure with the information given. It is something more than just
```groovy
dependencies {
  compile files('foo.xml')
}
```
but it's unclear what.

Astro03:
Unfortunately, I thought the issue was reproducible by that simple snippet but it's not. 

The plugin must do something else other than put it on the compile classpath. 

Do you by change know how items could get onto the "components" list in the Jar Analyzer? I was trying to walk through the code but haven't tracked it down yet.

autonomousapps:
I'm merging this. I'd like a regression test for your use-case, but neither of us has managed to reproduce it yet :upside_down_face: The changes are safe, though.

PR #436, opened 2021-07-01
Don't report java-library plugin as redundant.

autonomousapps:
Don't emit warning about redundant plugins when behavior is set to ignore.


autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/392.
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/431.

PR #437, opened 2021-07-02
Fix Android manifest parsing in case of queries node

eugene-krivobokov:
### The problem

```
> ./gradlew :lib:projectHealth

Execution failed for task ':lib:extractPackageNameFromManifestDebug':

java.lang.IllegalStateException: it.attributes.getNamedItem(component.attrName) must not be null
  at com.autonomousapps.tasks.ManifestPackageExtractionTask.componentNames(ManifestPackageExtractionTask.kt:86)
  at com.autonomousapps.tasks.ManifestPackageExtractionTask.extractManifestComponents(ManifestPackageExtractionTask.kt:66)
  at com.autonomousapps.tasks.ManifestPackageExtractionTask.action(ManifestPackageExtractionTask.kt:43)
```

It happens for manifests with `provider` inside [queries](https://developer.android.com/guide/topics/manifest/queries-element#provider). It has other attributes.

Example: [com.yandex.android:mobmetricalib:3.20.1](https://repo1.maven.org/maven2/com/yandex/android/mobmetricalib/3.20.1/)

### Changes

Narrowed searching of components in the manifest to the `<application>` node.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/ManifestParser.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/ManifestParser.kt
@@ -1,0 +1,72 @@
+package com.autonomousapps.internal
+
+import com.autonomousapps.internal.utils.buildDocument
+import com.autonomousapps.internal.utils.mapToSet
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+import java.io.File
+
+internal class ManifestParser {
+
+  class ParseResult(
+    val packageName: String,
+    val components: Map<String, Set<String>>
+  )
+
+  fun parse(manifest: File): ParseResult {
+    val document = buildDocument(manifest)
+
+    val packageName = packageName(document)
+    val application = application(document)
+    // "service" is enough to catch LeakCanary, and "provider" makes sense in principle. Trying not to be too aggressive.
+    val services = application?.componentNames(Manifest.Component.SERVICE, packageName) ?: emptySet()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/autonomousapps/internal/ManifestParser.kt
```diff
@@ -0,0 +1,72 @@
...
+  class ParseResult(
+    val packageName: String,
+    val components: Map<String, Set<String>>
+  )
+
+  fun parse(manifest: File): ParseResult {
+    val document = buildDocument(manifest)
+
+    val packageName = packageName(document)
+    val application = application(document)
```
eugene-krivobokov, line 20: > Narrowed searching of components in the manifest to the <application> node.

This is the only change in behavior

src/test/kotlin/com/autonomousapps/internal/ManifestParserTest.kt
```diff
@@ -0,0 +1,102 @@
...
+          <queries>
+          
+            <package android:name="com.otherapp" />
+            
+            <intent>
+                <action android:name="com.otherapp.ACTION" />
+            </intent>
+            
+            <provider
+              android:authorities="com.otherapp.provider" />
```
eugene-krivobokov, line 68: Example

These elements are ignored. They are about an external app.

autonomousapps (approved):
Thanks so much. This looks great.

PR #441, opened 2021-07-26
Resolves bundle issue.

autonomousapps:
* Now supports test dependencies.
* Now supports multiple hops between dependencies in a bundle.

Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/425.


autonomousapps:
Build failed

PR #484, opened 2021-09-23
Analyze Dagger's annotation processor.

autonomousapps:
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/479

~Still WIP, as this clearly warrants a regression test.~

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
@@ -20,6 +20,7 @@ import org.gradle.api.tasks.Optional
 import java.io.BufferedReader
 import java.io.File
 import java.io.Writer
+import java.net.URL
 import java.net.URLClassLoader
 import java.util.*
 import java.util.zip.ZipFile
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
```diff
@@ -90,30 +91,28 @@ abstract class FindDeclaredProcsTask : DefaultTask() {
...
+    val annotationProcessorProcs = procs(annotationProcessorArtifacts, apClassLoader)
     val procs = kaptProcs + annotationProcessorProcs
 
     outputFile.writeText(procs.toJson())
     outputPrettyFile.writeText(procs.toPrettyString())
   }
 
+  private fun newClassLoader(name: String, files: FileCollection?): ClassLoader? {
+    val urls = files?.toList()?.map { it.toURI().toURL() }?.toTypedArray()
+    return urls?.let { FirstClassLoader(name, urls, javaClass.classLoader) }
```
autonomousapps, line 113: Fix is to use this new custom classloader instead of a `URLClassLoader` directly.

PR #494, opened 2021-10-07
Enum only JARs should not be reported as compileOnly

mvegter:


mvegter:
Hey @autonomousapps , sorry for the so far cryptic PR's. I'm working on moving internal changes for internal projects to the public side of the internet. Will add some minimal projects to reproduce the issue we had so that I can share that.

autonomousapps:
I'd appreciate some reproducers, yes.

Here's the build scan with the failed tests, ftr https://gradle.com/s/ze5i2urw2bu36

PR #500, opened 2021-10-10
Do not suggest that ABI dependencies are unused

### LINKED ISSUE #501: Plugin suggests that ABI dependencies are unused while they are needed for both compile
**Plugin version**
[main@ccaadc5fc86155f57650325784d6f105afe1be9b](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/commit/ccaadc5fc86155f57650325784d6f105afe1be9b) / 0.78.1-snapshot

**Gradle version**
7.2

**Describe the bug**
```
> Task :impl:projectHealth
  api(project(":abstract"))
```

**To Reproduce**
`abstract/src/main/java/IPermission.java`
```java
public interface IPermission {
    String name();
    boolean isAdmin();
    default Object getLogging() {
        return name();
    }
}
```

`impl/src/main/java/WebPermission.java`
```java
public enum WebPermission implements IPermission {

    ViewData(false),
    EditData(true),
    ;

    private final boolean superRole;

    WebPermission(boolean superRole) {
        this.superRole = superRole;
    }

    @Override
    public boolean isAdmin() {
        return superRole;
    }

}
```

**Expected behavior**
Do not suggest changing the dependency `api project(':abstract')` for the `:impl` project to unused. As removing this dependency won't compile the project.

**Additional context**
Changing the dependency from `api` to `implementation` will cause upstream issue, e.g. `Cannot access IPermission`

mvegter:
This resolves #501

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/AbiUsedClassSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/AbiUsedClassSpec.groovy
@@ -1,0 +1,25 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.AbiUsedClassProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class AbiUsedClassSpec extends AbstractJvmSpec {
+
+  def "do not suggest that ABI dependencies are unused (#gradleVersion)"() {
+    given:
+    def project = new AbiUsedClassProject()
+    gradleProject = project.gradleProject
+
+    when:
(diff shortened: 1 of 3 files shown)


autonomousapps:
Could you file an issue explaining the need for this?

autonomousapps:
Finally had time to look at this, and found the root cause. Somehow the asm code wasn't tracking interfaces! :scream: 

I have a simple fix for it, and will steal your spec as a perfect regression test, but I'm not going to take the change to `Advisor.kt`, as it was more of a band-aid. Much appreciated.

autonomousapps:
Closing as resolved.

PR #505, opened 2021-10-14
Declared CompileOnly file dependencies should not be marked  as transitive

mvegter:
This also solved a bug where file (.jar) dependency returned null, will see if I can reproduce it in the test framework.


autonomousapps:
> You sort of do, with the AGP 7. The problem was the sub type of the files object, see the latest (amended) change. Currently preferring null over throwing an Exception to keep it as close to old behavior, but perhaps the unmapped subtype should throw too.

Thanks for the updated PR. I'm happier about this. I want to take a closer look later before merging.

autonomousapps:
Thanks for the contribution! I merged it and made a minor change to the spec to avoid exceeding access rights quite so egregiously :)

PR #516, opened 2021-10-18
Non interesting configuration should not prevent suggestion on main configuration

mvegter:


mvegter:
Perhaps you could have a look (or give some pointers) here too @autonomousapps ; The [failing test](https://scans.gradle.com/s/tt2p6xjsaqfly/tests/overview?outcome=failed&toggled=WyI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC5hbm5vdGF0aW9ucHJvY2Vzc29yc3BlYy01OSIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmFkdmljZXNwZWMtNDciLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC5rdHhzcGVjLTYzIiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmRlcGVuZGVuY2llc3NwZWMtNDQiLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC5jb21waWxlb25seXNwZWMtNTAiLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC5zZXJ2aWNlbG9hZGVyc3BlYy02MiIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLnJlc3NwZWMtNDgiLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC52YXJpYW50c3BlYy02NSIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmRhdGFiaW5kaW5nc3BlYy01NSIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmxpbnRqYXJzcGVjLTUzIiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmFuZHJvaWQudmlld2JpbmRpbmdzcGVjLTUxIiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmFuZHJvaWQubGVha2NhbmFyeXNwZWMtNjEiLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC50ZXN0c291cmNlc3BlYy00NSIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmRhZ2dlcnNwZWMtNTQiLCI6ZnVuY3Rpb25hbHRlc3QtMC1jb20uYXV0b25vbW91c2FwcHMuYW5kcm9pZC5ub29wc3BlYy01NyIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLnRyYW5zaXRpdmVhZHZpY2VzcGVjLTU2IiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmFuZHJvaWQuYXBwY29tcGF0c3BlYy02NyIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmRvdWJsZWRlY2xhcmF0aW9uc3BlYy02NCIsIjpmdW5jdGlvbmFsdGVzdC0wLWNvbS5hdXRvbm9tb3VzYXBwcy5hbmRyb2lkLmFuZHJvaWRrb3RsaW5pbmxpbmVzcGVjLTU4IiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmFuZHJvaWQuYW5kcm9pZHRlc3RkZXBlbmRlbmNpZXNzcGVjLTY2IiwiOmZ1bmN0aW9uYWx0ZXN0LTAtY29tLmF1dG9ub21vdXNhcHBzLmFuZHJvaWQuYXJiaXRyYXJ5ZmlsZXNwZWMtNTIiXQ) seem to be related to the Android Debug configuration, I'm not familiar with Android so I'm a bit in the dark about the approach to take here.

src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
```diff
@@ -94,7 +94,7 @@ abstract class ArtifactsReportTask : DefaultTask() {
         }
       }.filterToSet { art ->
         exclusions.none { ex ->
-          art.dependency.identifier == ex.identifier
+          art.dependency.identifier == ex.identifier && art.dependency.configurationName == ex.configurationName
```
autonomousapps, line 97: My intuition is that this is the wrong approach. (My intuition might be wrong.) I don't have advice right now, but I'll try to take a look when I get a chance.

mvegter: Hey @autonomousapps , do you by any chance have any news regarding this PR

autonomousapps: Thanks for the reminder. I will try to take a look soon. Basically spent the past month on-call at work, a bit burned out!

autonomousapps: I think I'm just not clear on what your goal is here. Could you file an issue? What problem are you running into that you're trying to solve?

mvegter: The issue is that a transitively used ABI dependency (NotificationSender parameter in MainApplication) is not part of the suggestions because the dependency is excluded from the locations (because the integrationTest is marked as not interesting, but it filters on the whole project instead of project + sourceSet).

autonomousapps: It would help if you could write it up a bit more concretely as an issue. Even pseudocode is helpful. Sample build.gradle with dependency declarations, sample source code where dependencies are being used.

mvegter:
The test passes with `v2` so this Spec can be used as another test case

autonomousapps (commented):
I'm inclined to accept this, but I want to run something past you. I think that I will _only_ analyze the `main` and `test` source sets in an upcoming PR. Those two source sets have a well-known relationship that I can (and currently do) encode. However, that cannot be said for arbitrary source sets added by users. At best, I suspect I might add a feature to opt in additional source sets. The upshot of this in the current context is that I _think_ your spec would change. The advice would instead be
```groovy
Advice.ofAdd(dependency(project('api', ':core')), 'api')
```
I.e., it would ignore `integrationTest` as if it didn't exist and only say to add that dependency to `api`. WDYT?

mvegter:
Sounds good to me. Although the testFixtures source set analysis might be required for proper suggestion on test. That being said, for now the Spec will cover that the suggestion, either add or change, will exist.

autonomousapps:
> Although the testFixtures source set analysis might be required for proper suggestion on test.

Noted. Maybe you can help write a couple of (failing) specs when the time comes?

PR #521, opened 2021-11-01
Detect more res-by-res usages.

autonomousapps:
This is much more comprehensive than what came before, hopefully such that it will now capture all such usages.

Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/513

~WIP: Needs a test before merging.~


src/main/kotlin/com/autonomousapps/tasks/AndroidResToResAnalysisTask.kt
```diff
@@ -177,5 +176,58 @@ private sealed class Res
...
+ */
+private data class AttrRes(
+  val type: String,
+  val id: String
+) : Res() {
+
+  companion object {
+
+    private val TYPE_REGEX = Regex("""@(?:.+:)?(.+)/""")
+    private val ATTR_REGEX = Regex("""\?(?:.+/)?(.+)""")
```
autonomousapps, line 190: Ugh regex. If anyone feels like double-checking, have at it.

PR #530, opened 2021-12-04
Fixes issue 529.

cve92:
File.separator instead of / so that file system paths are processed properly when running builds on windows. Actually the problem was covered by TestDependenciesSpec "unused test dependencies are reported" already: Running it on windows failed before doing this change. With the change it now passes.


autonomousapps:
Thank  you for this contribution! It'll be in the next release.

Issue #531, opened 2021-12-20
NullPointerException at com.autonomousapps.tasks.AttrRes$Companion.type


ColtonIdle:
**Plugin version**
0.79.0

**Gradle version**
7.3.2

**(Optional) Android Gradle Plugin (AGP) version**
7.2.0-alpha03

**Describe the bug**
Running the buildHealth task fails with this stacktrace

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':CustomViewLibrary:findAndroidResByResUsageDebug'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:145)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:143)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:131)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
[...]

PR #543, opened 2022-01-04
Filter out DataBinding expressions when extracting resources from xml files

nsk-mironov:
I've tried running `buildHealth` with `0.79.0` version of the plugin and got the same exception as in #531 and #542. The only difference that it has nothing to do with `@null` in my case and is actually caused by [data binding expressions](https://developer.android.com/topic/libraries/data-binding/expressions) like `binding:subtitle="@{model.subtitle}"`.

The fix from #534 actually prevents the exception in my case, but it doesn't cover all databinding related scenarios. For example, expressions like `binding:outlineRadius="@{@dimen/radius_normal}"` are still parsed as a valid `AttrRef(type = "{@dimen", id = "radius_normal}")` object.

This PR introduces the following changes:
- `AttrRef.from` now always returns `null` for all data binding expressions.
- `ResSpec.groovy` runs one more case which is specially dedicated to data binding expressions.


src/functionalTest/kotlin/com/autonomousapps/fixtures/DataBindingWithExpressionsProject.kt
```diff
@@ -0,0 +1,48 @@
+package com.autonomousapps.fixtures
+
+import com.autonomousapps.advice.Advice
+
+class DataBindingWithExpressionsProject(
```
autonomousapps, line 5: Please rewrite this fixture to use a similar style to `AttrResWithNullProject` from the spec just above your new one. This is the first style of fixture used in this project and... I hate it. I'd like to eliminate it entirely at some point.

src/main/kotlin/com/autonomousapps/model/Source.kt
```diff
@@ -71,14 +71,16 @@ data class AndroidResSource(
...
+       * 3. They're not a data binding expression (don't start with `@{` and end with `}`)
+       * 4. Their value starts with `?`, like `?themeColor`.
+       * 5. Their value starts with `@`, like `@drawable/`.
        *
        * Will return `null` if the map entry doesn't match an expected pattern.
        */
       fun from(mapEntry: Map.Entry<String, String>): AttrRef? {
         if (mapEntry.isId()) return null
         if (mapEntry.isToolsAttr()) return null
+        if (mapEntry.isDataBindingExpression()) return null
```
autonomousapps, line 83: Thank you for updating this in both locations!

autonomousapps (changes requested):
Thank you for this contribution! Very much appreciated. Only one change I'd like to see.

src/functionalTest/kotlin/com/autonomousapps/fixtures/DataBindingWithExpressionsProject.kt
nsk-mironov: Sure, not a problem. Please check my latest commit with an updated `DataBindingWithExpressionsProject` implementation.

autonomousapps (approved):
LGTM!

PR #552, opened 2022-01-13
Change log level to not spam console output

daanschipper:


src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
```diff
@@ -68,7 +68,7 @@ internal fun Project.isV1(): Boolean {
     .forUseAtConfigurationTime()
     .getOrElse("1") == "1"
 
-  if (!isV1) logger.quiet("Using plugin v2")
+  if (!isV1) logger.info("Using plugin v2")
```
autonomousapps, line 71: I agree this is annoying spam, but as I said in another review, this flag is intended for internal use only. That said, I think we could just remove the logging entirely, since I can tell which version is being used based on the tasks that get executed.

autonomousapps:
Thank you for the contribution. Please see the new PR I just pushed for this, which removes the logging entirely.

PR #553, opened 2022-01-13
Do not retrieve non-main declarations when shouldAnalyzeTests is disabled

mvegter:
Including the declarations of non-main configurations when analyzing with the following configuration:
```
systemProp.v=2
systemProp.dependency.analysis.test.analysis=false
```
Resulted in suggestions related to test (and not yet supported testFixtures) source sets. This PR fixes that when the test analysis is disabled.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/configuration/Configurations.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/configuration/Configurations.kt
@@ -23,6 +23,10 @@ internal object Configurations {
     return MAIN_SUFFIXES.any { suffix -> configurationName.endsWith(suffix = suffix, ignoreCase = true) }
   }
 
+  internal fun isMainStrict(configurationName: String): Boolean {
+    return MAIN_SUFFIXES.any { suffix -> configurationName == suffix }
+  }
+
   internal fun isAnnotationProcessor(configurationName: String): Boolean {
     return ANNOTATION_PROCESSOR_PREFIXES.any { prefix -> configurationName.startsWith(prefix) }
   }
--- a/src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
@@ -4,15 +4,16 @@ import com.autonomousapps.TASK_GROUP_DEP_INTERNAL
 import com.autonomousapps.internal.NoVariantOutputPaths
 import com.autonomousapps.internal.configuration.Configurations.isAnnotationProcessor
 import com.autonomousapps.internal.configuration.Configurations.isMain
+import com.autonomousapps.internal.configuration.Configurations.isMainStrict
 import com.autonomousapps.internal.utils.getAndDelete
 import com.autonomousapps.internal.utils.toIdentifiers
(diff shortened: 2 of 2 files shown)


autonomousapps:
Thank you for the contribution. There are some specs that are currently disabled for v2 because it doesn't (yet) support disabling analysis for test source sets. I can't move forward with this before first addressing that.

mvegter:
> Thank you for the contribution. There are some specs that are currently disabled for v2 because it doesn't (yet) support disabling analysis for test source sets. I can't move forward with this before first addressing that.
> 
> On that note, it would help if you could provide more information on what the output is without this change, and what you expect it to be, so that I can add another spec if necessary to cover the case.

Upon further investigation the problem is that, with test analysis disabled, any dependency declared on non-main sourceset is marked as unused if it is on the compileClasspath of main (both direct and transitive dependencies). When re-enabling test analysis there are suggestions related to JUnit/Mockito which were expected (as they are not on main).

Running the `projectHealth` with and without the dependency on the test sourceset shows that the output of the reports/dependency-analysis/main directory has not changed but only the top level reports/dependency-analysis. Files changed is the declarations.json and then the advice/output related files.

autonomousapps:
Could you provide more information on this case? I'd like to have a test case so that (1) I can see more clearly what is going on, and (2) support your use-case with a test.

That said, could you _also_ elaborate on why it's necessary to skip test analysis? What is the use-case there?

mvegter:
I will try to write up a spec al though the current TestDependenciesSpec might already cover it or cab be extended.

> That said, could you also elaborate on why it's necessary to skip test analysis? What is the use-case there?

It is not specifically the test sourceset but the non-main sourceset(s) that we want to exclude. Working on a monolith Gradle project the primary goal is a correct and smaller main claspath, and enforcing this.

Other than that it is also due to the lack of testFixtures support/detection but also custom sourcesets e.g. a `jmh` one which extends both main and testFixtures which is not a priority.

autonomousapps:
> I will try to write up a spec al though the current TestDependenciesSpec might already cover it or cab be extended.

Getting the remaining ignored tests passing with v2 is my current highest priority for this project.

autonomousapps:
Could you clarify if your project is an Android or JVM project? I was just looking at support for the test flag and realized that it does seem to already work for JVM projects. If not, then `TestDependenciesSpec` is missing a case. You don't need to write the spec, but if you could describe your project, that would help. If you have an Android project instead, then that's another story, because support for disabling test analysis for Android in v2 has not been implemented yet.

mvegter:
All projects are JVM. The problem occurs along the lines of
```
a.gradle

dependencies {
  api project(':b')
  testImplementation commons_codec 
}

b.gradle
dependencies {
  api commons_codec 
}
```
The `commons_codec` (or a subproject) is a unused transitive for the main sourceset in A but the testImplementation has it declared thus the suggestion is to remove it. With this change the testImplementation declaration is not in the declaration.json and thus not suggested.

autonomousapps:
The new PR I just posted resolves your issue and includes a test. I think we can close this?

autonomousapps:
Gonna close this in favor of the PR I just merged. Thanks for working it through with me :+1:

PR #555, opened 2022-01-13
Use a more descriptive flag for the model version

daanschipper:


autonomousapps (commented):
This system property is mainly intended for internal use while I iterate on v2. It is not intended as a public contract. Frankly, it's just easier to type `-Dv=2` than the alternative :)

My intention is that, once it reaches feature parity and all the tests pass, I will simply delete the v1 model.

daanschipper:
That also makes sense, we have been trying out the new version of the model already and thought a more descriptive name is more suited.

If your plan is to replace v1 with v2 then this pr is not needed.

PR #566, opened 2022-01-19
Don't suggest removing test dependencies when test analysis is disabled.

autonomousapps:
With inspiration from https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/553.


mvegter:
Initial run looks good, will test on more projects. Only thing is that the `Existing dependencies which should be modified to be as indicated:` are unsorted again :(

autonomousapps:
> Some suggestion changed from the change sourceSet / api category to the transitive which I believe to be the expected behavior.

I very much appreciate your feedback here. Sanity checks against real projects are critical for ensuring accuracy. Please continue to let me know what you find, and thanks so much!

PR #568, opened 2022-01-19
Sort the output of ProjectHealthConsoleReportBuilder on the printed suggestion

mvegter:


mvegter:
The Advice#compareTo is not the natural sorting that I was aiming for, this fixes it by doing it the old way of sorting the (to) print advice.

autonomousapps (commented):
At this point, I think this class should be tested so we can really know it works, and protect against regressions.

autonomousapps:
Thanks for keeping your branch up to date @mvegter. Are you interested in writing a unit test to validate the change?

src/test/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilderTest.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun adviceOfRemoveShouldBeSorted() {
+    val dependencyAdvice = setOf(
+      Advice.ofRemove(ModuleCoordinates("com.project.a", "1.0"), "implementation"),
+      Advice.ofRemove(ModuleCoordinates("com.project.c", "1.0"), "api"),
+      Advice.ofRemove(ModuleCoordinates("com.project.b", "1.0"), "api"),
+    )
+    val projectAdvice = ProjectAdvice("dummy", dependencyAdvice, emptySet())
+
+    val consoleText = ProjectHealthConsoleReportBuilder(projectAdvice).text
+    Truth.assertThat(consoleText).isEqualTo("" +
```
autonomousapps, line 21: nit: normally I prefer a static import of `Truth.assertThat`

autonomousapps (approved):
Thanks for the test!

Issue #435, opened 2021-06-30
Failed to apply plugin with Gradle 7.1


ianbrandt:
**Build scan link**
Unfortunately I'm not permitted to provide a build scan of my project for security reasons. If necessary I can try to create a minimal reproducer and provide a scan of that, but this looks like a fairly straightforward case of deprecated API use. That is, straightforward from a cause standpoint, which is not to say that it's necessarily easy to fix.

**Plugin version**
0.74.0

**Gradle version**
7.1

**(Optional) Android Gradle Plugin (AGP) version**
Not an Android project.

**Describe the bug**
After upgrading to Gradle 7.1 I get a `ProjectConfigurationException` when applying the plugin.

**To Reproduce**
1. Upgrade to Gradle 7.1.

**Expected behavior**
The project configures successfully.

**Additional context**

https://docs.gradle.org/7.1/userguide/upgrading_version_6.html#calling_project_afterevaluate_after_project_evaluation_is_now_an_error

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'myproject'.
> Failed to apply plugin 'com.autonomousapps.dependency-analysis'.
   > Cannot run Project.afterEvaluate(Action) when the project is already evaluated.

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'myproject'.
        at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:75)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:68)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.access$400(LifecycleProjectEvaluator.java:51)
        at org.gradle.configuration.project.LifecycleProjectEvaluator$NotifyAfterEvaluate.run(LifecycleProjectEvaluator.java:195)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$3.execute(DefaultBuildOperationRunner.java:75)
[...]

PR #574, opened 2022-01-26
Prevent project is already evaluated error

Artemych:
Please take a look at the issue [#435](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/435)

For some mutlimodule projects usage of that plugin could not be possible as users receives an error like: "Cannot run Project.afterEvaluate(Action) when the project is already evaluated" during the project sync.

It could be caused by manipulating with  the evaluation order. To prevent that we have to check the current internal project state.


autonomousapps:
I can't merge this without a regression test. Could you provide detailed steps on how I'd set up a project so that this would be necessary?

Artemych:
> I can't merge this without a regression test. Could you provide detailed steps on how I'd set up a project so that this would be necessary?

regression pipeline is passed.

About the steps: 
In our project we have some logic with the evaluation order like [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/435#issuecomment-872310565). Hope these description will be enough.

autonomousapps:
I appreciate you keeping your branch up to date. Unfortunately, I haven't had time to write a test on behalf of this change. Is that something you'd be interested in doing? I'm willing to offer guidance.

Artemych:
> I appreciate you keeping your branch up to date. Unfortunately, I haven't had time to write a test on behalf of this change. Is that something you'd be interested in doing? I'm willing to offer guidance.

Trying to do my best to add the test for that particular case. For now unfortunately I haven't time but hope i will do that by the end of the week.

PR #595, opened 2022-02-16
Inline usages regression in 1.0.0-rcXX

nsk-mironov:
This PR adds a test case that shows a regression in `1.0.0-rcXX` version of the plugin. It seems like plugin incorrectly handles cases when there are multiple files with `inline` functions under the same package:
- `lib-producer` module contains two files `Foo.kt` and `Bar.kt` under the same package.
- Each `Foo.kt` and `Bar.kt` declares an `inline` function.
- `lib-consumer` module calls a function declared in `Foo.kt`.
- Plugin reports `lib-producer` module as unused.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
@@ -183,18 +183,8 @@ internal class InlineMembersFinder(
           // Guaranteed to be non-empty
           inlineMembers = inlineMembers
         )
-      }.groupByPackageName().also {
+      }.toSortedSet().also {
         inMemoryCache.inlineMembers(zipFile.name, it)
-      }
-  }
-
-  private fun Sequence<InlineMemberCapability.InlineMember>.groupByPackageName(): Set<InlineMemberCapability.InlineMember> {
-    return groupBy { it.packageName }
-      .mapToOrderedSet { group ->
-        InlineMemberCapability.InlineMember(
-          packageName = group.key,
-          inlineMembers = group.value.flatMapToOrderedSet { it.inlineMembers }
-        )
       }
   }
 
--- a/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
@@ -183,8 +183,18 @@ internal class InlineMembersFinder(
           // Guaranteed to be non-empty
           inlineMembers = inlineMembers
         )
-      }.toSortedSet().also {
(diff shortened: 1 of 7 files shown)


nsk-mironov:
So, I went ahead and implemented a fix as well https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/595/commits/dea64848e89338d218d2921bee5219fe0ec2f216.

nsk-mironov:
Also, the issue has nothing to do with Android, so I tweaked the test case a bit. https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/595/commits/a9a6fdb1aa5e466c4987a990fe022393a90184cf

autonomousapps:
Thanks for the contribution!

autonomousapps:
I ran your spec without the changes to `FindInlineMembersTask` and it still passed. So I'm not entirely sure what's going on here.

nsk-mironov:
Btw, here is a [build log](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/runs/5220332834?check_suite_focus=true#step:6:816) that shows the very first commit fails on CI.

src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
```diff
@@ -183,11 +183,21 @@ internal class InlineMembersFinder(
           // Guaranteed to be non-empty
           inlineMembers = inlineMembers
         )
-      }.toSortedSet().also {
+      }.groupByPackageName().also {
```
autonomousapps, line 186: Rather than this change, I think I'd prefer to fix this by fixing the `compareTo` function of `InlineMemberCapability.InlineMember`. Is that a change you're willing to take on? A useful example might be in how `Advice` implements that [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/model/Advice.kt#L24-L27). I believe that would be clearer and more understandable in the future, when we might forget why this change was made in the first place.

autonomousapps (changes requested):

src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
nsk-mironov: Sure. I reverted my original fix and fixed `InlineMember.compareTo` impelementation. The only problem I don't know any built-in way to compare `Set` instances, so I also added custom `LexicographicIterableComparator` implementation. Please let me know if you know any better way to compare sets.

autonomousapps (approved):
Love it. Thanks!

PR #598, opened 2022-02-22
TestKit now supports included builds. Don't crash in presence of included build. 

autonomousapps:
Resolves crash associated with https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/565, but it doesn't handle any analysis across the composite boundary.


autonomousapps:
Wondering aloud whether to merge as-is to resolve the crash, or wait until I have a complete solution for included builds.

autonomousapps:
Alright, all tests pass, so merging as-is, with partial support for included builds. Not crashing is a step in the right direction!

PR #603, opened 2022-03-05
Improve support for annotation processor advice.

autonomousapps:
Partially resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/597. However, this issue also suggests the need to add bundle support (etc) for annotation processors, since it seems like Hilt + Dagger would make a natural bundle.

Will also need an integration test. Awaiting minimal reproducer from https://github.com/wbervoets.


autonomousapps:
Build failed https://gradle.com/s/4yqgna7mct6mu

PR #606, opened 2022-03-13
Allow common configuration through build logic and override on project level

mvegter:
To share configuration across multiple sub projects we make use of the Gradle convention plugins. We have a few common dependency which we want to ignore for unused/configuration advices but sometimes this is also needed per-subproject level. We ran into the following on RC3:

```
Caused by: org.gradle.api.GradleException: You must configure this project either at the root or the project level, not both
Caused by: org.gradle.api.InvalidUserDataException: Cannot add a ProjectIssueHandler with name ':BLA:BLA' as a ProjectIssueHandler with that name already exists.
```


autonomousapps:
Thanks for this PR. Been a busy week, but I will review it soon.

autonomousapps (approved):
LGTM! Thanks again.

PR #613, opened 2022-03-21
Added allowed/ignored variants support

tronku:
Fixes - https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/519

### DIFF
--- a/src/main/kotlin/com/autonomousapps/Flags.kt
+++ b/src/main/kotlin/com/autonomousapps/Flags.kt
@@ -15,12 +15,16 @@ object Flags {
   private const val FLAG_MAX_CACHE_SIZE = "dependency.analysis.cache.max"
   private const val FLAG_TEST_ANALYSIS = "dependency.analysis.test.analysis"
   private const val FLAG_PRINT_BUILD_HEALTH = "dependency.analysis.print.build.health"
+  private const val FLAG_ALLOWED_VARIANTS = "dependency.analysis.android.allowed_variants"
+  private const val FLAG_IGNORED_VARIANTS = "dependency.analysis.android.ignored_variants"
 
   internal fun Project.shouldAnalyzeTests() = getGradleOrSysProp(FLAG_TEST_ANALYSIS, true)
   internal fun Project.shouldAutoApply() = getGradleOrSysProp(FLAG_AUTO_APPLY, true)
   internal fun Project.silentWarnings() = getGradlePropForConfiguration(FLAG_SILENT_WARNINGS, false)
   internal fun Project.printBuildHealth() = getGradlePropForConfiguration(FLAG_PRINT_BUILD_HEALTH, false)
   internal fun Project.shouldClearArtifacts(): Boolean = getGradleOrSysProp(FLAG_CLEAR_ARTIFACTS, true)
+  internal fun Project.getAllowedVariants() = getBuildPropForConfigurations(FLAG_ALLOWED_VARIANTS)
+  internal fun Project.getIgnoredVariants() = getBuildPropForConfigurations(FLAG_IGNORED_VARIANTS)
 
   internal fun Project.cacheSize(default: Long): Long {
     return providers.systemProperty(FLAG_MAX_CACHE_SIZE)
@@ -52,4 +56,9 @@ object Flags {
       .forUseAtConfigurationTime()
       .getOrElse(default.toString())
       .toBoolean()
+
+  private fun Project.getBuildPropForConfigurations(name: String) =
+    providers.gradleProperty(name)
+      .forUseAtConfigurationTime()
+      .getOrElse("").split(',')
 }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/autonomousapps/Flags.kt
```diff
@@ -15,12 +15,16 @@ object Flags {
   private const val FLAG_MAX_CACHE_SIZE = "dependency.analysis.cache.max"
   private const val FLAG_TEST_ANALYSIS = "dependency.analysis.test.analysis"
   private const val FLAG_PRINT_BUILD_HEALTH = "dependency.analysis.print.build.health"
+  private const val FLAG_ALLOWED_VARIANTS = "dependency.analysis.android.allowed_variants"
```
autonomousapps, line 18: I think I'd prefer
```
"dependency.analysis.android.variants.allowed"
"dependency.analysis.android.variants.ignored"
```

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -509,6 +516,20 @@ internal class ProjectPlugin(private val project: Project) {
     }
   }
 
+  private fun Project.executeOnAllowedVariants(variant: String, action: () -> Unit) {
+    val allowedVariants = getAllowedVariants()
+    val ignoredVariants = getIgnoredVariants()
+    val commonVariants = allowedVariants.intersect(ignoredVariants)
+
+    if (commonVariants.isNotEmpty()) {
+      throw GradleException("Allowed and ignored variants must be different. Issue found in variant(s): $commonVariants")
```
autonomousapps, line 525: Let's make this an `InvalidUserDataException`

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -509,6 +516,20 @@ internal class ProjectPlugin(private val project: Project) {
...
+    } else if ((allowedVariants.isEmpty() && ignoredVariants.isEmpty()) ||
+          (!ignoredVariants.contains(variant) && allowedVariants.contains(variant))) {
```
autonomousapps, line 527: I don't think this quite works. If allowed variants is empty, then the variant should be permitted by default (unless explicitly ignored). Similarly if ignored variants is empty, then the variant should be permitted by default (unless allowed variants is not empty).

Alternatively, I think I'm not convinced that we want _both_ an allow-list and an ignore-list. Is there a reason for that complexity? Why not just pick one?

autonomousapps (changes requested):
Thank you for the contribution! Found a couple of minor issues, but the main one that would prevent me from merging this is we need a new test to verify this behavior. There are several recent PRs that add features / fix bugs and add new tests, including an [open PR right now](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/606). Please take a look for inspiration, and don't hesitate to ask for help.

src/main/kotlin/com/autonomousapps/Flags.kt
tronku: agreed, makes sense

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
tronku: will go through again with the condition
the reason of having both allowed and ignored list - in case of a big project, one can have many variants like beta, gms, hms, internal, prod... so if one don't want to execute the task on prod variants then instead of adding all the possible variants that needs to be allowed, one only has to add the prod variant in the ignored list

autonomousapps:
Please see failures at https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/runs/5634738312?check_suite_focus=true

tronku:
./gradlew check is failing locally for the main branch too, I don't have experience in gradle testing, can you help please?

autonomousapps:
> ./gradlew check is failing locally for the main branch too, I don't have experience in gradle testing, can you help please?

Do you have a build scan from your failure?

tronku:
> Do you have a build scan from your failure?

Here - https://gradle.com/s/ddfw4lg5mq2ni

autonomousapps:
Ugh, that build scan is broken. Could you provide more information on what is failing for you?

tronku:
[check_gradle.txt](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/files/8370908/check_gradle.txt)
@autonomousapps check this out, ./gradlew check --info

autonomousapps:
In one of the earlier failures, I see this message:
```
AndroidTestDependenciesSpec > buildHealth succeeds when a unit test variant is disabled (#gradleVersion AGP #agpVersion) > com.autonomousapps.android.AndroidTestDependenciesSpec.buildHealth succeeds when a unit test variant is disabled (Gradle 7.2 AGP 4.2.2) STANDARD_OUT

    FAILURE: 
    Build failed with an exception.


    * What went wrong:

    A problem occurred configuring project ':proj'.

    > 
    Allowed and ignored variants must be different. Issue found in variant(s): []
```
I believe the final line there comes from your changes. Does that help?

mformetal:
@tronku I'd love this feature on a project I'm working on - would you like any help?

mformetal:
@autonomousapps Is it ok with you if I take over this PR? I'd love to integrate this plugin's tasks in one of my projects but we have way too many flavors for it to be reasonable on CI for my team members

tronku:
@mformetal i need help with creating tests here, can you help?

PR #617, opened 2022-03-23
Fix AbiHandler.excludeAnnotations.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/610.


src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
```diff
@@ -130,7 +131,8 @@ internal fun List<ClassBinarySignature>.filterOutNonPublic(
   fun ClassBinarySignature.isExcluded(): Boolean {
     return (sourceFile?.let(exclusions::excludesPath) ?: false) ||
       exclusions.excludesClass(canonicalName) ||
-      annotations.any(exclusions::excludesAnnotation)
+      annotations.any(exclusions::excludesAnnotation) ||
+      memberSignatures.any { it.annotations.any(exclusions::excludesAnnotation) }
```
autonomousapps, line 135: here's the fix.

PR #624, opened 2022-03-30
Don't register tasks if they can't be used.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/463.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/android/DependenciesSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/DependenciesSpec.groovy
@@ -60,7 +60,7 @@ final class DependenciesSpec extends AbstractAndroidSpec {
 
     then:
     assertThat(AdviceHelper.actualBuildHealth(gradleProject))
-      .containsExactlyElementsIn(AdviceHelper.emptyBuildHealthFor(':app', ':'))
+      .containsExactlyElementsIn(AdviceHelper.emptyBuildHealthFor(':app'))
 
     where:
     [gradleVersion, agpVersion] << gradleAgpMatrix()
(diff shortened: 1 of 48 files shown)


autonomousapps:
The test failures appear to be of this nature:
```
expected   : [ComprehensiveAdvice(projectPath=:, dependencyAdvice=[], pluginAdvice=[], shouldFail=false), ComprehensiveAdvice(projectPath=:app, dependencyAdvice=[Advice(dependency=:lib, usedTransitiveDependencies=[], parents=null, fromConfiguration=implementation, toConfiguration=null)], pluginAdvice=[], shouldFail=false), ComprehensiveAdvice(projectPath=:lib, dependencyAdvice=[], pluginAdvice=[], shouldFail=false)] |  

but was    : [ComprehensiveAdvice(projectPath=:app, dependencyAdvice=[Advice(dependency=:lib, usedTransitiveDependencies=[], parents=null, fromConfiguration=implementation, toConfiguration=null)], pluginAdvice=[], shouldFail=false), ComprehensiveAdvice(projectPath=:lib, dependencyAdvice=[], pluginAdvice=[], shouldFail=false)] |  
```
Where the _actual_ advice under this PR is now missing this:
```ComprehensiveAdvice(projectPath=:, dependencyAdvice=[], pluginAdvice=[], shouldFail=false)```

This makes sense, since I no longer configure tasks on the root project if it has no plugins applied (like java-library, kotlin, android, etc). On the one hand, this is definitely a win, since it means the plugin is doing less work. On the other, nearly 200 tests need to be updated 😭

autonomousapps:
This was merged.

PR #625, opened 2022-03-31
Replace 'artifacts.clear()' with 'isVisible=false'

### LINKED ISSUE #607: Stop clearing artifacts configuration
The following message:
```
Clearing artifacts from the archives configuration. 
 Disable this behavior with -Pdependency.analysis.clear.artifacts=false
 Suppress this warning with -Pdependency.analysis.warnings.silent=true
```
is fairly opaque - I've no idea which artifacts we're talking about, why they were in the archives configuration, whether I do or do not want them there, what the consequence of them being or not being there is, and consequently whether I should disable the behaviour or suppress the warning. I ended up checking out the codebase to discover that this is normal, not a result of something odd I'm doing elsewhere, and that consequently I almost certainly want to suppress the warning.

(Also, suppressing the warning seems to be a generic flag for _all_ warnings - as it happens from examining the codebase I think this is the only one that uses it, but perhaps a more specific flag for this specific case?)

jjohannes:
This remove the call to 'configurations["archives"].artifacts.clear()'.

It is no longer necessary if we set the 'consumable' configuration to 'isVisible = false'. Which this commit does.

With this change, we also remove the FLAG_CLEAR_ARTIFACTS and FLAG_SILENT_WARNINGS as these are no longer needed.

Fixes #607


src/main/kotlin/com/autonomousapps/Flags.kt
```diff
@@ -8,19 +8,14 @@ import org.gradle.api.Project
 @Suppress("DEPRECATION") // forUseAtConfigurationTime()
 object Flags {
 
-  internal const val FLAG_CLEAR_ARTIFACTS = "dependency.analysis.clear.artifacts"
```
autonomousapps, line 11: Would you mind leaving these flags here? (removing the methods is fine.) I'm thinking I might like to do a check at some point to let folks know about flags that no longer do anything to help them clean up their builds.

autonomousapps (approved):
One small requested change, and thank you!

src/main/kotlin/com/autonomousapps/Flags.kt
jjohannes: I reverted this line.

mvegter:
You will have to fix the `AssembleArchivesSpec` too, as it is now "failing" on the correct task tree:
```
Caused by: com.google.common.truth.AssertionErrorWithFacts: unexpected (4): :proj:compileJava, :proj:processResources, :proj:classes, :proj:jar |  
-- | --
  | --- |  
  | expected      : [:proj:assemble] |  
  | but was       : [:proj:compileJava, :proj:processResources, :proj:classes, :proj:jar, :proj:assemble]


```

jjohannes:
Thanks @mvegter I did not see that test. Very good that it exists. Fixed it.

autonomousapps:
I also forgot that that test exists! Now I don't need to bother manually testing this :)

PR #635, opened 2022-04-15
Don't downgrade dependency when we only know about it via an imperfect heuristic.

autonomousapps:


autonomousapps:
This turned out to be a little half-baked.

PR #641, opened 2022-04-18
Improve naming around declarations.

autonomousapps:


src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
```diff
@@ -126,20 +129,20 @@ class LocationMetadata(
...
 }
 
-internal class Locator(private val locationContainer: LocationContainer) {
-  fun locations(): Set<Declaration> {
-    return locationContainer.mapping.asSequence()
-      .filter { (name, _) -> isMain(name) || isAnnotationProcessor(name) }
+internal class Locator(private val declarationContainer: DeclarationContainer) {
+  fun declarations(): Set<Declaration> {
+    return declarationContainer.mapping.asSequence()
+      // .filter { (name, _) -> isForRegularDependency(name) || isForAnnotationProcessor(name) }
```
autonomousapps, line 139: This looks redundant, since the exact same filter is applied when creating the declaration container in the first place.

PR #646, opened 2022-04-21
Give advice to change to runtimeOnly where appropriate

### LINKED ISSUE #387: Should advise to change dependencies to runtimeOnly when that makes sense
**Plugin version**
0.72.0

**Gradle version**
6.8.3

**Describe the bug**
We are using this plugin on a big monolith Multi-Project and identified some dependencies that are not used by our code, but not reported as unused by the plugin. The following dependencies are affected:
```
dependencies {
    implementation 'org.apache.cxf:cxf-core:3.1.2'
    implementation 'org.apache.poi:poi-ooxml:3.17'
    implementation 'org.hibernate:hibernate-core:5.2.13.Final'
    implementation 'org.hibernate:hibernate-envers:5.2.13.Final'
    implementation 'org.junit.jupiter:junit-jupiter-engine:5.7.1'
    implementation 'org.junit.vintage:junit-vintage-engine:5.7.1'
    implementation 'org.postgresql:postgresql:42.2.4'
    implementation 'org.springframework.batch:spring-batch-core:4.2.1.RELEASE'
    implementation 'org.springframework:spring-core:5.2.4.RELEASE'
}
```

**To Reproduce**
I created a small sample project to reproduce this issue: 
1. git clone https://github.com/lschwahn/gradle-dep-analysis-unused-dependencies.git
2. ./gradlew buildHealth

```
> Task :buildHealth
Advice for root project
Looking good! No changes needed


BUILD SUCCESSFUL in 2s
26 actionable tasks: 25 executed, 1 up-to-date
```

None of the dependencies are declared as unused.

**Expected behavior**
Every dependency from the list above is reported as unused, as they are not used by the java-code inside the sample project.

jjohannes:
If an advice was computed to change a declaration from 'implementation'/'api' to 'runtimeOnly', this is indeed correct
information. The dependency is *not* required at compile time. But it *might* be required at runtime.

It is not possible for this plugin to make a clear statement about if it is actually needed at runtime. When getting such an advice, the user might also remove the dependency completely. The 'reason' feature will give more insight on why the dependency *might* be needed at runtime. The following cases are covered:

- Contains Service Loader declarations
- Contains Security Provider declarations
- Contains Content Provider or Service declarations (Android)
- Contains Android linters (Lint-Registry entry)

Fixes #387

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/android/LintJarSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/LintJarSpec.groovy
@@ -20,7 +20,7 @@ final class LintJarSpec extends AbstractAndroidSpec {
 
     then:
     assertThat(AdviceHelper.actualBuildHealth(gradleProject))
-      .containsExactlyElementsIn(AdviceHelper.emptyBuildHealthFor(':app'))
+      .containsExactlyElementsIn(project.expectedBuildHealth)
 
     where:
     [gradleVersion, agpVersion] << gradleAgpMatrix()
--- a/src/functionalTest/groovy/com/autonomousapps/android/projects/LintJarProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/projects/LintJarProject.groovy
@@ -1,8 +1,12 @@
 package com.autonomousapps.android.projects
 
 import com.autonomousapps.AbstractProject
+import com.autonomousapps.advice.Advice
+import com.autonomousapps.advice.ComprehensiveAdvice
 import com.autonomousapps.kit.*
 
+import static com.autonomousapps.AdviceHelper.compAdviceForDependencies
+import static com.autonomousapps.AdviceHelper.dependency
 import static com.autonomousapps.kit.Dependency.appcompat
 import static com.autonomousapps.kit.Dependency.rxlint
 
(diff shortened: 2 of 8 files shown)


autonomousapps:
I think I'm pretty happy with this. I wonder if we should add this kind of advice to a new section of the report, though, since I know there's a large category of users that will find it confusing. We already do this for `compileOnly`, [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt#L198) for example.

But I'm not entirely convinced it's necessary -- I say this to provoke discussion. Thoughts?

jjohannes:
I think the separate section is a great idea, because then the report can clarify that you **should** change something, but that you can choose between:

1. Change to `runtimeOnly`
2. Remove completely

How about this?

```
Dependencies which should be removed or changed to runtime-only:
  runtimeOnly("com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.13.2") (was implementation)
  runtimeOnly("org.apache.logging.log4j:log4j-api:2.17.2") (was implementation)
```

autonomousapps:
> How about this?

Yes! That looks great. Thanks :)

autonomousapps (approved):

autonomousapps:
@jjohannes the next thing we need is an update to `IssueHandler` so permit filtering out runtimeOnly advice. It'll be a lot of copy-pasta from the existing implementation to wire it through. I'm going to merge this, but would you mind doing that in a follow-up PR?

jjohannes:
Thanks for integrating this!

> next thing we need is an update to `IssueHandler` so permit filtering out runtimeOnly advice. 

I wasn't aware. I will have a look and create a follow-up PR.

Issue #451, opened 2021-08-12
Support custom source sets (source sets other than "main" and "test")


DotSpy:
Hello, in project we have integration tests specified by:
```kotlin

configure(integrationTests()) {
  sourceSets {
    create(integrationTest) {
      withConvention(org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet::class) {
        kotlin.srcDir("src/integrationTest/kotlin")
        resources.srcDir("src/integrationTest/resources")
        compileClasspath += sourceSets["main"].output + configurations["testRuntimeClasspath"]
        runtimeClasspath += output + compileClasspath + sourceSets["test"].runtimeClasspath
      }
    }
  }

  task<Test>(integrationTest) {
    description = "Runs the integration tests"
    group = "verification"
    testClassesDirs = sourceSets[integrationTest].output.classesDirs
    classpath = sourceSets[integrationTest].runtimeClasspath
    mustRunAfter(tasks["test"])
    useJUnitPlatform()
  }
}

```
in one of subprojects i have only integration test(no unit) and plugin suggest me to remove all testImplementation libraries that i'm using there, can i somehow specify custom source set for it or somehow specify that dependency analysis should consume compileIntegrationTestKotlin and compileIntegrationTestJava tasks?

autonomousapps:
The plugin currently provides no support for custom source sets. It only supports `main` and `test`. I'll consider this a feature request for a future release.

jjohannes:
This should work now with `v1.19.0`. (There might be more improvements with #854.)

Happy about feedback though.

autonomousapps:
I'm going to preemptively close this in the expectation that it works as desired. Feel free to reopen if you see issues!

Issue #411, opened 2021-06-09
False-positive redundant plugin when annotation processing is used only in androidTest


eugene-krivobokov:
**Plugin version**

0.73.0

**Gradle version**

6.9

**Android Gradle Plugin (AGP) version**

4.1.3

**Describe the bug**

False-positive `onRedundantPlugins` for `kotlin-kapt` plugin when it's used only in instrumentation tests (`androidTestImplementation`).

**To Reproduce**
A minimal reproducer: https://github.com/eugene-krivobokov/dependency-analysis-kapt-androidTests

**Expected behavior**

No error.

**Additional context**

autonomousapps:
@eugene-krivobokov could you try this again with 0.74.0? I've recently made some improvements to how test code and dependencies are processed.

eugene-krivobokov:
Unfortunately, still reproducible. I've updated a reproducer accordingly.

autonomousapps:
Part of the issue here is that the plugin currently offers no support for androidTests, and I simply don't have time to implement it right now. After my experiencing with normal tests, I know it's not a trivial improvement. As such, I will be working on https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/427 first, and removing this issue from milestone 1.0.

autonomousapps:
I have observed that, despite the plugin not supporting androidTest in any way, it can still emit advice about "unused" dependencies in the androidTest source set. So at the very least it needs to stop doing that for v1.0.

autonomousapps:
Sorry to keep going back and forth on this, but after some analysis I'm back to where I started: I'll implement androidTest support post-1.0. For anyone else who runs into this, you can configure the plugin to exclude such advice like so:
```groovy
// build.gradle
dependencyAnalysis {
  issues {
    all {
      onRedundantPlugins {
        exclude("kotlin-kapt")
      }
    }
  }
}
```

autonomousapps:
The associated PR finally adds support for androidTest analysis, but I don't yet have a functional test in place verifying this case. I will be adding that.

autonomousapps:
Fixed.

Issue #298, opened 2020-10-06
Plugin incorrectly(?) reporting unused dependency


Mahoney:
**Build scan link**
https://scans.gradle.com/s/wqljzkve3gazi

**Plugin version**
0.61.0

**Gradle version**
gradle-6.8-branch-eskatos_kotlin_embedded_1_4-20200922125722+0000

**Android Gradle Plugin (AGP) version**
N/A

**Describe the bug**
Getting a failure claiming I have an unused dependency that I think is invalid - possibly due to the plugin not knowing about the `testFixtures` configuration? Or not knowing about how kotlin works with top level functions?

**To Reproduce**
1. Check out git@github.com:Mahoney-playground/goos.git
2. Run ./gradlew build

**Expected behavior**
I think the build should pass

**Additional context**
https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures

autonomousapps:
You might be right about test fixtures. This plugin currently makes no accommodation for that.

You can get some insight into why the plugin is recommending something by eg:
```
./gradlew module:reason --id <identifier>
```
I'd give you an example for `<identifier>`, but you didn't tell me which dependency had been incorrectly declared unused :)

Mahoney:
Yes, sorry, that was a rubbish bug report - rushing to get to bed. Added the obvious extra details.

Mahoney:
I realise I've muddied the waters by using a snapshot version of gradle, but I see exactly the same behaviour reverting to gradle 6.6.1.

Mahoney:
Updated the bug report to a branch using gradle 6.6.1 - same behaviour.

autonomousapps:
Thanks, appreciated. I think this must be due to the plugin having no support yet for `testFixtures`. I will update this ticket to be a "feature request" to support that gradle feature.

For now, I think you can configure the plugin to ignore this "unused" dependency, following guidance in the wiki at https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Customizing-plugin-behavior

daanschipper:
I've done some more investigation into this issue. The `locateDependenciesMain` task does not take the capabilities of a project dependency into account when determining all declared dependencies.

[...]

PR #647, opened 2022-04-21
Spike: support additional source sets

jjohannes:
⚠️ _Just a spike to discuss the direction - current state breaks things (especially for Android)_ ⚠️ 

@autonomousapps after we had a short chat about this, I took a look at the code and just tried some things. I just wanted to share this with you. This draft PR might be thrown away (or used as base for a real implementation).

This is to tackle (or at least provide a solid foundation for):
- #451 
- #298
- #411 (`androidTest` support)

Somme comments below on my thoughts on what should be changed to make it work (I might be completely wrong with some things)

With this spike, I got the dependency analysis working in an example project for _test fixtures_.

```
  testFixturesApi("org.apache.zookeeper:zookeeper:3.8.0") (was testFixturesImplementation)
```


src/main/kotlin/com/autonomousapps/model/declaration/Variant.kt
```diff
@@ -1,35 +1,49 @@
...
  * 1. For the JVM, it is simply the source set (e.g., "main" and "test").
- * 2. For Android, it is the combination of _variant_ (e.g., "debug" and "release") and [SourceSetKind] ("main" and
+ * 2. For Android, it is the combination of _variant_ (e.g., "debug" and "release") and [SourceSet] ("main" and
  * "test").
  */
 data class Variant(
-  val variant: String,
-  val kind: SourceSetKind
+  val androidVariant: String,
+  val sourceSetName: String
```
jjohannes, line 14: This needs more clarification. I think if this class would get the two aspects clearly separated:
- sourceSetName
- androidVariantName (buildType+flavor)

And then this class could encapsulate the knowledge how these information is combined into a complete "Variant Name" used for Buckets and Tasks (see also `variantSpecificBucketName()` below). I think then the code using this could become cleaner. 

Right now there seems to be a bit of a mixup between when something is just a 'Source Set' and something is an 'Android Variant'.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -321,61 +315,33 @@ internal class ProjectPlugin(private val project: Project) {
...
-            )
-          } else {
-            JavaLibAnalyzer(
-              project = this,
-              sourceSet = sourceSet,
-              testSourceSet = null,
-              kind = SourceSetKind.TEST,
-              hasAbi = false
-            )
+      sourceSets.forEach { sourceSet ->
```
jjohannes, line 318: When we work with the "Source Set Name" instead of a "SourceSetType", for plain Java/Kotlin we can iterate all source sets.

src/main/kotlin/com/autonomousapps/model/declaration/Variant.kt
```diff
@@ -1,35 +1,49 @@
...
-    val MAIN = Variant(MAIN_NAME, SourceSetKind.MAIN)
-    //val TEST = Variant(TEST_NAME, SourceSetKind.TEST)
-    //val ANDROID_TEST = Variant(ANDROID_TEST_NAME, SourceSetKind.TEST)
+    return "${sourceSetName}${androidVariant.replaceFirstChar(Char::uppercase)}${bucket.capitalizeSafely()}"
+      .replaceFirstChar(Char::lowercase)
+  }
+
+  companion object {
+    // In non-Android builds there is only one 'variant' for each source set
+    const val BASE_VARIANT = "BASE"
```
jjohannes, line 42: This should probably go....

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -143,7 +142,7 @@ internal class ProjectPlugin(private val project: Project) {
       }
 
       unitTestSourceSets?.let { sourceSets ->
-        val variantSourceSet = newVariantSourceSet(name, SourceSetKind.TEST, sourceSets)
+        val variantSourceSet = newVariantSourceSet(name, SourceSet.TEST_SOURCE_SET_NAME, sourceSets)
```
jjohannes, line 145: For Android, we might (need to) keep the special handling of each source set we support (at leas initially?). Does adding arbitrary source sets make sense in Android anyway?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -408,55 +374,28 @@ internal class ProjectPlugin(private val project: Project) {
...
+                KotlinJvmAppAnalyzer(
+                  project = this,
+                  sourceSet = sourceSet
+                )
+              } else {
+                KotlinJvmLibAnalyzer(
+                  project = this,
+                  sourceSet = sourceSet,
+                  // TODO we could do bette by checking it '<sourceSetName>Api' configuration was setup
+                  hasAbi = !isTestSourceSet
```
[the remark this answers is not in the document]
autonomousapps: I don't think it's sufficient to check if the `...Api` configuration is created. From what I can tell, the Kotlin plugins create such a configuration even in cases where a project cannot logically have an API. For example, a project with the `com.android.application` plugin applied. Or a Java project with `application` applied. I want to enforce that only "library" projects can have an ABI.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
autonomousapps: Android can have arbitrary source sets up to a point. There are some docs [here](https://developer.android.com/studio/build/build-variants).

As far as I can tell, an Android project has support for source sets like the following:
```
// all standard. note that src/test and src/androidTest extend from src/main, 
// but do not relate to each other. This is an important point for computing 
// the advice in StandardTransform
src/main
src/test
src/androidTest

src/debug
src/release
src/flavorDebug
src/flavor1Flavor2Release
... etc.
```
In particular, note that only the production code can have variant-specific source-sets. Both `test` and `androidTest` do not have variants (I think). You can see some more examples [here](https://developer.android.com/studio/build/build-variants#sourceset-build).

Finally, I think it is highly unusual and possibly "impossible" for an Android project to have truly arbitrary source sets. The source sets you will find all relate precisely to some known combination of flavors and build types.

jjohannes: Thanks for the explanation. Took me a bit to wrap my head around this again. So the Android Plugin has its own SourceSet implementation and you can't see them when you do `the<SourceSetContainer>()`. (Not confusing at all when discussing this topic...)

But conceptually, from the dependency analysis perspective, both _source set_ concepts are similar.

I think if we put enough thought into this, it could be unified how the plugin handles any of these source sets and the associated configurations.

Do I understand correctly that the additional **Android** source sets are also not handled correctly right now?

For example, if I put code into "src/debug", that is not considered during the analysis, right? Or is there some special handling for this already?

jjohannes: Never mind. For Android there is one `Android*Analyzer` for each _source set_ already. 

Maybe we can then basically transfer that approach to the JVM world. 🤔 

What would need to be generalised somehow is the relationship between source sets. At the moment relationships between _source sets_ are hardcoded. For example between "Main" and "Test" (Test sees dependencies of Main) - for which `SourceSetType` is used right now IIUC.
I presume the relationship between "main" and "debug" (Debug seems dependencies of Main) is also encoded somewhere.

I am thinking that the analysis needs additional inputs like this:
- What does the "compileClasspath" extend from? From this we could find out which dependencies are included (e.g. Test includes "Main" and "Test")
- The list of all sourceSetNames (need to be computed differently for Android and JVM) from which we can make a better mapping from a configuration name to a "Variant" (and with that the associated compile tasks).
 
Then this can be fixed:

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/03a4ad7e6655beaecb75b1dabd85c4b87af08d8d/src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt#L40-L41

jjohannes: An 'issue' I face in my project is actually that I sometimes get wrong advice (e.g. for testFixture dependencies). 

It might already be a good improvement to address the TODO above by ignoring the unsupported source sets for now. Something that is done for 'androidTest' explicitly in some other place in the code already I think.

If we pass the names of the supported source sets into the analysis, this can probably work. I will give it a try and I might open a PR only for that. Might be a good starting point.

autonomousapps: Some clarifications:
> I presume the relationship between "main" and "debug" (Debug seems dependencies of Main) is also encoded somewhere.

In fact, this plugin never analyzes an Android project's "main" source directly. It analyzes every available _variant_ of an Android project. For simplicity, imagine there are two (the most common case): `debug` and `release`. Therefore the plugin analyzes the debug variant and the release variant. The plugin's integration with AGP ensures it analyzes all the relevant source sets for each of these variants.

> What does the "compileClasspath" extend from? From this we could find out which dependencies are included (e.g. Test includes "Main" and "Test")

I'm a little worried this won't work as expected. A few months ago I tried building a DAG of configurations based on `extendsFrom` relationships, whereupon I discovered that AGP does not use that consistently. There are a number of "floating" configurations that have logical relationships with each other, that are _not_ encoded with any real API, but rather custom handling inside AGP. The relationships are deterministic, however, which is why I just hand-code some algorithms for managing all that.

> It might already be a good improvement to address the TODO above by ignoring the unsupported source sets for now. Something that is done for 'androidTest' explicitly in some other place in the code already I think.

I agree.

jjohannes: > I tried building a DAG...

If you have already been down that road, then it is probably not helpful to pursue this further.
Conceptually, I also wonder if this "extendsFrom" approach is even something to recommend. It always was like this for normal "test" dependencies in Gradle (which it also got from Maven I presume).
in case of _custom source sets_, It is very unclear what has an effect on what if you start to add your own "extendsFrom" statements. Because you, as the person that defines dependencies, would also need to know the extendsFrom-hierarchy to understand what you are doing. Which is quite hidden, if it does no longer follow the standard api/implementation etc. setup.
If I define my own source sets (e.g. for integration tests), I always prefer to not do that. And instead declare the needed dependencies explicitly again. Which is much in line what the plugin proposes I think -> Always directly declare what you use.
Maybe for _custom source sets_ the plugin should not attempt to derive anything from addition _extendsFrom_. And just treat them as if they are 'isolated'. Maybe later the plugin could offer a configuration option were users can explicitly say that they do not want declare dependencies for a certain source set if they are declared in for another source set already. 

> I agree.

Hope this goes in the right direction: #651

PR #651, opened 2022-04-25
Ignore dependencies for and to unsupported sourceSets/variants

jjohannes:
The plugin currently may give wrong advices for source sets that are not yet supported (see #451 and #298).

This PR:
- Adds tests to show the issues.
- Ignores the unsupported cases for now so that users can use these Gradle features and still use this plugin to analyse the dependencies of the source sets that are supported right now.

Problematic cases are:

### Dependencies from unsupported Source Sets/Variants

The plugin produces advices (often wrong) for source sets that are not
yet supported. For example:

```
testFixturesApi("org.apache.commons:commons-collections4:4.4")

extraFeatureApi("org.apache.commons:commons-collections4:4.4")
```

### Dependencies to unsupported Source Sets/Variants

The plugin may produce wrong advices by ignoring the 'requires capabilities' part of dependencies like this:

```
testImplementation(testFixtures(project(':proj')))

api(project(':proj')) { capabilities {
     requireCapabilities('examplegroup:proj-extra-feature')
} }
```

Note: testFixtures(...) also uses 'requires capabilities'

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
@@ -10,6 +10,7 @@ import com.autonomousapps.kit.SourceType
 
 import static com.autonomousapps.AdviceHelper.*
 import static com.autonomousapps.kit.Dependency.commonsCollections
+import static com.autonomousapps.kit.Dependency.project
 
 final class FeatureVariantTestProject extends AbstractProject {
 
@@ -30,6 +31,16 @@ final class FeatureVariantTestProject extends AbstractProject {
         bs.dependencies = [
           commonsCollections('api'),
           commonsCollections('extraFeatureApi')
+        ]
+        bs.additions = 'group = "examplegroup"'
+      }
+    }
+    builder.withSubproject('consumer') { s ->
+      s.sources = consumerSources
+      s.withBuildScript { bs ->
+        bs.plugins = [Plugin.javaLibraryPlugin]
+        bs.dependencies = [
+          project('api', ':proj', 'examplegroup:proj-extra-feature')
         ]
       }
     }
@@ -67,12 +78,34 @@ final class FeatureVariantTestProject extends AbstractProject {
     )
   ]
 
+  private consumerSources = [
+    new Source(
+      SourceType.JAVA, "Consumer", "com/example/consumer",
+      """\
+        package com.example.consumer;
+        
+        import com.example.extra.ExtraFeature;
+        
+        public class Consumer {
+          private ExtraFeature extra;
+        }
+      """.stripIndent()
+    )
+  ]
+
   @SuppressWarnings('GroovyAssignabilityCheck')
   List<ComprehensiveAdvice> actualBuildHealth() {
     actualBuildHealth(gradleProject)
   }
 
+  // Note: The 'proj-extra.jar' is considered part of the 'main variant' of ':proj', which is not correct.
+  private final Set<Advice> expectedConsumerAdvice = [
+    Advice.ofAdd(transitiveDependency(dependency: ':proj'), 'implementation'),
+  ]
+
   final List<ComprehensiveAdvice> expectedBuildHealth = [
+    // Not yet implemented: missing advice to move the dependency to 'extra' to implementation
+    compAdviceForDependencies(':consumer', expectedConsumerAdvice),
     // Not yet implemented: missing advice to move the dependency of 'extra' to extraFeatureImplementation
     emptyCompAdviceFor(':proj')
   ]
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
@@ -1,16 +1,15 @@
 package com.autonomousapps.jvm.projects
 
 import com.autonomousapps.AbstractProject
-import com.autonomousapps.advice.Advice
-import com.autonomousapps.advice.ComprehensiveAdvice
 import com.autonomousapps.kit.GradleProject
 import com.autonomousapps.kit.Plugin
 import com.autonomousapps.kit.Source
 import com.autonomousapps.kit.SourceType
+import com.autonomousapps.model.Advice
+import com.autonomousapps.model.ProjectAdvice
 
 import static com.autonomousapps.AdviceHelper.*
-import static com.autonomousapps.kit.Dependency.commonsCollections
-import static com.autonomousapps.kit.Dependency.project
+import static com.autonomousapps.kit.Dependency.*
 
 final class FeatureVariantTestProject extends AbstractProject {
 
@@ -22,7 +21,7 @@ final class FeatureVariantTestProject extends AbstractProject {
 
   private GradleProject build() {
     def builder = newGradleProjectBuilder()
-    builder.withSubproject('proj') { s ->
+    builder.withSubproject('producer') { s ->
       s.sources = sources
       s.withBuildScript { bs ->
         bs.plugins = [Plugin.javaLibraryPlugin]
@@ -40,7 +39,7 @@ final class FeatureVariantTestProject extends AbstractProject {
       s.withBuildScript { bs ->
         bs.plugins = [Plugin.javaLibraryPlugin]
         bs.dependencies = [
-          project('api', ':proj', 'examplegroup:proj-extra-feature')
+          project('api', ':producer', 'examplegroup:producer-extra-feature')
         ]
       }
     }
@@ -93,21 +92,22 @@ final class FeatureVariantTestProject extends AbstractProject {
     )
   ]
 
-  @SuppressWarnings('GroovyAssignabilityCheck')
-  List<ComprehensiveAdvice> actualBuildHealth() {
-    actualBuildHealth(gradleProject)
+  Set<ProjectAdvice> actualBuildHealth() {
+    return actualProjectAdvice(gradleProject)
   }
 
-  // Note: The 'proj-extra.jar' is considered part of the 'main variant' of ':proj', which is not correct.
+  // Note: 'producer-extra-feature.jar' is considered part of the 'main variant' of ':producer', which is not correct.
   private final Set<Advice> expectedConsumerAdvice = [
-    Advice.ofAdd(transitiveDependency(dependency: ':proj'), 'implementation'),
+    Advice.ofAdd(projectCoordinates(':producer'), 'implementation'),
   ]
 
-  final List<ComprehensiveAdvice> expectedBuildHealth = [
-    // Not yet implemented: missing advice to move the dependency to 'extra' to implementation
-    compAdviceForDependencies(':consumer', expectedConsumerAdvice),
-    // Not yet implemented: missing advice to move the dependency of 'extra' to extraFeatureImplementation
-    emptyCompAdviceFor(':proj')
+  final Set<ProjectAdvice> expectedBuildHealth = [
+    // Not yet implemented:
+    // missing advice to move dependency 'consumer' -> 'producer-extra-feature' to implementation
+    projectAdviceForDependencies(':consumer', expectedConsumerAdvice),
+    // Not yet implemented:
+    // missing advice to move dependency 'producer-extra' -> 'commons-collections4' to extraFeatureImplementation
+    emptyProjectAdviceFor(':producer')
   ]
 
 }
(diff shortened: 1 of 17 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/IntegrationTestProject.groovy
```diff
@@ -133,7 +134,7 @@ final class IntegrationTestProject extends AbstractProject {
     emptyCompAdviceFor(':lib'),
     emptyCompAdviceFor(':core'),
     new ComprehensiveAdvice(':proj', [
-      Advice.ofChange(dependency(project('integrationTestImplementation', ':core')), 'api')
+      Advice.ofAdd(transitiveDependency(dependency(':core'), []), 'api')
```
jjohannes, line 137: This expectation was "wrong" before. The 'proj.integrationTest' sources do depend on 'core'. So the dependency should not be moved from there.

Unless you assume that 'integrationTest' always inherits all 'implementation' dependencies and the rule is that you do not need to redeclare them then. But this may not be the case (it is not in this test case). There is no guarantee that the setup is like this and thus such assumptions should not be made.

If #451 is implemented, further thoughts could be put into this.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -141,50 +141,46 @@ internal fun ComponentIdentifier.resolvedVersion(): String? = when (this) {
  * Given [Configuration.getDependencies], return this dependency set as a set of identifiers, per
  * [ComponentIdentifier.toIdentifier].
  */
-internal fun DependencySet.toIdentifiers(
-  metadataSink: MutableMap<String, Boolean> = mutableMapOf()
```
jjohannes, line 145: I think the idea with the metadata can not work like it was here. You would need the metadata for each dependency. E.g. if you have both:

```
api("org.apache.commons:commons-collections4:4.4")
testFixturesApi("org.apache.commons:commons-collections4:4.4")
```

The two declaration would have different metadata. But the approach here only stores one "Metadata" for `org.apache.commons:commons-collections4:4.4`.

It also feels to me like Gradle's dependency resolution should be used here to give you the result (i.e. to which Jar on the other side will the dependency resolve).
If you start storing attributes/capabilities here yourself and pass them around you will somehow end up re-implementing Gradle's variant matching to cover all the cases.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -141,50 +141,46 @@ internal fun ComponentIdentifier.resolvedVersion(): String? = when (this) {
...
-): Set<String> = mapNotNullToSet {
-  it.toIdentifier(metadataSink)
+internal fun DependencySet.toIdentifiers(): Set<Pair<String, Boolean>> = mapNotNullToSet {
```
jjohannes, line 144: The `Pair<String, Boolean>` now stores the information if the dependency is supported or should be ignored for now.

In the future that could then be replaced by a Type that holds more information (metadata of the dependency). For example, the Jar the dependency will resolve to (see comment above).

src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
```diff
@@ -85,40 +82,20 @@ abstract class FindDeclarationsTask : DefaultTask() {
     ): Sequence<Configuration> {
       val seq = configurations.asSequence()
         .filter { it.isForRegularDependency() || it.isForAnnotationProcessor() }
-        .filterNot {
-          // this is not ideal, but it will solve some problems till we can support androidTest analysis.
-          // will match, e.g., "androidTestImplementation", "debugAndroidTestImplementation", and "kaptAndroidTest".
-          it.name.contains("androidTest", true)
-        }
```
jjohannes, line 92: This is no longer needed, because it is covered by the more general "supported source sets" implementation.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -206,11 +202,21 @@ internal fun Dependency.resolvedVersion(): String? = when (this) {
   else -> throw GradleException("Unknown Dependency subtype: \n$this\n${javaClass.name}")
 }?.intern()
 
-private fun Dependency.isJavaPlatform(): Boolean = when (this) {
-  is ProjectDependency -> dependencyProject.pluginManager.hasPlugin("java-platform")
```
jjohannes, line 210: This is not needed IMO. The **ProjectDependency** is also a **ModuleDependency** and it will have the attribute added by `testFixtures(...)` which is checked by the code below.

src/test/groovy/com/autonomousapps/model/declaration/VariantSpec.groovy
```diff
@@ -6,25 +6,28 @@ class VariantSpec extends Specification {
...
+      'main', 'release', 'debug', 'test', 'testDebug', 'testRelease',
+      'flavor', 'flavorRelease', 'flavorDebug', 'testFlavorRelease', 'testFlavorDebug'
+    ] as Set<String>) == variant
 
     where:
     configurationName                  | variant
     'implementation'                   | new Variant('main', SourceSetKind.MAIN)
     'debugImplementation'              | new Variant('debug', SourceSetKind.MAIN)
-    'releaseFlavorImplementation'      | new Variant('releaseFlavor', SourceSetKind.MAIN)
+    'flavorReleaseImplementation'      | new Variant('flavorRelease', SourceSetKind.MAIN)
```
jjohannes, line 18: I checked with the [Android docs](https://developer.android.com/studio/build/dependencies#dependency_configurations) and with an example project. The naming scheme is always:
`$sourceSet$flavor$buildType$scope`

(sourceSet is empty in case of 'main')

IMO some of the test data here was wrong (switching 'flavor' and 'buildType' order). These won't work anymore as they do not exist in a real Android project.

autonomousapps: Seems like I had a copy-paste error.

autonomousapps (commented):
I _really_ appreciate this. I want to check out the changes locally and study them a bit to ensure I understand what you've done, but my impression is that it's an improvement.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -141,50 +141,46 @@ internal fun ComponentIdentifier.resolvedVersion(): String? = when (this) {
...
   }
   is FileCollectionDependency -> {
     // Note that this only gets the first file in the collection, ignoring the rest.
     when (files) {
       is ConfigurableFileCollection -> (files as? ConfigurableFileCollection)?.from?.let { from ->
         from.firstOrNull()?.toString()?.substringAfterLast("/")
-      }
-      is ConfigurableFileTree -> files.firstOrNull()?.name
+      }?.let { Pair(it, false) }
+      is ConfigurableFileTree -> files.firstOrNull()?.name?.let { Pair(it, false) }
```
autonomousapps, line 183: I think we lost the `intern()` calls on these two types.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -206,11 +202,21 @@ internal fun Dependency.resolvedVersion(): String? = when (this) {
...
     category?.name == Category.REGULAR_PLATFORM || category?.name == Category.ENFORCED_PLATFORM
   }
   else -> throw GradleException("Unknown Dependency subtype: \n$this\n${javaClass.name}")
 }
+
+/**
+ * Return true if the dependency does not point at the "main" variant of the other module.
+ * Other variants are for example testFixtures() or Feature Variants with different capabilities.
+ */
+private fun Dependency.isDependencyToNonMainFeature(): Boolean = when (this) {
```
autonomousapps, line 217: Could you add links to the documentation for test fixtures and feature variants in your kdoc? I have a fairly verbose style when it comes to this documentation because the domain is so complex and it's easy to forget context as time passes.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -692,6 +692,25 @@ internal class ProjectPlugin(private val project: Project) {
     }
   }
 
+  private fun Project.supportedSourceSetNames() = provider {
```
autonomousapps, line 695: When I checked the explicit type of this, I got `Provider<Collection<String>>`, when I would have expected `Provider<Set<String>>` (from line 710). I think we can achieve that (which I prefer), with this small change:
```kotlin
the<AppExtension>().applicationVariants.flatMapToSet {
  it.sourceSets.map { sourceSet -> sourceSet.name } +
    (it.unitTestVariant?.sourceSets?.map { sourceSet -> sourceSet.name } ?: emptySet())
  // Not yet supported: + it.testVariant.sourceSets.map { sourceSet -> sourceSet.name }
}
```
Note the use of the `flatMapToSet` extension function.

(the same change works for the `LibraryExtension` section.

autonomousapps: And then as a style matter, could you move this whole function _below_ the one that uses it? Semantically, I see this as more of a helper function, and therefore less important than the `configureAggregationTasks()` function.

autonomousapps (approved):
Especially given that all the tests pass, I'm very comfortable with these changes. I've left a few notes -- would you mind taking a look?

autonomousapps:
FYI, I just [merged a PR](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/656) that will conflict with `IntegrationTestProject`. Should be straightforward to resolve, and hopefully you'll agree it has simplified test fixture setup a bit :)

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
jjohannes: I think, if you want to get to a working solution here later, you should investigate how to do this:

> It also feels to me like Gradle's dependency resolution should be used here to give you the result (i.e. to which Jar on the other side will the dependency resolve).

It's not the goal of this PR to solve this. But feel free to ping me if you get to it later. Happy to discuss/explore this.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
```diff
@@ -0,0 +1,113 @@
...
+      """.stripIndent()
+    )
+  ]
+
+  @SuppressWarnings('GroovyAssignabilityCheck')
+  List<ComprehensiveAdvice> actualBuildHealth() {
+    actualBuildHealth(gradleProject)
+  }
+
+  // Note: The 'proj-extra.jar' is considered part of the 'main variant' of ':proj', which is not correct.
```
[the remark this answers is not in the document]
jjohannes: In general yes. Although I haven't looked in detail, why this particular thing is happening. This is not changed by this PR. The test just reveals the existing behavior.

IMO this particular "false positive" is not super critical. Because often if you use an "extra feature" of a project, you also use the main variant and would have that dependency anyway.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantTestProject.groovy
```diff
@@ -0,0 +1,113 @@
...
+  private final Set<Advice> expectedConsumerAdvice = [
+    Advice.ofAdd(transitiveDependency(dependency: ':proj'), 'implementation'),
+  ]
+
+  final List<ComprehensiveAdvice> expectedBuildHealth = [
+    // Not yet implemented: missing advice to move the dependency to 'extra' to implementation
```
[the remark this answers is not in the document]
jjohannes: No I meant 'to', but the word 'to' is used much too often in that sentence for anyone to understand. 😄 I expanded the comments so that it is (hopefully) clearer what is missing.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
jjohannes: Good catch. Sorry for that. Added them.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
jjohannes: Yes that's better. I made the changes and also added a short Javadoc comment to the function.

jjohannes:
Thank you for the detailed review @autonomousapps. I rebased and addressed the feedback. Let me know if there is anything else you would like to have addressed before merging this.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
autonomousapps: I think one of the problems was doing this in the context of _declarations_, right? At that point, I don't want to _resolve_ any dependencies at all. I just want to know what the user _declared_.

Issue #648, opened 2022-04-22
compileOnly: report unused compileOnly dependencies as 'should be removed'


jjohannes:
**Is your feature request related to a problem? Please describe.**

If we define dependencies like this but do not use them at all they are not reported.

```
    compileOnly("androidx.annotation:annotation")
    compileOnly("org.json:json")
```

The plugin currently discards all advices for dependencies declared as `compileOnly`. I think this is the right behavior if it is about cases where the dependencies _could_ be `implementation` or `api`. Because the plugin cannot know, if a dependency is (not) needed at runtime.

However, the plugin would know if the dependency is not needed at compile time. In this case, the dependency can be safely removed altogether and I think the plugin should give that advice.

**Describe the solution you'd like**

The plugin should list `compileOnly` dependencies that are unused in the _Unused dependencies which should be removed_ part of the report.

**Describe alternatives you've considered**

_..._

**Additional context**

A related topic I just thought of (but this is really a very rare case I assume) :
- `compileOnly` dependencies which are detected as _api_ should be recommended as `compileOnlyApi`
- `compileOnlyApi` dependencies which are detected as _implementation_ should be recommended as `compileOnly`

autonomousapps:
I agree with your general idea.

Secondly, I do want to eventually support `compileOnlyApi`, but care must be taken because only the core Gradle plugins support this currently. At least, I don't think that Kotlin nor AGP support it.

PR #660, opened 2022-05-06
Put 'compileOnlyApi' and 'providedCompile' into the COMPILE_ONLY bucket

jjohannes:
Similar to #651, the goal of this to avoid **wrong advices** to be reported rather than fully supporting what could be possible here (a follow up could be #648). The test in this PR shows an example where the plugin right now suggests moving dependencies aways from "compileOnlyApi" or "providedCompile".

From the perspective of this plugin right now, dependencies defined as 'compileOnlyApi' or 'providedCompile' should be treated the same as 'compileOnly'. Which is, there are no advices produced for them.

Some context on 'providedCompile':
The configuration is added by the 'war' Gradle core plugin if applied. This configuration exists for years already. If it would be introduced today, it would probably be called 'providedApi'. It has the same behaviour as the 'api' configuration, but with the difference that it is not packaged into the 'war' file that makes up the final application. Because the Jar behind the dependency is expected to be provided by the runtime environment. Currently, there is no 'providedImplementation'. So an analysis whether a 'providedCompile' dependency has "api" or "implementation" characteristics won't be useful because there is not distinction right now.

Using the 'war' plugin and 'providedCompile' can be seen as a bit of an edge case. But there are many Java (Enterprise) projects out there using it as it exists for a long time. This change would make it easier for such projects to use this plugin without having to manually exclude certain analysis results.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/CompileOnlyJarSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/CompileOnlyJarSpec.groovy
@@ -1,6 +1,7 @@
 package com.autonomousapps.jvm
 
 import com.autonomousapps.jvm.projects.CompileOnlyJarProject
+import com.autonomousapps.jvm.projects.WarTestProject
 
 import static com.autonomousapps.utils.Runner.build
 import static com.google.common.truth.Truth.assertThat
@@ -14,6 +15,27 @@ final class CompileOnlyJarSpec extends AbstractJvmSpec {
 
     when:
     build(gradleVersion, gradleProject.rootDir, ':external:jar')
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  // The plugin cannot decide if something that is required for compilation is only needed at compile time.
+  // Currently, such dependencies produce now advice at all. In the future the plugin could:
+  // - Give an advice if one of these dependencies can be removed completely
+  //   https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/648
+  // - Give an advice id a dependency could be moved between 'compileOnly' <-> 'compileOnlyApi'
+  //   (in projects where 'compileOnlyApi' exists)
+  def "no advices for compileOnly, compileOnlyApi and providedCompile"() {
+    given:
+    def project = new WarTestProject()
+    gradleProject = project.gradleProject
+
+    when:
     build(gradleVersion, gradleProject.rootDir, 'buildHealth')
 
     then:
--- a/src/main/kotlin/com/autonomousapps/model/declaration/Bucket.kt
+++ b/src/main/kotlin/com/autonomousapps/model/declaration/Bucket.kt
@@ -4,11 +4,9 @@ package com.autonomousapps.model.declaration
 internal enum class Bucket(val value: String) {
   API("api"),
   IMPL("implementation"),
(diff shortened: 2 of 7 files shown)


src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
```diff
@@ -77,7 +82,11 @@ internal object Configurations {
 
   // we want dependency buckets only
   fun Configuration.isForRegularDependency() =
-    !isCanBeConsumed && !isCanBeResolved && isForRegularDependency(name)
+    // do not check '!isCanBeConsumed && !isCanBeResolved' due to https://github.com/gradle/gradle/issues/20547 or
+    // similar situations. Other plugins or users (although not recommended) might change these flags. Since we know
+    // the exact names of the Configurations we support (based on to which source set they are linked) this check
+    // is not necessary.
+    isForRegularDependency(name)
```
jjohannes, line 89: I hope your are comfortable with this change.

src/main/kotlin/com/autonomousapps/model/declaration/Bucket.kt
```diff
@@ -4,12 +4,10 @@ package com.autonomousapps.model.declaration
 internal enum class Bucket(val value: String) {
   API("api"),
   IMPL("implementation"),
+  // These configurations go into the compileOnly bucket: '...CompileOny', '...CompileOnlyApi', 'providedCompile'
   COMPILE_ONLY("compileOnly"),
```
jjohannes, line 8: This probably changes the meaning of what `Bucket` is supposed to mean here. I assume it is meant to be a 1-to-1 relationship between a _Bucket_ and a _Configuration_.
Adding three Buckets here, would mean three buckets that are treated the same in a bunch of other places. Could also be an option but it felt like a lot of copy/paste to me for this "small" improvement.

My thinking was to do it like this for now (compact solution covering what the plugin is capable right now) and adjust it more later if a distinction between "compileOnly" and "compileOnlyApi" becomes important.

For #648 (suggesting removals) all these three Configurations can be treated the same. So this implementation would work as a base for that.

src/functionalTest/groovy/com/autonomousapps/jvm/CompileOnlyJarSpec.groovy
```diff
@@ -22,4 +23,25 @@ final class CompileOnlyJarSpec extends AbstractJvmSpec {
     where:
     gradleVersion << gradleVersions()
   }
+
+  // The plugin cannot decide if something that is required for compilation is only needed at compile time.
+  // Currently, such dependencies produce now advice at all. In the future the plugin could:
```
autonomousapps, line 28: nit: `produce now advice` -> `produce no advice`

src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
```diff
@@ -10,7 +10,8 @@ internal object Configurations {
   internal const val CONF_ADVICE_ALL_CONSUMER = "adviceAllConsumer"
   internal const val CONF_ADVICE_ALL_PRODUCER = "adviceAllProducer"
 
-  private val MAIN_SUFFIXES = listOf("api", "implementation", "compileOnly", "runtimeOnly")
+  private val COMPILE_ONLY_SUFFIXES = listOf("compileOnly", "compileOnlyApi", "providedCompile")
```
autonomousapps, line 13: I think this is fine in practice, but in the interest of thinking aloud, it seems unlikely that Android would ever support something like `debugProvidedCompile`. And of course android-land is the reason we have to deal with things like suffixes and prefixes...

src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
autonomousapps: The comment + a comprehensive test suite gives me great confidence :)

autonomousapps (approved):

PR #666, opened 2022-05-09
Support a primary entry point for bundles.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/662.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/BundleProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/BundleProject.groovy
@@ -1,0 +1,155 @@
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Plugin
+import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.SourceType
+import com.autonomousapps.model.Advice
+import com.autonomousapps.model.ProjectAdvice
+
+import static com.autonomousapps.AdviceHelper.*
+import static com.autonomousapps.kit.Dependency.project
+
+final class BundleProject extends AbstractProject {
+
+  final GradleProject gradleProject
+
+  BundleProject() {
+    this.gradleProject = build()
+  }
+
+  private GradleProject build() {
+    def builder = newGradleProjectBuilder()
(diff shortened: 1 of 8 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/BundleProject.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+  private GradleProject build() {
+    def builder = newGradleProjectBuilder()
+    builder.withRootProject {
+      it.withBuildScript { bs ->
+        bs.additions = """
+          dependencyAnalysis {
+            dependencies {
+              bundle('facade') {
+                primary(':entry-point')
```
autonomousapps, line 30: This is what the new API looks like in practice. Note that I haven't added an override to the `bundle()` function, primarily due to limitations in Gradle that make it annoying to create domain objects with custom constructors.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.autonomousapps.model.intermediates.Usage
+
+/**
+ * :proj
+ * |
+ * B -> unused, not declared, but top of graph (added by plugin)
+ * |
+ * C -> used as API, part of bundle with B. Should not be declared!
+ */
+internal class Bundles(private val dependencyUsages: Map<Coordinates, Set<Usage>>) {
```
autonomousapps, line 20: Moved from elsewhere for easier testing.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -0,0 +1,120 @@
...
+      projectNode: ProjectCoordinates,
+      dependencyGraph: Map<String, DependencyGraphView>,
+      bundleRules: SerializableBundles,
+      dependencyUsages: Map<Coordinates, Set<Usage>>,
+      ignoreKtx: Boolean
+    ): Bundles {
+      val bundles = Bundles(dependencyUsages)
+
+      // Handle bundles with primary entry points
+      bundleRules.primaries.forEach { (name, primaryId) ->
```
autonomousapps, line 72: Some ugly duplication here, but it works ™️

PR #669, opened 2022-05-12
Add a trailing space for path linkifying

ZacSweers:
Terminal UIs often only linkify paths if there's a trailing space, so this should help avoid these not linking otherwise


autonomousapps (approved):

autonomousapps:
I think this may also resolve https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/627.

PR #670, opened 2022-05-12
Clarify the getting started instructions

### LINKED ISSUE #668: Task projectHealth is not created
**Build scan link**
https://scans.gradle.com/s/5egkb67qz3e5a

**Plugin version**
1.2.0

**Gradle version**
7.3.3

This is not for an android project.

**Describe the bug**
No advice is generated, only two tasks are created: `buildHealth` and `generateBuildHealth`.
The task `projectHealth` is not created:

```

* What went wrong:
Task 'projectHealth' not found in root project 'bitbucket-rest'.
```

**To Reproduce**

```bash
git clone git@github.com:martinda/bitbucket-rest.git --branch fix-api
cd bitbucket-rest
./gradlew buildHealth
./gradlew projectHealth
```

**Expected behavior**
I expect the task `projectHealth` to exist, but I don't understand why it doesn't.

martinda:
Fix #668 with getting started instructions for the new user.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -18,13 +18,14 @@ behaviors, but they may be interesting for some advanced users.
 For detailed instructions, see
 https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Adding-to-your-project[the wiki].
 
-The simplest approach is to add the following:
+The simplest approach is to add the plugin `build.gradle` as shown below.
+For that to work the [supported Java plugins](../../wiki/Compatibilities-&-limitations) must also be in your build.
 
 .root build.gradle
 [source,groovy]
 ----
 plugins {
-  id 'java-library' // or 'application'
+  id '<<supported java plugin(s)>>'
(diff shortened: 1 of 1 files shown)


autonomousapps (changes requested):
I understand your intent here, but this isn't correct advice to give to all users of this plugin.

The documentation you're looking for is in the wiki [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Compatibilities-&-Limitations). If you think that could be improved, I'm open to suggestions. Unfortunately, it isn't possible to PR against the wiki :-/

martinda:
Maybe this can be introduced differently, perhaps a statement saying that at the minimum one of the supported plugins must be applied, and tell the user to see the wiki for which ones.

autonomousapps:
> Maybe this can be introduced differently, perhaps a statement saying that at the minimum one of the supported plugins must be applied, and tell the user to see the wiki for which ones.

Sure. Do you want to propose a different change to the readme?

PR #681, opened 2022-05-28
WIP. Use SourceSet.output for better JVM support.

autonomousapps:
Proof of concept that, if it works, will mean a much easier time adding support for new JVM languages (such as Groovy and Scala).


autonomousapps:
Build failed https://gradle.com/s/olqm5n72dx7su

PR #688, opened 2022-06-01
Support Components with multiple Jar files

jjohannes:
In Gradle, one component - i.e. one logical component identified by GA coordinates or Gradle project path - can have multiple Jars. So far, the plugin would silently ignore all Jars except for the first it finds. With this fix, all Jars are considered part of the component during the analysis.

This used to work in older versions; tested with 0.71.0

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
@@ -1,0 +1,24 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.JarTransformingProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class ComponentWithMultipleArtifactsSpec extends AbstractJvmSpec {
+
+  def "one component can have multiple Jars (#gradleVersion)"() {
+    given:
+    def project = new JarTransformingProject()
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+}
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
@@ -1,6 +1,7 @@
 package com.autonomousapps.jvm
 
 import com.autonomousapps.jvm.projects.JarTransformingProject
+import com.autonomousapps.jvm.projects.MultipleJarsProject
 
 import static com.autonomousapps.utils.Runner.build
 import static com.google.common.truth.Truth.assertThat
@@ -8,6 +9,21 @@ import static com.google.common.truth.Truth.assertThat
 final class ComponentWithMultipleArtifactsSpec extends AbstractJvmSpec {
 
   def "one component can have multiple Jars (#gradleVersion)"() {
+    given:
+    def project = new MultipleJarsProject()
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  def "one component can have multiple Jars produced by a transform (#gradleVersion)"() {
     given:
     def project = new JarTransformingProject()
     gradleProject = project.gradleProject
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/JarTransformingProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/JarTransformingProject.groovy
@@ -1,0 +1,113 @@
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Plugin
+import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.SourceType
+import com.autonomousapps.model.Advice
+import com.autonomousapps.model.ProjectAdvice
+
+import static com.autonomousapps.AdviceHelper.*
+import static com.autonomousapps.kit.Dependency.commonsCollections
+
+final class JarTransformingProject extends AbstractProject {
+
+  final GradleProject gradleProject
+
+  JarTransformingProject() {
+    this.gradleProject = build()
+  }
+
+  private def commonsCollections = commonsCollections("api")
+
+  private GradleProject build() {
+    def builder = newGradleProjectBuilder()
+    builder.withSubproject('proj') { s ->
+      s.sources = PROJ_SOURCES
+      s.withBuildScript { bs ->
+        bs.plugins = [Plugin.javaLibraryPlugin]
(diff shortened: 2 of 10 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/JarTransformingProject.groovy
```diff
@@ -0,0 +1,113 @@
...
+          dependencies.artifactTypes.jar.attributes.attribute(split, false)
+          dependencies.registerTransform(SplitJarTransform) {
+            from.attribute(split, false).attribute(artifactType, "jar")
+            to.attribute(split, true).attribute(artifactType, "jar")
+          }
+          configurations.compileClasspath.attributes.attribute(split, false)
+          configurations.runtimeClasspath.attributes.attribute(split, false)
+          
+          afterEvaluate {
+            tasks.named("artifactsReportMain") {
```
autonomousapps, line 50: So, for the plugin to work in the context of this Gradle feature, users would have to manually configure this task?

src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
```diff
@@ -43,7 +43,7 @@ internal class JarExploder(
   private fun explodeJar(artifact: PhysicalArtifact): ExplodingJar {
     val zip = ZipFile(artifact.file)
 
-    val alreadyExplodingJar: ExplodingJar? = inMemoryCache.explodedJar(artifact.coordinates.toString())
+    val alreadyExplodingJar: ExplodingJar? = inMemoryCache.explodedJar(artifact.toString())
```
autonomousapps, line 46: Can you explain this change?

jjohannes: I think I misunderstood parts of the code here. This change has no effect as it is.

But I think, the previous code is also wrong. Later we have this:

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/b7a7fbc71b15b6cc7277dd8402a99ab66b5abc24/src/main/kotlin/com/autonomousapps/internal/JarExploder.kt#L76

So shouldn't this be `artifact.file.name`? I think right now (with or without my change) `inMemoryCache.explodedJar(...)` always returns `null`.

Using just the "file name" as key might also not be enough though. Maybe both places should be changed to use `absolutePath`. Wdyt?

(I made this change now, but can revert if I misunderstood the intentions.)

jjohannes: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/688/commits/0083fddc453f82e15adc096d4f104d5ee0a36feb

src/functionalTest/groovy/com/autonomousapps/jvm/projects/JarTransformingProject.groovy
jjohannes: Well the situation can also be created differently. As I wrote in one comment: (one Variant of) one _Component_ can have several Jars. Other ways to produce this is to have published Metadata with more than one Jar in the _api_ (compile classpath) variant. Or by having a local project that, for whatever reason, produces more than one Jar for one variant. In these cases, the Jars will just be on the classpath and no adjustment would be needed to the _report_ task inputs.

(These are _mostly_ theoretical cases as in a default Java project, you won't do this. But you know how things go in large long-living projects....)

The test case here is motivated by what I have in a real project, where the Jars are produced by a _Transform_. The adjustment is needed because the _report_ task has an `artifactView` as input and not directly the `compileClasspath` Configuration. If you look closely at the setup, I reproduce the `artifactView` the task normally has as input, but add another attribute to it. You can attribute this to some of the _shortcomings_ of the rather complex _Transformations+ArtifactView_ APIs. Ideally, there would be some less invasive way to express this (somehow just adding the attribute - needed to trigger the transform - on top). But I don't know of another way to do this right now.

I will have a loot at adding a second case, where another local project produces the two Jars.

jjohannes: b6bc3d484e525209c833599a830eda1450269e3a

src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
```diff
@@ -73,7 +73,7 @@ internal class JarExploder(
       analyzedClasses = analyzedClasses,
       ktFiles = ktFiles,
       androidLintRegistry = findAndroidLinter(artifact)
-    ).also { inMemoryCache.explodedJars(zip.name, it) }
+    ).also { inMemoryCache.explodedJars(artifact.file.absolutePath, it) }
```
autonomousapps, line 76: Thanks. I think you were right, this never worked. A bit of an error when I changed models for the v1 release.

src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+
+import com.autonomousapps.jvm.projects.JarTransformingProject
+import com.autonomousapps.jvm.projects.MultipleJarsProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class ComponentWithMultipleArtifactsSpec extends AbstractJvmSpec {
+
+  def "one component can have multiple Jars (#gradleVersion)"() {
```
autonomousapps, line 11: I know this is an obvious question, but this test would fail without the changes you made? It's not clear to me which change would cause the test to switch from failing to passing.

src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
mvegter: @jjohannes could you cherry pick this to a different pull request

src/functionalTest/groovy/com/autonomousapps/jvm/ComponentWithMultipleArtifactsSpec.groovy
jjohannes: Yes it fails (both new tests fail).

The important changes are these two `compareTo` implementations:

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/688/files#diff-a7da78989269bbf240e2f6adc5f5e1a6ac7c18bcf0ba4b3bc1b2fddad041defeR197

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/688/files#diff-8f376c467c8ba4a5dcbe0c0dadc0d561d71916cd68a5afb0781f1d63c0421776R17

There are two places where a list is converted into an OrderedSet. Without this changes, the OrderedSets only have one entry per _coordinates_. If there are more than one file for one _coordinates_, entries will be dropped with the old _compareTo_ implementations.

Here are the places where the sets are created:

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/03a4ad7e6655beaecb75b1dabd85c4b87af08d8d/src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt#L88

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/33c25ca68b248ae13fab454f05c580115f7dfbb8/src/main/kotlin/com/autonomousapps/internal/JarExploder.kt#L31


**Note:** The change in `Dependency.kt` does not have any effect, because the code seems to be unused. But I would do it, because otherwise it is inconsistent. Another option would be to remove the `Dependency.file(s)` field completely if it is not needed:
 https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/688/files#diff-9e4cd8d8f8a7cf963f2c64cc6f1e5c071bb9f6c2378d2e59bf874f9bf9b95ba8R13

src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
jjohannes: Sure. I removed the change from this PR and moved it here: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/699

autonomousapps:
Thank you for the contribution!

PR #691, opened 2022-06-04
Detect custom application defined in a different module.

autonomousapps:


testkit/src/main/kotlin/com/autonomousapps/kit/GradleProject.kt
```diff
@@ -147,6 +147,7 @@ class GradleProject(
       val builder = androidSubprojectMap[name] ?: AndroidSubproject.Builder()
       builder.apply {
         this.name = name
+        //this.manifest = AndroidManifest.defaultLib(name) // TODO
```
autonomousapps, line 150: TODO in a separate commit or PR.

PR #695, opened 2022-06-04
Less aggressive

autonomousapps:
See also https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/694.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/BytecodeParsers.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/BytecodeParsers.kt
@@ -1,5 +1,6 @@
 package com.autonomousapps.internal
 
+import com.autonomousapps.internal.ClassNames.canonicalize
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/autonomousapps/internal/BytecodeParsers.kt
```diff
@@ -71,14 +71,14 @@ private class BytecodeReader(
 
     return ExplodedClass(
       source = classAnalyzer.source,
-      className = classAnalyzer.className,
+      className = canonicalize(classAnalyzer.className),
       usedClasses = constantPool.asSequence().plus(classAnalyzer.classes)
         // Filter out `java` packages, but not `javax`
         .filterNot { it.startsWith("java/") }
```
autonomousapps, line 77: should this be `java.`?

PR #696, opened 2022-06-06
Remove unused jars and classes map of InMemoryCache

mvegter:


autonomousapps (commented):
Hm. Maybe the solution here isn't to remove the "unused" feature, but actually use it. Looks like the usage probably got dropped with the move to the new model.

PR #697, opened 2022-06-06
New DominatorViewTask to print dominator tree.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/256.


src/functionalTest/groovy/com/autonomousapps/android/DominanceTreeSpec.groovy
```diff
@@ -0,0 +1,21 @@
+package com.autonomousapps.android
+
+import com.autonomousapps.android.projects.DominanceTreeProject
+import org.gradle.util.GradleVersion
+
+import static com.autonomousapps.utils.Runner.build
+
+// TODO probably convert this to a JVM spec
```
autonomousapps, line 8: TODO.

autonomousapps:
Build failed https://gradle.com/s/p6obyrj33ctvo

autonomousapps:
Build failed https://gradle.com/s/eobg3hh3pb73c

autonomousapps:
Build failed https://gradle.com/s/6j5bhwvojm54c

autonomousapps:
Build failed https://gradle.com/s/zo3k7co2wf5rq

src/main/kotlin/com/autonomousapps/internal/graph/DominanceTree.kt
```diff
@@ -0,0 +1,320 @@
+/*
+ * Original license:
+ *
+ * Renjin : JVM-based interpreter for the R language for the statistical analysis
+ * Copyright © 2010-2019 BeDataDriven Groep B.V. and contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
```
vlsi, line 8: I am afraid if you include GPL2 pieces, you have to switch the license to GPL2 which would be awful :(

autonomousapps: IANAL. How much modification before it's considered a new thing?

I am actually considering rewriting it with a different algorithm, which would make it a moot point.

vlsi: I would avoid touching GPL in libraries/plugins at any costs possible.
In most cases, even if you just combine GPL with something else (without modifications), you **have to relicense** the combined thing under GPL.
Of course, there are escape hatches like "if you talk to GPL program using something like regular shell scripts without tight integration with the program interfaces in GPL code, then you can avoid relicensing under GPL". However, it is incredibly moot, and I would strongly suggest you keep out of GPL, LGPL, AGPL, etc unless you really really understand what you are doing.

Here's a recent thread on AGPL: https://twitter.com/marcan42/status/1512267944436600836

Apache Software Foundation has a list of "resolved" cases where licenses are approved or rejected for including into the ASF software: https://www.apache.org/legal/resolved.html#category-x
Note: it does not mean those licenses are incompatible with "Apache license". The list of resolved cases is more like "those are the licenses that violate the spirit of the Apache Software Foundation that was established to produce open-source software that can be used and modified by the end-users".
As you license `d-a-a-g-p` under `Apache 2.0`, it probably makes sense to use https://www.apache.org/legal/resolved.html#category-x as a cheat-sheet for IANAL :)

Apache LEGAL JIRA is yet another nice reference to get the understanding behind the reasons certain licenses are forbidden in the ASF: https://issues.apache.org/jira/browse/LEGAL-192

vlsi: Just in case, `GPL with ClassPath-Exception` is OK as it basically means "if you just use the interfaces without modifying GPL code, then you are free to use the software and you don't need to relicense".

A common case for `GPL with ClassPath-Exception` is OpenJDK. For instance, OpenJDK has JMH library for micro-benchmarking: https://openjdk.java.net/projects/code-tools/jmh/
It is licensed under `GPL with ClassPath-Exception`, so it is fine to use JMH, and you don't have to relicense thanks to `ClassPath-Exception`.

autonomousapps: I'm not entirely sure what the "classpath-exception" means, but based on the name I assume it means it would be ok if the algorithm came from a binary dependency (library) that was simply "on the classpath." That is not an option here, since I had to make significant changes to make the algorithm generic. So my question there is, at what point are the changes so significant that I actually have a new thing? Never? I also find it confusing because it was just a Java implementation of a decades-old, very well known algorithm. I think there's an argument that any implementation in Java would basically look like that. And now that I've _seen_ that implementation, how would I implement it "from scratch" in an Apache license-compatible way? In a sense, my mind is now "polluted" with the knowledge of this specific implementation. But again, it's just a straightforward implementation of an algorithm that has long been well-known to the field of computer science.

Alllll that said, I still think I may go ahead and re-implement it based on a different algorithm entirely, rendering this discussion moot. But I still find it moderately annoying.

vlsi: >In a sense, my mind is now "polluted" with the knowledge of this specific implementation

That is why it makes sense to avoid reading GPL code in the first place.
In this case I've stopped reading the file as soon I realized it was GPL

 You could search for Apache-2.0 or similar implementation or you could try to forget or you could ask someone for a clean-room implementation.

vlsi: >But I still find it moderately annoying.

Licencing is hard, and embedding others' code makes it even harder :-/
I can relate your pain here.

autonomousapps: This file has been deleted and has been replaced with something entirely different.

src/main/kotlin/com/autonomousapps/tasks/DominatorViewTask.kt
```diff
@@ -0,0 +1,159 @@
...
+    // TODO temporary until task split into output-creating and console-printing
+    outputs.upToDateWhen { false }
+  }
+
+  @get:Input
+  abstract val projectPath: Property<String>
+
+  @get:PathSensitive(PathSensitivity.NONE)
+  @get:InputFile
+  abstract val physicalArtifacts: RegularFileProperty
```
vlsi: Should it rather be `@Input Configuration`?

autonomousapps:
@vlsi I have a final draft now with improved console output. The core code is also generic with the idea that I might be able to spin it off in the future, but that's speculative only.

What remains is some way of asking it to print different configurations, but the PR is already so large I'd rather tackle that separately. That is also non-trivial when you consider the Android use-case, which differs in significant ways from the JVM.

Could you give it another look and let me know what you think?

src/main/kotlin/com/autonomousapps/tasks/DominatorViewTask.kt
autonomousapps: Using this input lets me re-use existing plugin infrastructure. To achieve your desired goal of printing arbitrary configurations, this would indeed have to change. I need to take some time to consider how best to implement it for the various use-cases that might be worth supporting, including Android as well as JVM.

It's very frustrating that Gradle tasks only support flags when you invoke the specific task that takes the flag. It would be great if a flag (like `--configuration runtimeClasspath`) could be consumed my multiple tasks in the task graph. Because that's not the case, I need to explode the size of this relatively simple task and have it duplicate functionality from other tasks. I don't want to do that without careful consideration. I don't want to create headaches for my future self.

autonomousapps:
Build failed https://gradle.com/s/m6ozgvl4uo5po

PR #698, opened 2022-06-06
Add support for type-safe project accessors

### LINKED ISSUE #480: Support type-safe project dependencies in config

Can't use [type-safe project accessors](https://docs.gradle.org/7.0/userguide/declaring_dependencies.html#sec:type-safe-project-accessors) in config:

```kotlin
exclude("my-project:subproject")
```
We migrated to type-safe project accessors but have to use string paths in configs. 
It'd great to use the same notation.

To pass type-safe project accessors to the config:

```kotlin
exclude(projects.myProject.subproject)
```

**Additional context**

SimonMarquis:
Closes #480

### DIFF
--- a/src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
@@ -6,6 +6,7 @@ import org.gradle.api.Action
 import org.gradle.api.GradleException
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Named
+import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
 import org.gradle.api.provider.SetProperty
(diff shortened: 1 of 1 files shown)


autonomousapps (commented):
Thank you for the contribution! I'd love to see a test validating this works as desired. Happy to work with you on that if writing Gradle tests is unfamiliar.

SimonMarquis:
Sure, I was going to take a shot at this, but first I tried to run the test suite locally, and I encountered a lot of issues...

Certainly related to Windows AV (or Gradle itself) as many files were locked by a process (Gradle daemon?) and tests were failing while trying to delete jUnit tmp directories.
I forced the `--max-workers=4` but then [it took ages to complete](https://scans.gradle.com/s/pcvgxzock2xuy).

SimonMarquis:
I suppose you'd rather prefer functional tests (`GradleBuildSrcConventionMultiConfigProject`?) over unit tests on the `Issue` class?

autonomousapps:
Hm, I could be open to a unit test of this class, if that test showed that the new method were equivalent to the others. I do tend to have more confidence in the functional test suite, though. Unit tests of Gradle plugins tend to involve a lot of mocking and/or stubbing, which limits their efficacy in my view.

I'm also open to any improvements to the functional test suite to make it run better on Windows. I admit I've never tried to run these tests in a Windows environment, only Linux and macOS.

One thing you can do is run a subset of the tests. So let's say you add a new functional test named `TypeSafeProjectAccessorsSpec`, you can then run that test with `./gradlew :functionalTest --tests TypeSafeProjectAccessorsSpec -DfuncTest.quick`. The system property will ensure it only runs one iteration of your new spec, rather than against the whole set of supported Gradle versions. To the extent that your issue with running the tests is about the parallelization, that should help a lot.

autonomousapps:
@SimonMarquis if you rebase this, I'd be open to writing a quick test case. Let me know.

SimonMarquis:
Hi, really sorry for not doing much here. I didn't had enough time to focus on this project lately.

PR #732, opened 2022-07-04
Infer if an android project could be a plain java/kotlin project.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/16.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 Dependency Analysis Plugin Changelog
 
+# Version 1.10.0
+* [New] `buildHealth` and `projectHealth` now include a new category of advice, "module structure advice." This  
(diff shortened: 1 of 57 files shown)


sboishtyan:
Is there a way to disable this check?

sboishtyan:
I've check the impact of this task. And it's a 5d per week for PR builds.
I don't know why it could take so much time. But I wanna to discover and it would be great to be able to disable this check

autonomousapps:
> I've checked the impact of this task. And it's 5d per week for our PR builds. We have a large monorepo but it's a top 10 task in our CI. I don't know why it could take so much time. But I wanna discover and it would be great to be able to disable this check

Please file a new issue and I'll take a look.

PR #744, opened 2022-07-11
Explain module structure advice.

autonomousapps:


src/main/kotlin/com/autonomousapps/internal/reason/ModuleAdviceExplainer.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.autonomousapps.tasks.ReasonTask
+import org.gradle.kotlin.dsl.support.appendReproducibleNewLine
+
+internal class ModuleAdviceExplainer(
+  private val project: ProjectCoordinates,
+  private val unfilteredAndroidScore: AndroidScore?,
+  private val finalAndroidScore: AndroidScore?,
+) : ReasonTask.Explainer {
+
+  override fun computeReason() = buildString {
```
handstandsam, line 15: Wow, today I learned about the method `buildString` 😱 .  I just used `StringBuilder().apply{}` before.

PR #764, opened 2022-08-11
Fixes #763

### LINKED ISSUE #763: Annotations that reference multiple classes are not correctly exposed as part of the public ABI
**Plugin version**
1.12.1-SNAPSHOT

**Gradle version**
7.4.2

**Describe the bug**
If an annotation references multiple types the plugin does not consider them as part of the abi.

For example given the following annotations:
```java
@Target(ElementType.TYPE)
public @interface MyJavaSingleAnnotation {
    Class<?> clazz();
}

@Target(ElementType.TYPE)
public @interface MyJavaMultipleAnnotation {
    Class<?>[] clazz() default {};
}

class InnerClassJ {}

@MyJavaSingleAnnotation(clazz = InnerClassJ.class)
public class MyCoolJavaSingleClassInA {}

@MyJavaMultipleAnnotation(clazz = {
        InnerClassJ.class, MyCoolJavaSingleClassInA.class
})
public class MyCoolJavaClassInA {}
```
And kotlin code
```kotlin
@Target(AnnotationTarget.CLASS)
annotation class MyKotlinAnnotationSingle(
    val clazz: KClass<*>
)

@Target(AnnotationTarget.CLASS)
annotation class MyKotlinAnnotationMultiple(
    val clazz: Array<KClass<*>>,
)

class InnerClass {}

@MyKotlinAnnotationSingle(InnerClass::class)
class MyCoolSingleClassInA {}

@MyKotlinAnnotationMultiple([InnerClass::class, MyCoolClassInA::class])
class MyCoolClassInA {}
```

The public-abi file generated looks as follows:
```
cat module-a/build/reports/dependency-analysis/main/abi-dump.txt 
@Lkotlin/Metadata;
public final class InnerClass {
        public fun <init> ()V
}

@Lcom/test/MyKotlinAnnotationMultiple;
@Lkotlin/Metadata;
public final class MyCoolClassInA {
        public fun <init> ()V
}

@Lcom/test/MyKotlinAnnotationSingle;
@LInnerClass;
@Lkotlin/Metadata;
public final class MyCoolSingleClassInA {
        public fun <init> ()V
}

@Lcom/test/MyJavaMultipleAnnotation;
public class com/test/MyCoolJavaClassInA {
        public fun <init> ()V
}

@Lcom/test/MyJavaSingleAnnotation;
@Lcom/test/InnerClassJ;
public class com/test/MyCoolJavaSingleClassInA {
        public fun <init> ()V
}
```

Notice that the multiple annotations (in both java and kotlin) do not correctly reference the two classes that are used within them while the single one does.

**To Reproduce**
I've attached a sample project that reproduces the issue.

[sample-dep-analysis-bug.zip](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/files/9308852/sample-dep-analysis-bug.zip)

**Expected behavior**
We would expect the following abi file to be outputted. Notice that the Multiple* annotations now correctly reference all the classes that are utilized.
```
@Lkotlin/Metadata;
public final class InnerClass {
	public fun <init> ()V
}

@Lcom/test/MyKotlinAnnotationMultiple;
@LInnerClass;
@LMyCoolClassInA;
@Lkotlin/Metadata;
public final class MyCoolClassInA {
	public fun <init> ()V
}

@Lcom/test/MyKotlinAnnotationSingle;
@LInnerClass;
@Lkotlin/Metadata;
public final class MyCoolSingleClassInA {
	public fun <init> ()V
}

@Lcom/test/MyJavaMultipleAnnotation;
@Lcom/test/InnerClassJ;
@Lcom/test/MyCoolJavaSingleClassInA;
public class com/test/MyCoolJavaClassInA {
	public fun <init> ()V
}

@Lcom/test/MyJavaSingleAnnotation;
@Lcom/test/InnerClassJ;
public class com/test/MyCoolJavaSingleClassInA {
	public fun <init> ()V
}
```

**Additional context**
It seems like the root cause of the bug is
```kotlin
anno.values.orEmpty().filterIsInstance<Type>().forEach { type ->
      types.add(type.descriptor)
    }
```
annotation values can also be of type `ArrayList<*>` which might then contain references to the types. PR incoming.

FredLoh:
If an annotation references multiple types they are not correctly exposed as part of a module's public ABI. This is because we filter for an annotation values to only include a single `Type`. However annotations may have a value of type `ArrayList<*>` which can contain references to other classes.

This patch updates the `List<AnnotationNode>?.annotationTypes()` method to correctly handle the above scenario and adds some unit tests for it.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
@@ -24,8 +24,14 @@ internal fun List<AnnotationNode>?.annotationTypes(): Set<String> {
 
   forEach { anno ->
     types.add(anno.desc)
-    anno.values.orEmpty().filterIsInstance<Type>().forEach { type ->
-      types.add(type.descriptor)
+
+    anno.values.orEmpty().filter {
+      it is Type || it is ArrayList<*>
+    }.forEach { value ->
+      when (value) {
+        is Type -> types.add(value.descriptor)
+        is ArrayList<*> -> value.filterIsInstance<Type>().forEach { type -> types.add(type.descriptor) }
+      }
     }
   }
 
--- a/src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
@@ -26,11 +26,11 @@ internal fun List<AnnotationNode>?.annotationTypes(): Set<String> {
     types.add(anno.desc)
 
     anno.values.orEmpty().filter {
-      it is Type || it is ArrayList<*>
+      it is Type || it is List<*>
     }.forEach { value ->
       when (value) {
         is Type -> types.add(value.descriptor)
-        is ArrayList<*> -> value.filterIsInstance<Type>().forEach { type -> types.add(type.descriptor) }
+        is List<*> -> value.filterIsInstance<Type>().forEach { type -> types.add(type.descriptor) }
       }
     }
   }
--- a/src/test/kotlin/com/autonomousapps/internal/utils/RegexKtTest.kt
+++ b/src/test/kotlin/com/autonomousapps/internal/utils/RegexKtTest.kt
@@ -3,6 +3,7 @@ package com.autonomousapps.internal.utils
 import com.autonomousapps.internal.asm.Type
 import com.autonomousapps.internal.asm.tree.AnnotationNode
 import com.google.common.truth.Truth
+import com.google.common.truth.Truth.assertThat
 import org.junit.jupiter.api.Test
 
 
@@ -13,15 +14,15 @@ class AnnotationNodeExtensionTest {
     annotationNode.values = mutableListOf()
     annotationNode.values.add(
       arrayListOf(
-        "a string element that should be filtered",
+        "a string element that should be filtered out",
         Type.getType("Lcom/test/AReferencedClass;"),
         Type.getType("Lcom/test/AnotherOne;")
       )
     )
 
     val annotationTypes = listOf(annotationNode).annotationTypes()
-    Truth.assertThat(annotationTypes).hasSize(3)
-    Truth.assertThat(annotationTypes).containsExactly(
+    assertThat(annotationTypes).hasSize(3)
+    assertThat(annotationTypes).containsExactly(
       "Lcom/test/AnnotationWithMultipleReferences;", "Lcom/test/AReferencedClass;", "Lcom/test/AnotherOne;"
     )
   }
@@ -34,8 +35,8 @@ class AnnotationNodeExtensionTest {
     )
 
     val annotationTypes = listOf(annotationNode).annotationTypes()
-    Truth.assertThat(annotationTypes).hasSize(2)
-    Truth.assertThat(annotationTypes).containsExactly(
+    assertThat(annotationTypes).hasSize(2)
+    assertThat(annotationTypes).containsExactly(
       "Lcom/test/AnnotationWithSingleReference;", "Lcom/test/AReferencedClass;"
     )
   }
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
```diff
@@ -24,8 +24,14 @@ internal fun List<AnnotationNode>?.annotationTypes(): Set<String> {
...
     types.add(anno.desc)
-    anno.values.orEmpty().filterIsInstance<Type>().forEach { type ->
-      types.add(type.descriptor)
+
+    anno.values.orEmpty().filter {
+      it is Type || it is ArrayList<*>
+    }.forEach { value ->
+      when (value) {
+        is Type -> types.add(value.descriptor)
+        is ArrayList<*> -> value.filterIsInstance<Type>().forEach { type -> types.add(type.descriptor) }
```
autonomousapps, line 33: Just checking, could this be a `List<List<...>>`?

Also, must it be an `ArrayList`? Why not just `List`?

src/test/kotlin/com/autonomousapps/internal/utils/RegexKtTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+
+class AnnotationNodeExtensionTest {
+
+  @Test fun `should correctly parse an annotation node that references a list of Types`() {
+    val annotationNode = AnnotationNode("Lcom/test/AnnotationWithMultipleReferences;")
+    annotationNode.values = mutableListOf()
+    annotationNode.values.add(
+      arrayListOf(
+        "a string element that should be filtered",
```
autonomousapps, line 16: could you say "filtered out" instead? `filter` usually implies _filtering in_.

src/test/kotlin/com/autonomousapps/internal/utils/RegexKtTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        Type.getType("Lcom/test/AReferencedClass;"),
+        Type.getType("Lcom/test/AnotherOne;")
+      )
+    )
+
+    val annotationTypes = listOf(annotationNode).annotationTypes()
+    Truth.assertThat(annotationTypes).hasSize(3)
```
autonomousapps, line 23: nit: static import `assertThat`

autonomousapps (commented):
Thanks so much for this PR! Left a couple of questions, but it looks really good and I look forward to merging it.

src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
FredLoh: The type comes through as `ArrayList`. A naive check for `List` fails but I can cast and check if it succeeds as an option?

src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
```diff
@@ -24,8 +24,14 @@ internal fun List<AnnotationNode>?.annotationTypes(): Set<String> {
 
   forEach { anno ->
     types.add(anno.desc)
-    anno.values.orEmpty().filterIsInstance<Type>().forEach { type ->
-      types.add(type.descriptor)
+
+    anno.values.orEmpty().filter {
+      it is Type || it is ArrayList<*>
```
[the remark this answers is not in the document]
FredLoh: Kept since it matches previous code style

FredLoh (commented):
Will wait for a response then upload a commit with the fixes.

FredLoh:
Sorry for the delay on the response, was on PTO 😄

autonomousapps:
> Sorry for the delay on the response, was on PTO 😄

No need to apologize! This is OSS, timing is always best-effort 🙃

src/main/kotlin/com/autonomousapps/internal/utils/regex.kt
autonomousapps: I'm just wondering if there's some guarantee that it's always an `ArrayList`. For obvious reasons I prefer to use a more general type whenever possible.

FredLoh: Ah I see. Yeah you're right, List work's here and matches what ASM expects: https://asm.ow2.io/javadoc/org/objectweb/asm/tree/AnnotationNode.html#values

FredLoh:
Should we run the workflow's on these? I can fix any failures that result

autonomousapps:
> Should we run the workflow's on these? I can fix any failures that result

I don't know why it keeps asking me to do that 🤦

FredLoh:
> > Should we run the workflow's on these? I can fix any failures that result
> 
> I don't know why it keeps asking me to do that 🤦

I guess they're afraid I'll start opening hundreds of PR's to mine crypto 🗑️

autonomousapps (approved):

autonomousapps:
Thanks for your contribution!

PR #772, opened 2022-08-21
Add support for path: in gradle script.

FredLoh:
This is the default way a project is auto added if you utilize Android Studio's "Add dependency on module" auto fix so it seems to be very prevalent in build scripts.

CleanShot 2022-08-21 at 10 43 52@2x

Results in:
CleanShot 2022-08-21 at 10 44 04@2x


FredLoh:
Just realized that https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/752/ does exactly this!

PR #802, opened 2022-11-05
Update max agp version

liutikas:
Tested it locally and it works just fine on this version


autonomousapps (approved):

autonomousapps (changes requested):
Test matrix should be updated before merging these changes.

Issue #798, opened 2022-10-26
`Unsupported class file major version 63`


boris-petrov:
**Plugin version**
1.13.1

**Gradle version**
7.5.1

**Describe the bug**
Running `gradle buildHealth` on the `build.gradle` file below leads to:

```
Execution failed for task ':explodeJarMain'.
> A failure occurred while executing com.autonomousapps.tasks.ExplodeJarTask$ExplodeJarWorkAction
   > Unsupported class file major version 63
```

**To Reproduce**
`build.gradle`:
```gradle
plugins {
	id 'antlr'
	id 'com.autonomousapps.dependency-analysis' version '1.13.1'
}

repositories {
	mavenCentral()
}

dependencies {
	implementation 'org.apache.lucene:lucene-analysis-common:9.4.1'
}
```

**Expected behavior**
That to work.

**Additional context**
I've only added the `antlr` plugin so that `buildHealth` runs `explodeJar*` tasks - I couldn't make it do that without it.
Downgrading `lucene` to `9.3.0` fixes the issue.

JacksonBailey:
I have this problem too but don't use lucene. Judging by the error we are both getting (`Unsupported class file major version 63`) I think the problem is class files compiled for Java 19. In the [change log for lucene 9.4.0][changelog] I found this mentioned, [MR-JAR rewrite of MMapDirectory with JDK-19 preview Panama APIs (>= JDK-19-ea+23)][pr]. To me this seems to indicate 9.4.0 and up use Java 19.

In my project I am using Java 19 (via toolchain) and get the same error you do. When I change it to 18 (or even 17) I don't get the error. So I think the problem is that this doesn't work with Java 19.

[changelog]: https://lucene.apache.org/core/9_4_1/changes/Changes.html#v9.4.0.new_features
[pr]: https://github.com/apache/lucene/pull/912

uschindler:
Lucene only has some class file in META-INF/versions/19.
Lucene 9.4 is still on Java 11, it is only some class files for newer Java versions.

I think the plugin should either:
- Do not look into class files it does not undertsand (maybe print a warning, optionally)
- Exclude Java versions from Multi-Release-JARs (which Lucene uses).

autonomousapps:
Thanks for the issue and the discussion. I'm open to contributions if someone wants to try to fix this. I can also take a look but I can't make any promises on timing right now.

JacksonBailey:
[...]

PR #806, opened 2022-11-15
Use asm unshadowed and upgrade to 9.4

JacksonBailey:
Big heads up:

1. I did not run many tests and I even saw a failure. Not sure if I had something setup wrong or what.
2. The test I added here *passes* even when I had 9.2 in use so I know it's not working. That said, when pulling from maven local the snapshot version I no longer see the error on my project so I know this fixes it (but could cause other problems).

I will look into running the full suite (and even the quick suite) soon, but this is just something quick and dirty so show the proof of concept.

For what it is worth, [here is my project](https://github.com/JacksonBailey/wheel/blob/3337502695e10c85efe31551340ab9b9584aea1a/build.gradle.kts#L3-L12) where the fix is shown in use. It's not really anything to look at but if you'd like to clone it and try you can. The `buildHealth` task works with the snapshot build here but not with `1.14.1`.

See #798

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
@@ -6,6 +6,7 @@ import static com.autonomousapps.utils.Runner.build
 import static com.google.common.truth.Truth.assertThat
 
 final class JavaToolchainSpec extends AbstractJvmSpec {
+
   def "does not fail with Java #javaToolchainVersion toolchain (#gradleVersion)"() {
     given: "a Gradle project using Java toolchain version #javaToolchainVersion that has been built"
     def project = new JavaToolchainProject(javaToolchainVersion)
@@ -25,11 +26,15 @@ final class JavaToolchainSpec extends AbstractJvmSpec {
 
     where:
     // Comment out all but the last for speed
+    // See https://docs.gradle.org/8.5/userguide/compatibility.html#java
     gradleVersion | javaToolchainVersion
-    GRADLE_7_5    | 18
-    GRADLE_7_5    | 19
-    GRADLE_7_6    | 18
+//    GRADLE_7_5    | 18
+//    GRADLE_7_5    | 19
+//    GRADLE_7_6    | 18
     GRADLE_7_6    | 19
+    // TODO(tsr): some errors running these two scenarios
+//    GRADLE_8_3    | 20
+//    GRADLE_8_5    | 21
 
     classFileMajorVersion = javaToolchainVersion + 44 // 19 + 44 = 63, is this safe?
   }
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
@@ -1,0 +1,36 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.JavaToolchainProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class JavaToolchainSpec extends AbstractJvmSpec {
+  def "does not fail with Java #javaToolchainVersion toolchain (#gradleVersion)"() {
+    given: "a Gradle project using Java toolchain version #javaToolchainVersion that has been built"
+    def project = new JavaToolchainProject(javaToolchainVersion)
+    gradleProject = project.gradleProject
+    build(gradleVersion, gradleProject.rootDir, 'build')
+
+    when: 'running buildHealth'
+    def result = build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then: 'abiAnalysisMain and explodeByteCodeSourceMain tasks were ran'
+    // These two are the ones that were actually failing, sanity check that java library plugin is applied
+    assert result.tasks.any { (it.getPath() == ':proj:abiAnalysisMain') }
+    assert result.tasks.any { (it.getPath() == ':proj:explodeByteCodeSourceMain') }
+
+    and: 'the report should be empty'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    // Comment out all but the last for speed
+    gradleVersion | javaToolchainVersion
+    GRADLE_7_5    | 18
+    GRADLE_7_5    | 19
+    GRADLE_7_6    | 18
+    GRADLE_7_6    | 19
+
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -136,7 +136,9 @@ dependencies {
     because("Graphs")
   }
   implementation(libs.relocated.antlr)
-  implementation(libs.relocated.asm)
+  //implementation(libs.relocated.asm)
```
JacksonBailey, line 139: Left comments in because it's WIP -- You may have reasons I don't understand for using the shadow jar so didn't want to go too far down this route.

src/functionalTest/groovy/com/autonomousapps/jvm/Java19Spec.groovy
```diff
@@ -0,0 +1,25 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.Java19Project
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class Java19Spec extends AbstractJvmSpec {
```
JacksonBailey, line 8: Gives false positive when using 9.2 (the bad version) of `asm`.

build.gradle.kts
autonomousapps: Yup, as I said on the issue, I do have reasons to use the shadow jar.

JacksonBailey: Sounds good. I may need some help with the final step since it seems like it's two parts (publish new shaded jar ***then*** test new version). In the mean time I'll revert the changes I made here. I can use the bad one for now since I'm trying to get a failing test case first.

poc/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
...
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+    dependencies {
+        classpath("com.autonomousapps:dependency-analysis-gradle-plugin:1.14.1-SNAPSHOT")
+    }
+}
+
+apply(plugin = "com.autonomousapps.dependency-analysis")
```
JacksonBailey, line 11: Easier to access POC if you're curious. I'll make sure this is not in my branch's history before merge that way it doesn't clutter anything.

src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
```diff
@@ -0,0 +1,39 @@
...
+  def "does not fail with Java #javaToolchainVersion toolchain (#gradleVersion)"() {
+    given: "a Gradle project using Java toolchain version #javaToolchainVersion that's been built"
+    def project = new JavaToolchainProject(javaToolchainVersion)
+    gradleProject = project.gradleProject
+    build(gradleVersion, gradleProject.rootDir, 'tasks', '--all')
+    build(gradleVersion, gradleProject.rootDir, 'build')
+
+    when: 'running buildHealth'
+    // TODO This passes because it is not running abiAnalysisMain nor explodeByteCodeSourceMain which are the problem
+    // I added tasks --all above to see and indeed they're not part of buildHealth like in the /poc folder I made
```
JacksonBailey, line 20: @autonomousapps At the moment this is where I need your wisdom the most. How can I make the tasks `abiAnalysisMain` and `explodeByteCodeSourceMain` be part of the project. Those are the two that actually break right now. As you can see in the poc I added I didn't really "do" anything special to get them into the project so I'm confused why they aren't being added here.

JacksonBailey:
## Steps

1. [ ] Add a test that fails with Java 19 toolchain and 9.2 shaded asm.
2. [ ] (Maybe) Catch the exception and try to give a better error message. Something like "Got a class version error, asm may need to be updated from version X.Y to work with newer versions of Java, [link to asm changelog]"
4. [ ] Actually use new version of asm 9.4 and publish shaded jar 😄 (will need assistance publishing when we get here)
5. [ ] Use new version of shaded jar and verify the test from step 1 passes
6. [ ] Publish new plugin version

autonomousapps:
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/813

src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
JacksonBailey: Never mind, I figured it out. I needed to apply java lib plugin instead of java plugin.

src/functionalTest/groovy/com/autonomousapps/jvm/JavaToolchainSpec.groovy
```diff
@@ -0,0 +1,36 @@
...
+    and: 'the report should be empty'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    // Comment out all but the last for speed
+    gradleVersion | javaToolchainVersion
+    GRADLE_7_5    | 18
+    GRADLE_7_5    | 19
+    GRADLE_7_6    | 18
+    GRADLE_7_6    | 19
```
JacksonBailey, line 32: Questions

1. Do you want this to use the `gradleVersions()` method like the other tests?
2. Do you want me to query Adoptium API to try and get latest version of Java so it's easier to catch? It's gross but it might be useful. Not sure your opinion.
3. If I do multiple versions of Java toolchain and gradle versions, do you want me to make it so it's all toolchains on one version of gradle then all versions of gradle with one version of toolchain? That way it's not n^2 and just 2n tests haha.

JacksonBailey:
I got a test failure finally so went ahead and rebased. Sure enough, it works now! Were you interested in the error message thing or the testing against more versions of gradle/java?

JacksonBailey:
@autonomousapps no worries, take your time. Enjoy your Thanksgiving (if you're from the states). Glancing back over this now the change you made should fix the problem found in the issue. At the moment this is just a unit test to make sure it doesn't come back. Essentially your other PR bumping to 9.4 should fix everything, this is just the extra assurance.

autonomousapps:
Wow you were not wrong with the "Comment out all but the last for speed" comment 😅 

I've updated the tests since there have been some breaking changes in the test fixtures support API.

I also updated the spec matrix to reference more versions of Gradle. I'm currently running it only against Gradle 7.6.

I tried to run it against Gradle 8.3 and 8.5, but go some errors. I don't have the bandwidth now to address them, but if you feel like picking this up again I will try to move faster.

I will merge this when green (running against 7.6).

JacksonBailey:
> I don't have the bandwidth now to address [Gradle 8.3 and 8.5 errors], but if you feel like picking this up again I will try to move faster.

Don't beat yourself up at all, we appreciate your work. ❤️ I might take a peek, but no promises. At the moment I am more interested in getting Java 20 and Java 21 to work (I assume they don't, just responding because I saw this in my email) than getting more Gradle versions to work.

PR #807, opened 2022-11-15
Duplicated declaration does not lead to wrong analysis result

jjohannes:
If there are two declarations to the same component with different scopes, e.g.

```
dependencies {
  runtimeOnly("org.apache.commons:commons-collections4:4.4")
  implementation("org.apache.commons:commons-collections4:4.4")
}
```

the analysis discarded one of them by using 'find' instead of 'filter' in StandardTransform.kt. This can lead to a wrong recommendation. In the example provided in the test, the plugin recommended to remove the 'implementation' dependency.

A related feature could be that the plugin also detects the duplication, and recommends to remove the unneeded declaration - in this case `runtimeOnly("...")`. Doing this, would be an additional analysis imo, as the plugin would need to check how the different configurations relate. And maybe this is also not wanted in all cases. The setup is not "wrong" as the classpaths don't change by removing the duplicates. It may be out of scope of this plugin altogether.

This fix is only about not reporting something wrong.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
+++ b/src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
@@ -103,7 +103,7 @@ internal class StandardTransform(
     val usageIter = usages.iterator()
     while (usageIter.hasNext()) {
       val usage = usageIter.next()
-      val declarationsForVariant = declarations.filter { it.variant(supportedSourceSets) == usage.variant }.toSet()
+      val declarationsForVariant = declarations.filterToSet { it.variant(supportedSourceSets) == usage.variant }
 
       // We have a declaration on the same variant as the usage. Remove or change it, if necessary.
       if (declarationsForVariant.isNotEmpty()) {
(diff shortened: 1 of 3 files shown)


autonomousapps:
> A related feature could be that the plugin also detects the duplication, and recommends to remove the unneeded declaration

The plugin actually used to do this, pre-1.0. I forget all the context, but I stopped doing it after some complaints surfaced about false positives.

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
```diff
@@ -103,38 +103,40 @@ internal class StandardTransform(
     val usageIter = usages.iterator()
     while (usageIter.hasNext()) {
       val usage = usageIter.next()
-      val decl = declarations.find { it.variant(supportedSourceSets) == usage.variant }
+      val declarationsForVariant = declarations.filter { it.variant(supportedSourceSets) == usage.variant }.toSet()
```
autonomousapps, line 106: there's a `filterToSet` extension function that I'd prefer to use to reduce intermediate object creation.

autonomousapps (changes requested):
Thanks for the contribution! I have one minor change request.

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
jjohannes: Done.

autonomousapps (approved):

PR #808, opened 2022-11-15
Fix issues with included builds

jjohannes:
This PR adds test coverage and fixes two issues with included builds:

1. The requested version is irrelevant if resolved to an included build (sub)project. The analysis also needs to ignore it. Otherwise, it won't be able to connect the computed dependency information correctly to the declared dependencies.
2. If a build is included, inside that build the projects often still depend on each other using `project("...")` dependencies. This was not working.

See also commit messages and comments in code for more details.

Before the fix, if you would run the example that's part of the new test, you would have three files for the same dependency, where two of them were empty:

s

(The empty files were cause by the "fallback for pom-only dependencies" in `SynthesizeDependenciesWorkAction`)

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildProject.groovy
@@ -68,7 +68,7 @@ final class IncludedBuildProject extends AbstractProject {
   final Set<ProjectAdvice> expectedBuildHealth = [
     projectAdviceForDependencies(':', [
       Advice.ofRemove(
-        includedBuildCoordinates('second:second-build', projectCoordinates(':second-build')),
+        includedBuildCoordinates('second:second-build', projectCoordinates(':')),
         'implementation'
       )
     ] as Set<Advice>)
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildProject.groovy
@@ -68,7 +68,7 @@ final class IncludedBuildProject extends AbstractProject {
   final Set<ProjectAdvice> expectedBuildHealth = [
     projectAdviceForDependencies(':', [
       Advice.ofRemove(
-        includedBuildCoordinates('second:second-build', '1.0', projectCoordinates(':second-build')),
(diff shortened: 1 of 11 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildProject.groovy
```diff
@@ -68,7 +68,7 @@ final class IncludedBuildProject extends AbstractProject {
   final Set<ProjectAdvice> expectedBuildHealth = [
     projectAdviceForDependencies(':', [
       Advice.ofRemove(
-        includedBuildCoordinates('second:second-build', '1.0', projectCoordinates(':second-build')),
+        includedBuildCoordinates('second:second-build', projectCoordinates(':')),
```
jjohannes, line 71: The project coordinate `project(':second-build')` does not exist in the Gradle-sense. There is an included build in which the root project would be referred to as `project(':')`.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -69,7 +69,7 @@ internal fun ResolvedArtifactResult.toCoordinates(): Coordinates {
 
   return IncludedBuildCoordinates.of(
     requested = requested,
-    resolvedProject = identity
+    resolvedProject = resolved
```
jjohannes, line 72: Preserve the "real" project coordinate here. `identity` is an artificial project coordinate that does not really exists in Gradle. See also adjustment and comment on `IncludedBuildProject.groovy` test fixture.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithSubprojectsProject.groovy
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  final Set<ProjectAdvice> expectedBuildHealth = [
+    projectAdviceForDependencies(':', [] as Set<Advice>)
+  ]
+
+  // Health of the included build
+  Set<ProjectAdvice> actualIncludedBuildHealth() {
+    def included = gradleProject.includedBuilds[0]
+    def project = new GradleProject(new java.io.File(gradleProject.rootDir, 'second-build'), null, included, [], [])
```
autonomousapps, line 106: oo, looks like I have a missing API if you had to do this to get the advice 🙈

autonomousapps (approved):
Thanks for the contribution!

PR #811, opened 2022-11-15
Use information from 'module-info' to determine if a package is public

### LINKED ISSUE #800: api vs implementation: respect private package visibility of Java Modules

When using the Java Module System (aka JPMS), packages can be private. This is defined in the `module-info.java` (`module-info.class` in bytecode). When you have a `module-info.java` in a Java project in Gradle, Gradle automatically builds that project "as module", which means that it uses the concept of the Module Path and respects what is defined in `module-info.class` files. With this, all previously public API (classes, methods) that are inside a package that is **not** exported become private API.

It would be great if this plugin could respect that when deciding if a dependency scope should be `api` vs `implementation`.

I think it could just work on top of what is currently there:

1. If exists, also read the `module-info.class` of the currently analysed project.
2. When making the api vs implementation decision: If `module-info.class` exists, only consider types in public API that is in an exported package.

**Note**: I think it is not necessary to further use this information outside of a certain project (by consumers of that project) or to read the `module-info` of 3rd party Jars. Even if a project/Jar has a `module-info`, it may be used as plain Jar by other projects. And then the information in `module-info` is not relevant (for the consumer). On the other hand, if the module-info is used by the consumer, the compiler will already forbid to use and API from private packages. There is no need for this plugin to check that.

...

**Additional context**

@autonomousapps if you think this is in scope and can be added without interfering with anything else, I would attempt to implement it and provide a PR.

jjohannes:
Read the information from `module-info` (if available) and use it to decide if a package if public or not.
This can influences if a dependency declaration should be `implementation` or `api`.

Resolves #800


autonomousapps (approved):
Thank you!

PR #812, opened 2022-11-16
Use the gradle build github action

oheyadam:
The migrates the project to Gradle's own [gradle-build-action](https://github.com/gradle/gradle-build-action). This plugin would allow running the build with different Gradle versions, and handles caching better than `actions/cache`. Details [here](https://github.com/gradle/gradle-build-action#caching)

The PR is split into 3 commits. The first one configures GH actions to cancel a running build if a newer one is pushed to the same branch to save resources, the second one updates the existing actions to their respective latest versions, and the third one migrates to this particular action. I can split these into different PRs if this one has too many changes.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
       - name: Cache gradle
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -11,56 +11,51 @@ on:
...
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   gradle:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ ubuntu-latest ]
```
oheyadam, line 22: I think this can be simplified to run on `ubuntu-latest` directly as it's currently not really taking advantage of the matrix strategy, but I left it as is for now

autonomousapps: Every now and then someone complains about a Windows-specific failure and I'm _almost_ sympathetic enough to try to run CI on Windows, but I think that would be a big lift, absent funding that is 🙃

autonomousapps (approved):
LGTM! Thanks :)

autonomousapps:
Looks like the check is using the updated workflow files! So that's cool.

PR #814, opened 2022-11-17
Add flag for ignoring android variants

### LINKED ISSUE #519: Select the variants to run on

Our build has a few variants, (flavor / build type combo's), some of them are invalid in certain circumstances. However it seems this plugin runs on all combinations?

Is there a way, that we can pass in what variants we want to run on?

For example, perhaps I just want to analyse the `debugPaid`, `debugFree` & `releaseFree` variants but not the `releasePaid` (as a contrived exampled).  Or another example run on `debugUnitTest` but skip `releaseUnitTest` (right now you can disable all test analysis, but not a subset `systemProp.dependency.analysis.test.analysis = false`)

My gradle foo isn't up to scratch but I imagine you can also disable these variants elsewhere outside the plugin, but that wouldn't be explicit as to the reason of why we are doing it.

**Additional context**

Thanks for your time!
Sorry if this is obviously a thing I should do in gradle and not in your plugin. 😅

sboishtyan:
Fixes #519
Hi I saw someone tried to resolve this by #613 I think one flag will be enough

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
@@ -1,0 +1,64 @@
+package com.autonomousapps.android
+
+import com.autonomousapps.android.projects.AllVariantsIgnoredProject
+import com.autonomousapps.android.projects.ReleaseVariantIgnoredProject
+import com.autonomousapps.android.projects.DebugVariantIgnoredProject
+
+import static com.autonomousapps.advice.truth.BuildHealthSubject.buildHealth
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertAbout
+
+@SuppressWarnings("GroovyAssignabilityCheck")
+final class IgnoredVariantSpec extends AbstractAndroidSpec {
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored debug)"() {
+    given:
+    def project = new DebugVariantIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth',)
+
+    then:
+    assertAbout(buildHealth())
+      .that(project.actualBuildHealth())
+      .isEquivalentIgnoringModuleAdvice(project.expectedBuildHealth())
+
+    where:
+    [gradleVersion, agpVersion] << gradleAgpMatrix()
+  }
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored release)"() {
+    given:
+    def project = new ReleaseVariantIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth',)
+
+    then:
+    assertAbout(buildHealth())
+      .that(project.actualBuildHealth())
+      .isEquivalentIgnoringModuleAdvice(project.expectedBuildHealth())
+
+    where:
+    [gradleVersion, agpVersion] << gradleAgpMatrix()
+  }
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored all variants)"() {
+    given:
+    def project = new AllVariantsIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth',)
+
+    then:
+    assertAbout(buildHealth())
+      .that(project.actualBuildHealth())
+      .isEquivalentIgnoringModuleAdvice(project.expectedBuildHealth())
+
+    where:
+    [gradleVersion, agpVersion] << gradleAgpMatrix()
+  }
+}
--- a/src/functionalTest/groovy/com/autonomousapps/android/projects/AbstractVariantProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/projects/AbstractVariantProject.groovy
@@ -3,7 +3,6 @@ package com.autonomousapps.android.projects
 
 import com.autonomousapps.AbstractProject
 import com.autonomousapps.kit.*
-import com.autonomousapps.model.Advice
 import com.autonomousapps.model.ProjectAdvice
 
 import static com.autonomousapps.AdviceHelper.*
@@ -13,12 +12,12 @@ import static com.autonomousapps.AdviceHelper.*
  * use `implementation` but should be on `debugImplementation` and `releaseImplementation`, and a
  * `debugImplementation` dependency will not be seen as unused.
  */
-final class VariantProject extends AbstractProject {
+abstract class AbstractVariantProject extends AbstractProject {
 
   final GradleProject gradleProject
-  private final String agpVersion
+  protected final String agpVersion
 
-  VariantProject(String agpVersion) {
+  AbstractVariantProject(String agpVersion) {
     this.agpVersion = agpVersion
     this.gradleProject = build()
   }
@@ -26,7 +25,7 @@ final class VariantProject extends AbstractProject {
   private GradleProject build() {
     def builder = newGradleProjectBuilder()
     builder.withRootProject { root ->
-      root.gradleProperties = GradleProperties.minimalAndroidProperties()
+      root.gradleProperties = projectGradleProperties
       root.withBuildScript { bs ->
         bs.buildscript = BuildscriptBlock.defaultAndroidBuildscriptBlock(agpVersion)
       }
@@ -46,14 +45,14 @@ final class VariantProject extends AbstractProject {
     return project
   }
 
-  private List<Plugin> plugins = [
+  protected final List<Plugin> plugins = [
     Plugin.androidAppPlugin,
     Plugin.kotlinAndroidPlugin
   ]
 
-  private AndroidBlock androidBlock = AndroidBlock.defaultAndroidAppBlock(true)
+  protected final AndroidBlock androidBlock = AndroidBlock.defaultAndroidAppBlock(true)
 
-  private List<Dependency> dependencies = [
+  protected final List<Dependency> dependencies = [
     Dependency.kotlinStdLib("implementation"),
     Dependency.appcompat("implementation"),
     Dependency.constraintLayout("implementation"),
@@ -65,7 +64,7 @@ final class VariantProject extends AbstractProject {
     Dependency.commonsMath("debugImplementation")
   ]
 
-  private List<Source> sources = [
+  protected final List<Source> sources = [
     new Source(
       SourceType.KOTLIN, "MainActivity", "com/example",
       """\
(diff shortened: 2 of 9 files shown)


sboishtyan:
@autonomousapps Hi Idk how add you as a review
I could write a test If you help me copy-paste a little bit.
I saw the VariantSpec I could add a test there or create a new IgnoredVariantSpec. What do you think?

src/main/kotlin/com/autonomousapps/Flags.kt
```diff
@@ -16,10 +16,17 @@ object Flags {
   private const val FLAG_MAX_CACHE_SIZE = "dependency.analysis.cache.max"
   private const val FLAG_TEST_ANALYSIS = "dependency.analysis.test.analysis"
   private const val FLAG_PRINT_BUILD_HEALTH = "dependency.analysis.print.build.health"
+  /**
+   * Android build variant to not analyze i.e.
+   *
+   * dependency.analysis.android.ignored_variants=release
+   */
+  private const val FLAG_ANDROID_IGNORED_VARIANTS = "dependency.analysis.android.ignored_variants"
```
autonomousapps, line 24: I think this should be `dependency.analysis.android.ignored.variants`. From experience, I know that having different word delimiters can be annoying.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -126,8 +127,13 @@ internal class ProjectPlugin(private val project: Project) {
 
   /** Has the `com.android.application` plugin applied. */
   private fun Project.configureAndroidAppProject() {
+    val project = this
     val appExtension = the<AppExtension>()
-    appExtension.applicationVariants.all {
+    val ignoredVariantNames = androidIgnoredVariants()
+    val allowedVariants = appExtension.applicationVariants.matching { variant ->
+      !ignoredVariantNames.contains(variant.name)
```
autonomousapps, line 134: do we really want to use `contains()` instead of `==`?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -177,7 +183,13 @@ internal class ProjectPlugin(private val project: Project) {
 
   /** Has the `com.android.library` plugin applied. */
   private fun Project.configureAndroidLibProject() {
-    the<LibraryExtension>().libraryVariants.all {
+    val project = this
+    val libraryExtension = the<LibraryExtension>()
+    val ignoredVariantNames = androidIgnoredVariants()
+    val allowedVariants = libraryExtension.libraryVariants.matching { variant ->
+      !ignoredVariantNames.contains(variant.name)
```
autonomousapps, line 190: `contains()` vs `==`

autonomousapps (changes requested):
Thanks for the PR! I'm feeling generally positive about the direction, but have a couple of change requests/questions. In addition, as you mentioned, this needs a test. Feel free to copy-paste something that looks close and modify it for this case. I'd also like some documentation. Maybe you could propose something and I could add it to the wiki.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
sboishtyan: As I see because ignoredVariantNames is a collection we can't use `==` to compare those params.
So for clarify, we have set of `ignoredVariantNames` and we should intersect it with `applicationVariants`

sboishtyan:
> Thanks for the PR! I'm feeling generally positive about the direction, but have a couple of change requests/questions. In addition, as you mentioned, this needs a test. Feel free to copy-paste something that looks close and modify it for this case. I'd also like some documentation. Maybe you could propose something and I could add it to the wiki.

I will add a test. Could you help me with place where I should put documentation about this flag?

sboishtyan:
Simple note that you already don't have test for `FLAG_TEST_ANALYSIS`
Maybe good first issue for the future contributors

src/functionalTest/groovy/com/autonomousapps/android/projects/AbstractVariantProject.groovy
```diff
@@ -13,20 +12,20 @@ import static com.autonomousapps.AdviceHelper.*
  * use `implementation` but should be on `debugImplementation` and `releaseImplementation`, and a
  * `debugImplementation` dependency will not be seen as unused.
  */
-final class VariantProject extends AbstractProject {
+abstract class AbstractVariantProject extends AbstractProject {
```
sboishtyan, line 15: @autonomousapps So i tried to reuse that project

src/functionalTest/groovy/com/autonomousapps/android/projects/AbstractVariantProject.groovy
```diff
@@ -46,14 +45,14 @@ final class VariantProject extends AbstractProject {
     return project
   }
 
-  private List<Plugin> plugins = [
+  protected final List<Plugin> plugins = [
```
sboishtyan, line 48: Those private property weren't seen in the closure
```
builder.withRootProject {}
```
As I understand Closure works like that I see actual class not an abstract one because of that I marked those props protected

src/main/kotlin/com/autonomousapps/Flags.kt
```diff
@@ -16,10 +16,17 @@ object Flags {
   private const val FLAG_MAX_CACHE_SIZE = "dependency.analysis.cache.max"
   private const val FLAG_TEST_ANALYSIS = "dependency.analysis.test.analysis"
   private const val FLAG_PRINT_BUILD_HEALTH = "dependency.analysis.print.build.health"
+  /**
+   * Android build variant to not analyze i.e.
+   *
+   * dependency.analysis.android.ignored.variants=release
+   */
+  const val FLAG_ANDROID_IGNORED_VARIANTS = "dependency.analysis.android.ignored.variants"
```
sboishtyan, line 24: public for reusing in tests

src/main/kotlin/com/autonomousapps/Flags.kt
sboishtyan: done

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
autonomousapps: oh sorry! I wasn't thinking clearly. That's a list 🤦

sboishtyan:
@autonomousapps So, today I add one more test scenario if user will ignore all variants
Everything works correctly as for me.
The next steps are add documentation and get your feedback. So feel free to give it when you will have a time.
Then I could make nightly build and test it on my project

autonomousapps:
@sboishtyan would you mind rebasing?

sboishtyan:
> @sboishtyan would you mind rebasing?

done

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+@SuppressWarnings("GroovyAssignabilityCheck")
+final class IgnoredVariantSpec extends AbstractAndroidSpec {
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored debug)"() {
+    given:
+    def project = new DebugVariantIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'clean', 'buildHealth', '--no-build-cache')
```
autonomousapps, line 20: why `clean` and `--no-build-cache`?

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+    [gradleVersion, agpVersion] << gradleAgpMatrix()
+  }
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored release)"() {
+    given:
+    def project = new ReleaseVariantIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'clean', 'buildHealth', '--no-build-cache')
```
autonomousapps, line 37: same question

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+    [gradleVersion, agpVersion] << gradleAgpMatrix()
+  }
+
+  def "plugin ignore android variants (#gradleVersion AGP #agpVersion ignored all variants)"() {
+    given:
+    def project = new AllVariantsIgnoredProject(agpVersion)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'clean', 'buildHealth', '--no-build-cache')
```
autonomousapps, line 54: and same question

src/main/kotlin/com/autonomousapps/Flags.kt
autonomousapps: It can actually be `private` (which I would prefer) because Groovy can access the private field regardless (Groovy is a jerk).

autonomousapps (changes requested):
Looks really good, thanks! Just make that one constant private, and I think we're good to go.

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
sboishtyan: it's copy-paste from `VariantSpec`. I will remove that

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
sboishtyan: done

src/functionalTest/groovy/com/autonomousapps/android/IgnoredVariantSpec.groovy
sboishtyan: done

sboishtyan:
As for documentation we could add:
```
Dependency analysis is heavy. If you have many build variants in your android application or library those are not much different from each other you maybe want to skip analysis for them.

For example we have `debug`, `release` and `staging`.
`staging` in our case is a build variant that as same as `release` but have different resources. But dependencies in `staging` and `release` are the same.  So we assume that you want to skip analysis for `staging` and you can with flag: `dependency.analysis.android.ignored.variants`.

Add to your `gradle.properties` file line similar to `dependency.analysis.android.ignored.variants=<list of build variants separated by coma>` i.e.  `dependency.analysis.android.ignored.variants=staging,debug`
Those variants will be excluded from analysis.
```

autonomousapps (approved):
LGTM. If it's green, will merge. Thanks!

sboishtyan:
Tell me please what release will contain those changes

autonomousapps:
it'll be in 1.18.0.

PR #819, opened 2022-11-21
Fix NPE in asmUtil.exportedPackages()

jjohannes:
Although you cannot see this in the Kotlin code, this (coming out of ASM) can be `null`.

Sorry for this follow up, I was tricked by being in Kotlin, not even thinking about that this can be `null`. 😐 

Follow up to #800


autonomousapps (approved):
Thanks again :)

PR #824, opened 2022-11-23
Use buffered writer when writing ExplodedJar to file

### LINKED ISSUE #823: OutOfMemoryError in ExplodeJarTask
**Build scan link**
Private, can't share unfortunately.

**Plugin version**
`v1.13.1`, however no changes in `ExplodeJarTask` in ` main..v1.13.1`

**Gradle version**
`7.4.2`

Not applicable.

Not applicable.

**Describe the bug**
When serializing the output of `ExplodeJarTask` to file, it sometimes throws an `OutOfMemoryError`:
```
Caused by: java.lang.OutOfMemoryError: Java heap space
at okio.Buffer.readString(Buffer.kt:313)
at okio.Buffer.readUtf8(Buffer.kt:297)
at com.squareup.moshi.JsonAdapter.toJson(JsonAdapter.java:94)
at com.autonomousapps.tasks.ExplodeJarTask$ExplodeJarWorkAction.execute(ExplodeJarTask.kt:126)
```

Specifically the call `explodedJars.toJson()` causes this. To give some context, `explodedJars.toJson().length = 5385952` in our case. The daemon is configured to run with 8Gb already.

**To Reproduce**
It's not happening deterministically unfortunately,  but having an insane number of dependencies sometimes triggers this.

**Expected behavior**
No OOM exception.

**Additional context**
From some initial investigation the entire `Set<ExplodedJar>` is first serialized to json and then written to file. Maybe it makes more sense to serialize the `ExplodedJar` one by one?

daanschipper:
Fixes #823

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
@@ -10,6 +10,8 @@ import com.squareup.moshi.*
 import com.squareup.moshi.Types.newParameterizedType
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import dev.zacsweers.moshix.sealed.reflect.MetadataMoshiSealedJsonAdapterFactory
+import okio.buffer
+import okio.sink
 import java.io.File
 
 val MOSHI: Moshi by lazy {
@@ -104,6 +106,32 @@ inline fun <reified T> T.toPrettyString(withNulls: Boolean = false): String {
 
 inline fun <reified K, reified V> Map<K, V>.toPrettyString(withNulls: Boolean = false): String {
   return getJsonMapAdapter<K, V>(withNulls).indent("  ").toJson(this)
+}
+
+/**
+ * Buffers writes of the set to disk, using the indent to make the output human-readable.
+ * By default, the output is compacted.
+ *
+ * @param set The set to write to file
+ * @param indent The indent to control how the result is formatted
+ */
+inline fun <reified T> File.bufferWriteJsonSet(set: Set<T>, indent: String = "") {
+  JsonWriter.of(sink().buffer()).use { writer ->
+    getJsonSetAdapter<T>().indent(indent).toJson(writer, set)
+  }
+}
+
+/**
+ * Buffers writes of the set to disk, using the indent to make the output human-readable.
+ * By default, the output is compacted.
+ *
+ * @param set The set to write to file
+ * @param indent The indent to control how the result is formatted
+ */
+inline fun <reified T> File.bufferWriteJson(set: T, indent: String = "") {
+  JsonWriter.of(sink().buffer()).use { writer ->
+    getJsonAdapter<T>().indent(indent).toJson(writer, set)
+  }
 }
 
 @Suppress("unused")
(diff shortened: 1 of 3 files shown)


daanschipper:
Disclaimer: I've not used moshi before, open for suggestions on some improvement. I'm not sure about the pretty output, I'd was looking for something along the lines of `getJsonSetAdapter<ExplodedJar>().toPrettyJson()` but couldn't find it.

autonomousapps (approved):
Thank you, I really appreciate this! Will merge once green on CI.

autonomousapps:
There are an awful lot of test failures, and unfortunately that means the build scan is too large to be viewed.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
```diff
@@ -106,6 +108,32 @@ inline fun <reified K, reified V> Map<K, V>.toPrettyString(withNulls: Boolean =
...
+ */
+inline fun <reified T> File.bufferWriteJsonSet(jsonObject: Set<T>, indent: String = "") {
+  val writer = JsonWriter.of(this.sink().buffer())
+  getJsonSetAdapter<T>().indent(indent).toJson(writer, jsonObject)
+  writer.close() // need to explicitly close the writer
+}
+
+/**
+ * Buffers writes the json object to file using the indent to make the output human-readable.
+ * By default the output is compacted.
```
autonomousapps, line 126: > Buffers writes of the set to disk, using the indent to make the output human-readable. By default, the output is compacted.

use that for both new methods :)

src/functionalTest/groovy/com/autonomousapps/jvm/ExplodeJarSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+
+import com.autonomousapps.jvm.projects.ExplodeJarProject
+import com.autonomousapps.kit.utils.Files
+
+import java.util.stream.Collectors
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class ExplodeJarSpec extends AbstractJvmSpec {
```
autonomousapps, line 11: I'm a little worried about the fragility of this test. Any change to the relevant models will break the test, but that, to me, is just noise—not useful information.

I think you should just delete this spec, the `ExplodeJarProject`, and the two new resources files as well. I don't think they add anything useful when compared to the maintenance cost.

autonomousapps (changes requested):
Thanks for the fix. Left some comments.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
```diff
@@ -106,6 +108,32 @@ inline fun <reified K, reified V> Map<K, V>.toPrettyString(withNulls: Boolean =
...
+ *
+ * @param jsonObject The set to write to file
+ * @param indent The indent to control how the result is formatted
+ */
+inline fun <reified T> File.bufferWriteJson(jsonObject: T, indent: String = "") {
+  val writer = JsonWriter.of(this.sink().buffer())
+  getJsonAdapter<T>().indent(indent).toJson(writer, jsonObject)
+  writer.close() // need to explicitly close the writer
+}
```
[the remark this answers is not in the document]
daanschipper: Was not aware of `.use { }`, that looks cleaner indeed!

autonomousapps (approved):
Thanks!

PR #827, opened 2022-12-01
Support string value `?`

### LINKED ISSUE #826: NullPointerException when executing explodeXmlSourceDebug
**Plugin version**
1.16.0

**Gradle version**
7.6

7.3.1

**Describe the bug**
When running the tasks `buildHealth printDuplicateDependencies --stacktrace` on a large codebase, we encounter this NPE:

```

* What went wrong:
Execution failed for task ':[MyModule]:explodeXmlSourceDebug'.
> A failure occurred while executing com.autonomousapps.tasks.XmlSourceExploderTask$XmlSourceExploderWorkAction
> java.lang.NullPointerException (no error message)
```

This module is an Android library module with some plugins applied to it (`org.jetbrains.kotlin.kapt`, `androidx.navigation.safeargs.kotlin`, `kotlin-parcelize`) like many others in the codebase.

I can't find where this NPE is exactly located in the `XmlSourceExploderWorkAction` (even with the `--stacktrace` option)

SimonMarquis:
Fixes #826


SimonMarquis:
Not 100% sure about this commit https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/827/commits/a215f50c9fadf26867b98add51561489047232dd

The type signature `fun from(mapEntry: Pair<String, String>)` does not help distinguishing what it represents :)

src/main/kotlin/com/autonomousapps/model/Source.kt
```diff
@@ -120,15 +120,15 @@ data class AndroidResSource(
...
           )
           else -> null
         }
       }
 
-      private fun Pair<String, String>.isId() = first.startsWith("@+") || second.startsWith("@id")
+      private fun Pair<String, String>.isId() = second.startsWith("@+") || second.startsWith("@id")
       private fun Pair<String, String>.isToolsAttr() = first.startsWith("tools:")
-      private fun Pair<String, String>.isDataBindingExpression() = first.startsWith("@{") && first.endsWith("}")
+      private fun Pair<String, String>.isDataBindingExpression() = second.startsWith("@{") && second.endsWith("}")
```
autonomousapps, line 131: Why did you change these?

autonomousapps:
Haha whoops, didn't mean to mark as ready for review. Just wanted to see the results of the CI run.

src/main/kotlin/com/autonomousapps/model/Source.kt
SimonMarquis: From what I understand, a databinding XML file containing this instruction `binding:text="@{`Hello`}"` will be parsed as a `Pair(first = "binding:text", second = "@{`Hello`}")`? Or did I mixed things?

autonomousapps: Well, first, you have me wondering if there was a typo in my earlier code, except for the fact that earlier code was backed by tests that verified that it works. And then second, this change has no obvious connection to the fix you want to make, so I think you shouldn't be doing it because it conflates things. And in general I'd prefer you not change variable names without a good reason (which you also did in this PR). Again, it conflates your fix with other changes, making it harder for me to verify that your fix does just what it says and nothing else.

SimonMarquis: Sure, I'll revert those changes.
Can you show me which and where are these tests? I'm trying to understand how this AttrRef.from gets its values from.

autonomousapps (commented):
This new changeset is so much cleaner, thank you!

src/main/kotlin/com/autonomousapps/model/Source.kt
autonomousapps: Check out `ResSpec`. E.g.,
```groovy
def "gracefully handles dataBinding expressions in res files (#gradleVersion AGP #agpVersion)"
```

autonomousapps: btw, if you discover I have a typo and a hidden bug, do let me know! 😅

SimonMarquis: Thanks, I'll take a look at that test class before making this PR "ready for review".
I'm currenly trying to re-enable it on a fairly large codebase (~900 modules atm) so I might discover some things along the way.

autonomousapps: For me to be really comfortable merging this, I'd like a new test verifying that it resolves your issue. You could add a new spec to `ResSpec`.

SimonMarquis: I've created an `AndroidTextQuestionMarkProject` containing a single XML layout file with the `android:text="?"` that was causing the issue. I've confirmed it produced the same NPE without the fix in https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/827/commits/ca2b1f39a293cd3e7af987e17baee48431bc9804.
I tried to follow the conventions of others project classes, but I'm not really used to Groovy, so I might have done something not quite right.
Anyway, you can take a look at it when you'll have time.

src/main/kotlin/com/autonomousapps/model/Source.kt
```diff
@@ -104,7 +104,7 @@ data class AndroidResSource(
 
         val id = mapEntry.second
         return when {
-          id.startsWith('?') -> AttrRef(
+          ATTR_REGEX.containsMatchIn(id) -> AttrRef(
```
autonomousapps, line 107: Sorry for the long delay in re-reviewing. Work has been very busy, and I wanted time to think about this change, which is more subtle than might be obvious.

**tl;dr** I think this would be safer it it were `mapEntry.first != "android:text" && id.startsWith('?')`. WDYT? I'll explain some of the rationale in a comment directly on your new unit test.

src/test/kotlin/com/autonomousapps/model/AttrRefTest.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.autonomousapps.model.AndroidResSource.AttrRef.Companion.from
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Test
+
+class AttrRefTest {
+
+    @Test
+    fun `question mark character as inlined text is not parsed as an AttrRef`() {
+        assertNull(from("android:text" to "?"))
```
autonomousapps, line 11: What do you think the result should be if the test were
```kotlin
assertNull(from("android:text" to "?foo"))
```
?

With your current changeset, the test fails because `from(...)` returns
```
expected: <null> but was: <AttrRef(type=attr, id=foo)>
Expected :null
Actual   :AttrRef(type=attr, id=foo)
```
which is probably not what we want? Presumably `android:text` should be able to display any valid string, including things of the form `"?foo"`, without the Android system thinking that it's a reference to some attribute named "foo".

SimonMarquis: Unfortunately this is not how Android (AAPT) handles this special case 😖

For instance, this layout is invalid:

```xml
<TextView android:text="?test" />
```

and produces:

```
AGPBI: {"kind":"error","text":"Android resource linking failed","sources":[{"file":".\\app\\build\\intermediates\\incremental\\debug\\mergeDebugResources\\stripped.dir\\layout\\activity_main.xml","position":{"startLine":1}}],"original":"ERROR:.\\app\\build\\intermediates\\incremental\\debug\\mergeDebugResources\\stripped.dir\\layout\\activity_main.xml:1: AAPT: error: resource attr/test (aka com.example:attr/test) not found.\n    ","tool":"AAPT"}

* What went wrong:
Execution failed for task ':app:processDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
   > Android resource linking failed
     ERROR:.\app\build\intermediates\incremental\debug\mergeDebugResources\stripped.dir\layout\activity_main.xml:1: AAPT: error: resource attr/test (aka com.example:attr/test) not found.
```

The same thing happen with resource files:

```xml
<resources>
    <string name="referrer_test">?test</string>
</resources>
```

produces:

```
AGPBI: {"kind":"error","text":"Android resource linking failed","sources":[{"file":".\\app\\src\\main\\res\\values\\styles.xml","position":{"startLine":8,"startColumn":4,"startOffset":324,"endLine":15,"endColumn":12,"endOffset":792}}],"original":"ERROR:.\\app\\src\\main\\res\\values\\styles.xml:9:5-16:13: AAPT: error: resource attr/test (aka com.example:attr/test) not found.\n    ","tool":"AAPT"}

* What went wrong:
Execution failed for task ':app:processDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
   > Android resource linking failed
     ERROR:.\app\src\main\res\values\styles.xml:9:5-16:13: AAPT: error: resource attr/test (aka com.example:attr/test) not found.
```

So I guest the following test matches the expected Android behavior:

```kotlin
assertEquals(
  AttrRef(type = "attr", id = "foo"),
  from("android:text" to "?foo"),
)
```

---

But this bring one last special case. If someone were to use something like that:

```xml
<TextView android:text="\?test" />
```

Then, AAPT "sees" the string "?test", and the corresponding test should be:

```kotlin
assertNull(from("android:text" to "\\?foo"))
```

Which means we should not use `Regex.containsMatchIn(CharSequence)` but rather `Regex.matchEntire(CharSequence)`.
Then, the current `ATTR_REGEX.matchEntire(id) != null -> AttrRef(...)` would behave like AAPT.

autonomousapps: omg I hate AAPT.

src/test/kotlin/com/autonomousapps/model/AttrRefTest.kt
```diff
@@ -0,0 +1,29 @@
+package com.autonomousapps.model
+
+import com.autonomousapps.model.AndroidResSource.AttrRef
+import com.autonomousapps.model.AndroidResSource.AttrRef.Companion.from
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Test
+
+class AttrRefTest {
```
autonomousapps, line 9: Nice tests, thanks.

src/main/kotlin/com/autonomousapps/model/Source.kt
```diff
@@ -82,8 +82,8 @@ data class AndroidResSource(
   data class AttrRef(val type: String, val id: String) {
     companion object {
 
-      private val TYPE_REGEX = Regex("""@(?:.+:)?(.+)/""")
-      private val ATTR_REGEX = Regex("""\?(?:.+/)?(.+)""")
+      private val TYPE_REGEX = Regex("""@(\+|\w+:)?(?<type>\w+)/(\w+)""")
+      private val ATTR_REGEX = Regex("""\?((\w+:\w+|\w+)/)?(?<attr>\w+)""")
```
autonomousapps, line 86: I'm going to need to spend a few minutes studying these regexes to ensure I understand them 😅

autonomousapps (approved):
Looks good, will merge after I convince myself I understand the updated regex.

src/main/kotlin/com/autonomousapps/model/Source.kt
SimonMarquis: :D if that helps in any way, I can add dedicated unit tests for that as well.

SimonMarquis: And here is an easier way to "visualize" them with examples:
- https://regex101.com/r/N3ZPTB/1
- https://regex101.com/r/GH2uoF/1

autonomousapps: If you want you could add tests, but even just a comment explaining what each component of the regex does would be _super_ helpful.

SimonMarquis: 👍 I've been able to find links to the official documentation regarding these regexes.

SimonMarquis: `?android:foo` seems to be valid as well, I'll update the regex accordingly

src/main/kotlin/com/autonomousapps/model/Source.kt
```diff
@@ -82,8 +82,27 @@ data class AndroidResSource(
...
+       * This will match references to style attributes `?[<package_name>:][<resource_type>/]<resource_name>`:
+       *
+       * - `?foo`
+       * - `?attr/foo`
+       * - `?android:foo`
+       * - `?android:attr/foo`
+       *
+       * @see <a href="https://developer.android.com/guide/topics/resources/providing-resources#ReferencesToThemeAttributes">Referencing style attributes</a>
+       */
+      private val ATTR_REGEX = Regex("""\?(\w+:)?(\w+/)?(?<attr>\w+)""")
```
autonomousapps, line 105: I love it, thanks for the docs!

PR #830, opened 2022-12-08
Possible fix for excluded annotations to be removed from ABI

### LINKED ISSUE #831: Excluded annotations removes annotated class completely form the ABI, not the annotation itself.
**Build scan link**
Private, can't share unfortunately.

**Plugin version**
v1.17.0

**Gradle version**
7.4.2

We had some issues with excluded annotations:

**What we expected:**
To have only the excluded annotations excluded from the API (hence they can be implementation instead of api).

**What happened instead:**
Whenever a class A had some excluded annotations, the entire class was considered excluded (not just the annotations itself). This caused quite weird behavior and doesn't seem likely to be intended.

Breus:
Hi, 

We had some issues with excluded annotations:

**What we expected:** 
To have only the excluded annotations excluded from the API (hence they can be implementation instead of api).

**What happened instead:**
Whenever a class A had some excluded annotations, the entire class was considered excluded (not just the annotations itself). This caused quite weird behavior and doesn't seem likely to be intended.

To showcase what I mean, I've created this PR to include (what I think) the fix for this issue and corresponding tests.

I've never written any Kotlin before, so feel free to refactor wherever you see needed 😄 

Resolves #831

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/AbiExclusionsProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/AbiExclusionsProject.groovy
@@ -32,6 +32,7 @@ final class AbiExclusionsProject extends AbstractProject {
             abi {
               exclusions {
                 excludeClasses("com\\\\.example\\\\.Main")
+                excludeClasses("com\\\\.example\\\\.FactoryFactory")
                 excludeAnnotations(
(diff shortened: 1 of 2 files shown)


autonomousapps:
Per https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/831#issuecomment-1373337726, this PR cannot be merged as-is. Closing for now. Feel free to re-open if the discussion changes.

PR #833, opened 2022-12-19
Buffered reads and writes

daanschipper:


src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
```diff
@@ -10,10 +10,14 @@ import com.squareup.moshi.*
 import com.squareup.moshi.Types.newParameterizedType
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import dev.zacsweers.moshix.sealed.reflect.MetadataMoshiSealedJsonAdapterFactory
+import okio.BufferedSource
 import okio.buffer
 import okio.sink
 import java.io.File
 
+const val noJsonIndent = ""
+const val prettyJsonIndent = "  "
```
autonomousapps, line 19: can these be private?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -181,17 +181,16 @@ abstract class ComputeAdviceTask @Inject constructor(
         moduleAdvice = androidScore
       )
 
-      output.writeText(projectAdvice.toJson())
+      output.bufferWriteJson(projectAdvice)
       // These must be transformed so that the Coordinates are Strings for serialization
-      dependencyUsagesOut.writeText(dependencyUsages.toSimplifiedJson())
-      annotationProcessorUsagesOut.writeText(annotationProcessorUsages.toSimplifiedJson())
-      // TODO consider centralizing this logic in a separate PR
```
autonomousapps, line 188: Can we get these two comments back?

src/main/kotlin/com/autonomousapps/tasks/PrintDominatorTreeTask.kt
```diff
@@ -21,6 +21,6 @@ abstract class PrintDominatorTreeTask : DefaultTask() {
   abstract val consoleText: RegularFileProperty
 
   @TaskAction fun action() {
-    logger.quiet(consoleText.readText())
+    consoleText.readLines().forEach { logger.quiet(it) }
```
autonomousapps, line 24: Why this change?

autonomousapps (changes requested):
Thank you! Looks pretty good, but I left a couple of questions. There's also a (probably small) merge conflict to deal with.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
daanschipper: No, getting `Error:(98, 46) Public-API inline function cannot access non-public-API 'private const val prettyJsonIndent: String defined in com.autonomousapps.internal.utils in file moshi.kt'` for `inline fun <reified T> T.toPrettyString(withNulls: Boolean = false): String` and others then.

src/main/kotlin/com/autonomousapps/tasks/PrintDominatorTreeTask.kt
daanschipper: The `readText()` method loads the entire file in memory at once which was proven to be problematic in other places. Refactored it like this as it's functionally the same.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
autonomousapps: what about `internal`, then?

Oh hm, I note that all the functions here are public. I feel like I did that for a reason, but I don't remember now. I guess it's fine, I can take a look myself separately.

src/main/kotlin/com/autonomousapps/tasks/PrintDominatorTreeTask.kt
autonomousapps: Can you share the output here as an example so that I can see what it looks like?

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
daanschipper: Same issue `Public-API inline function cannot access non-public-API 'internal const val prettyJsonIndent: String defined in com.autonomousapps.internal.utils in file moshi.kt'`. I'm not too familiar with Kotlin and the extension functions, open for suggestions on how to change it :)

src/main/kotlin/com/autonomousapps/tasks/PrintDominatorTreeTask.kt
daanschipper: After giving it a bit more thought it's slightly different, the file will be printed but now line by line which means a task running in parallel could print other stuff in between which is undesired. Will revert this particular line.

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
autonomousapps: No worries, I'll take a look at this myself! Thanks.

autonomousapps (approved):
LGTM.

Issue #794, opened 2022-09-28
Use Gradle Version Catalogs if present


inktomi:
**Is your feature request related to a problem? Please describe.**

In a large project which makes use of version catalogs, it would be ideal if instead of suggestions like "add implementation 'androidx.activity:activity:1.4.0'" the plugin was able to resolve the dependency name from the gradle version catalog and instead report "add implementation libs.androidx.activity". 

**Describe the solution you'd like**

If there is a version catalog present, instead of reporting the raw dependency, look for a match in the catalog's items. There are methods off `VersionCatalog` which might be useful here. 

**Describe alternatives you've considered**

The alternative would be not doing this, which is reasonable - and perhaps there is not the API support from Gradle itself to do this reverse lookup. It'd be unfortunate though, as this lookup would make the suggestions and automatic fixes much more usable for projects using version catalogs.

autonomousapps:
The plugin currently has an API that would let you do this yourself. See [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/extension/DependenciesHandler.kt#L18-L27).

Alternatively, PRs are welcome (ideally leveraging the above-mentioned API). I don't have any immediate plans to implement this myself.

daanschipper:
For anyone interested, this is how you can add the dependencies in the version catalog:
```
dependencyAnalysis {
    dependencies {
        // Adds the defined aliases in the version catalog to be used when printing advice and rewriting build scripts.
        def versionCatalog = project.extensions.getByType(VersionCatalogsExtension).named("libs")
        versionCatalog.getLibraryAliases().forEach({ String alias ->
            map.put(versionCatalog.findLibrary(alias).get().get().toString(), "libs.${alias}")
        })
    }
}
```

BraisGabin:
I modified @daanschipper's snippet to work with the kotlin DSL:

```kotlin
dependencyAnalysis {
    structure {
        val versionCatalog = project.extensions.getByType<VersionCatalogsExtension>().named("libs")
        versionCatalog.libraryAliases.forEach { alias ->
            map.put(versionCatalog.findLibrary(alias).get().get().toString(), "libs.${alias}")
        }
    }
}
```

autonomousapps:
The plugin will automatically detect the presence of a version catalog and use those values whenever possible. This is complete and these workarounds aren't necessary.

PR #837, opened 2022-12-30
Advice to display Version Catalog aliases

Chasson1992:
#794

To allow advice to display the [Version Catalog](https://docs.gradle.org/current/userguide/platforms.html) aliases we can iterate through each alias and add a mapping of the alias to the dependency in the DependencyHandler#map extension.

My only concern is that the Version Catalog is for Gradle 7+, do you keep backward compatibility with Gradle versions?

I didn't add any functional tests, since it's fairly simple and safe, but I'm open to doing so.

A full working repository to demonstrate the issue is located [here](https://github.com/Chasson1992/DependencyAnalysisTest)

buildHealth output before the changes:

```
Advice for root project
Dependencies which should be removed or changed to runtime-only:
  runtimeOnly("ch.qos.logback:logback-classic:1.2.6") (was implementation)
```

buildHealth output after the changes:

```
Advice for root project
Dependencies which should be removed or changed to runtime-only:
  runtimeOnly(libs.logback.classic) (was implementation)
```

### DIFF
--- a/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
@@ -26,6 +26,7 @@ import com.autonomousapps.services.InMemoryCache
 import com.autonomousapps.tasks.*
 import org.gradle.api.Project
 import org.gradle.api.UnknownTaskException
+import org.gradle.api.artifacts.VersionCatalogsExtension
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.SourceSet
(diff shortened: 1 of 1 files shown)


autonomousapps:
Thanks for the PR!

autonomousapps (commented):
This is great, thank you.

Since this is the first feature to directly support the version catalog concept, I think having a functional test is worth doing, even though it does indeed look quite safe. The test's purpose would be more to ensure there isn't a future regression. Let me know if I can provide any pointers here.

Chasson1992:
@autonomousapps I started to implement a functional test (mirroring the test harness I created under https://github.com/Chasson1992/DependencyAnalysisTest).

I'm developing on a Windows 10 machine with IntelliJ Ultimate but when running the functionalTest task I'm getting a bunch of expections from:

```
Caused by: java.io.IOException at FileUtils.java:1344
    Caused by: java.nio.file.FileSystemException at WindowsException.java:92
```

I've been having a hard time figuring out the issue on this one. Any suggestions? I'll push what I have currently to my fork.

autonomousapps:
> I've been having a hard time figuring out the issue on this one. Any suggestions? I'll push what I have currently to my fork.

Can you post more of the stacktrace? TBH I don't develop at all on Windows, so I'm not sure I'll be able to help, but at the least I'd need the full stacktrace.

PR #846, opened 2023-01-13
bump kotlinx-metadata-jvm to 0.5.0

gabrielittner:
This supports reading Kotlin 1.8 and 1.9 metadata. The current version only goes until 1.7 and on our branch to update to Kotlin 1.8.0 we are seeing that dependency analysis does not find usages of Kotlin constants or extension methods anymore.

For it to work I also needed to update Kotlin itself.

Full changelog
https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ChangeLog.md

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -21,7 +21,7 @@ kotlin-dokka = { module = "org.jetbrains.dokka:kotlin-as-java-plugin", version.r
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
 kotlin-stdlib-jdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
-kotlinx-metadata-jvm = "org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.4.2"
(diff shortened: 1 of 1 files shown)


autonomousapps:
Lots of errors like this:
```
> Task :compileKotlin FAILED
w: Argument -Xopt-in is deprecated. Please use -opt-in instead
e: /home/runner/work/dependency-analysis-android-gradle-plugin/dependency-analysis-android-gradle-plugin/src/main/kotlin/com/autonomousapps/internal/asm.kt: (501, 7): Named arguments are not allowed for non-Kotlin functions
e: /home/runner/work/dependency-analysis-android-gradle-plugin/dependency-analysis-android-gradle-plugin/src/main/kotlin/com/autonomousapps/internal/asm.kt: (501, 7): Cannot find a parameter with this name: kind
e: /home/runner/work/dependency-analysis-android-gradle-plugin/dependency-analysis-android-gradle-plugin/src/main/kotlin/com/autonomousapps/internal/asm.kt: (502, 7): Named arguments are not allowed for non-Kotlin functions
```

gabrielittner:
Will take a look in the next days, I thought I fixed those already

gabrielittner:
I fixed that issue but there are some more things, when updating Kotlin, Moshi(X) needs to be updated as well but is then incompatible with metadata 0.6.0 because of the same breaking API change. So I've changed this pr to only update to 0.5.0 for now. That already comes with Kotlin 1.8.0 compatibility which was my original intention behind this pr. I'll open one for 0.6.0 a bit later.

PR #847, opened 2023-01-13
update tested Gradle and AGP versions

gabrielittner:
Updates tested Gradle version to 7.6 and tested AGP version to 7.4.0. 

I've also tried extending to AGP 8.0.0-alpha11 but that has 2 bigger hurdles (not unsolvable but requires more time):
- requires running `functionalTest` on JDK 17 but the oldest tested Gradle version does not run on that
- requires all Android projects to set a namespace instead of a package in the manifest


gabrielittner:
2 open issues that still need to be resolved:
- the toolchain for tests needs to be changed since AGP 8 needs to run on jdk8
- I assume we need to update all tests to use `namespace` instead of having a package in the manifest, however then the tests wouldn't cover `package` anymore for older AGP versions

autonomousapps:
> requires all Android projects to set a namespace instead of a package in the manifest

I've been aware of the need to make that update for a while now. It might be slightly painful 😅 

> requires running functionalTest on JDK 17 but the oldest tested Gradle version does not run on that

This may definitely be somewhat complicated, since I don't want to bump my plugin's minimum required Java version to 17. I agree that should be tackled separately, probably after resolving the first issue.

gabrielittner:
In the databinding tests the kotlin-stdlib is now considered unused. I added them as expected advice based on AGP version. The tests should now all be green but let me know if you prefer a different solution

src/functionalTest/groovy/com/autonomousapps/android/projects/DataBindingProject.groovy
```diff
@@ -0,0 +1,51 @@
...
+import com.autonomousapps.model.Advice
+import com.autonomousapps.model.ProjectAdvice
+import kotlin.Pair
+
+import static com.autonomousapps.AdviceHelper.moduleCoordinates
+import static com.autonomousapps.fixtures.Dependencies.APPCOMPAT
+import static com.autonomousapps.fixtures.Dependencies.KOTLIN_STDLIB_JDK7
+import static com.autonomousapps.fixtures.Dependencies.getDEPENDENCIES_KOTLIN_STDLIB
+
+final class DataBindingProject {
```
gabrielittner, line 14: Migrated this from Kotlin to Groovy so that `AgpVersion` becomes accessible

autonomousapps (approved):
LGTM!

PR #849, opened 2023-01-21
Kotlinx Metadata 0.6.0

gabrielittner:
Follow up to #846. This has more changes with the most notable one being that the minimum supported Gradle version is now 7.2, so it could still make sense to merge the other one for now.

Changes:
- replaced deprecated `KotlinClassHeader` with `Metadata`
- Kotlin had to be bumped to 1.7.22 because the metadata library is built with 1.8 and 1.6 can't handle that
- updated Moshi and MoshiX so that they support the new Kotlin version
- for `moshix-sealed-reflect` it is now using the non metadata artifact because it is not compatible with 0.6 yet
- updating Moshi required raising the minimum Gradle version to 7.2 because it requires a kotlin-stdlib method that was only added in that release (Does the plugin have no runtime check for a minimum Gradle version? I've searched but only found the test code)
- Dagger version in tests has been bumped because the previous version depended on a metadata version that can't process Kotlin 1.7


autonomousapps:
Looks like you need to resolve a conflict since I just merged your other PR @gabrielittner

gabrielittner:
Resolved it

autonomousapps:
There are some failing tests https://scans.gradle.com/s/74vddty5c3sry

gradle/libs.versions.toml
```diff
@@ -21,15 +21,14 @@ kotlin-dokka = { module = "org.jetbrains.dokka:kotlin-as-java-plugin", version.r
...
 kotlin-stdlib-jdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
-kotlinx-metadata-jvm = "org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.5.0"
+kotlinx-metadata-jvm = "org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.6.0"
 
 moshi-core = { module = "com.squareup.moshi:moshi", version.ref = "moshi" }
 moshi-kotlin = { module = "com.squareup.moshi:moshi-kotlin", version.ref = "moshi" }
 moshi-adapters = { module = "com.squareup.moshi:moshi-adapters", version.ref = "moshi" }
 
 moshix-ksp = { module = "dev.zacsweers.moshix:moshi-ksp", version.ref = "moshix" }
-moshix-sealed-codegen = { module = "dev.zacsweers.moshix:moshi-sealed-codegen", version.ref = "moshix" }
```
autonomousapps, line 31: is codegen unused?

gabrielittner:
Interesting, looks like the AGP 7.4 specific expected advice that I added in #847 is now not needed anymore.

gradle/libs.versions.toml
gabrielittner: Yes, none of the json classes has `generateAdapter = true` in the annotation

src/functionalTest/groovy/com/autonomousapps/AbstractFunctionalSpec.groovy
```diff
@@ -24,9 +22,7 @@ abstract class AbstractFunctionalSpec extends Specification {
   // For faster CI times, we only test min + max. Testing all would be preferable, but we don't have till the heat death
   // of the universe to wait.
   protected static final SUPPORTED_GRADLE_VERSIONS = [
-    GRADLE_7_0,
-//    GRADLE_7_1,
-//    GRADLE_7_2,
+    GRADLE_7_2,
```
autonomousapps, line 25: Note to self: update wiki to note that the new minimum required version of Gradle is now 7.2.

autonomousapps (approved):

autonomousapps:
Thanks!

PR #852, opened 2023-01-30
Analyze dependencies of all source sets in standard JVM projects

jjohannes:
This adds analyzing dependencies **FROM** other source sets (which are not `main` or `test`) in standard Gradle 'java-library' and 'application' projects.

For example: `testFixturesImplementation(...)` or `integrationTestImplementation(...)`

This partially addresses #298 and #451, in particular the first item I listed in this comment: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/298#issuecomment-1346159435

It is solved as follows:

- All source sets registered in the `SourceSetContainer` are marked as supported for 'java-library' and 'application' projects.
- For each source set, we check if it can have an ABI by checking if the `api...` configuration was created. This happens, if you define a Feature Variant on top of a source set.
- There is a new `SourceSetKind.CUSTOM_JVM` that is used for all source sets which are not 'main' or 'test'. In this case, the internal 'base variant' object is different (different name) depending on the source set's name.

At first, I considered having only one `SourceSetKind` for **all** JVM source sets (including 'main' and 'test'). But there is still the special behavior that **main** dependencies are automatically seen by **test**.

I think it is debatable, if this is really the right behavior. I.e., if I use something directly in my tests, why not also have a 'testImplementation' dependency in addition to 'implementation'. But we probably don't want to change the current behavior. It could be reconsidered in the future though and then I think all source sets in JVM projects could be treated equally (have the same SourceSetKind).

For custom source sets, I think the 'stricter' behavior is right, as they usually can't automatically see dependencies of other source sets.

**NOTE**: This is only the first half of what is described in https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/298#issuecomment-1346159435. When I work on that, I want to add more test cases to cover more scenarios and combinations.

TODO:

- [ ]  More test coverage as part of #854

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IntegrationTestProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IntegrationTestProject.groovy
@@ -15,11 +15,11 @@ final class IntegrationTestProject extends AbstractProject {
 
   final GradleProject gradleProject
 
-  IntegrationTestProject() {
-    this.gradleProject = build()
+  IntegrationTestProject(boolean withCorrectIntegrationTestDependencies) {
+    this.gradleProject = build(withCorrectIntegrationTestDependencies)
   }
 
-  private GradleProject build() {
+  private GradleProject build(boolean correctIntegrationTestDependencies) {
     def builder = newGradleProjectBuilder()
     builder.withSubproject('proj') { s ->
       s.sources = PROJ_SOURCES
@@ -28,7 +28,7 @@ final class IntegrationTestProject extends AbstractProject {
         bs.sourceSets = ['integrationTest']
         bs.dependencies = [
           project('implementation', ':lib'),
-          project('integrationTestImplementation', ':core')
+          project('integrationTestImplementation', correctIntegrationTestDependencies ? ':core' : ':lib')
         ]
       }
     }
@@ -123,11 +123,25 @@ final class IntegrationTestProject extends AbstractProject {
     return actualProjectAdvice(gradleProject)
   }
 
-  private static Set<Advice> addCore = [Advice.ofAdd(projectCoordinates(':core'), 'api')]
-
-  final Set<ProjectAdvice> expectedBuildHealth = [
-    emptyProjectAdviceFor(':lib'),
-    emptyProjectAdviceFor(':core'),
-    projectAdviceForDependencies(':proj', addCore)
-  ]
+  static Set<ProjectAdvice> expectedBuildHealth(boolean correctIntegrationTestDependencies) {
+    if (correctIntegrationTestDependencies) {
+      [
+        emptyProjectAdviceFor(':lib'),
+        emptyProjectAdviceFor(':core'),
+        projectAdviceForDependencies(':proj', [
+          Advice.ofAdd(projectCoordinates(':core'), 'api')
+        ] as Set<Advice>)
+      ]
+    } else {
+      [
+        emptyProjectAdviceFor(':lib'),
+        emptyProjectAdviceFor(':core'),
+        projectAdviceForDependencies(':proj', [
+          Advice.ofAdd(projectCoordinates(':core'), 'api'),
+          Advice.ofAdd(projectCoordinates(':core'), 'integrationTestImplementation'),
+          Advice.ofRemove(projectCoordinates(':lib'), 'integrationTestImplementation')
+        ] as Set<Advice>)
+      ]
+    }
+}
 }
--- a/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
@@ -104,7 +104,7 @@ internal abstract class JvmAnalyzer(
   }
 }
 
-internal class JavaAppAnalyzer(
+internal class JavaWithoutAbiAnalyzer(
   project: Project,
   sourceSet: SourceSet,
   kind: SourceSetKind
@@ -114,7 +114,7 @@ internal class JavaAppAnalyzer(
   hasAbi = false
 )
 
-internal class JavaLibAnalyzer(
+internal class JavaWithAbiAnalyzer(
   project: Project,
   sourceSet: SourceSet,
   kind: SourceSetKind,
--- a/src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
+++ b/src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
@@ -40,7 +40,7 @@ internal object Configurations {
    * For example: Test Fixtures or additional Feature Variants.
    */
   @OptIn(ExperimentalStdlibApi::class)
-  internal fun variantFrom(configurationName: String, supportedSourceSets: Set<String>): Variant? {
+  internal fun variantFrom(configurationName: String, supportedSourceSets: Set<String>, hasCustomSourceSets: Boolean): Variant? {
     val mainBucket = MAIN_SUFFIXES.find { configurationName.endsWith(suffix = it, ignoreCase = true) }
     val candidate = if (mainBucket != null) {
       val variantSlug = if (configurationName == mainBucket) {
@@ -51,7 +51,7 @@ internal object Configurations {
         configurationName.removeSuffix(mainBucket.replaceFirstChar(Char::uppercase))
       }
 
-      findVariant(variantSlug, supportedSourceSets)
+      findVariant(variantSlug, supportedSourceSets, hasCustomSourceSets)
     } else {
       val procBucket = ANNOTATION_PROCESSOR_TEMPLATES.find { it.matches(configurationName) }
       if (procBucket != null) {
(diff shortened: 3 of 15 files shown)


src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
```diff
@@ -79,13 +79,16 @@ internal object Configurations {
   }
 
   @OptIn(ExperimentalStdlibApi::class)
-  private fun findVariant(variantSlug: String, supportedSourceSets: Set<String>): Variant? {
+  private fun findVariant(variantSlug: String, supportedSourceSets: Set<String>, hasCustomSourceSets: Boolean): Variant? {
```
jjohannes, line 82: I added the `hasCustomSourceSets` flag here so that we know if we need to identify "android test source sets" or not. Otherwise a custom source set like "testIntegration" in a standard JVM project might be classified as Android Test.

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
```diff
@@ -159,7 +174,7 @@ internal class StandardTransform(
             usageIter.remove()
 
             // Don't change a single-usage match, it's correct!
-            if ((!(singleVariant && usage.bucket.matches(theDecl)))) {
+            if (!(singleVariant && usage.bucket.matches(theDecl))) {
```
jjohannes, line 177: Remove redundant brackets

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IntegrationTestProject.groovy
```diff
@@ -15,11 +15,11 @@ final class IntegrationTestProject extends AbstractProject {
 
   final GradleProject gradleProject
 
-  IntegrationTestProject() {
-    this.gradleProject = build()
+  IntegrationTestProject(boolean correctIntegrationTestDependencies) {
```
autonomousapps, line 18: is "correct" being used as a verb or an adjective here?

src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
```diff
@@ -104,7 +104,7 @@ internal abstract class JvmAnalyzer(
   }
 }
 
-internal class JavaAppAnalyzer(
+internal class JavaWithoutAbiAnalyzer(
```
autonomousapps, line 107: can application projects reasonably have an `api`?

src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
```diff
@@ -79,13 +79,16 @@ internal object Configurations {
...
     if (variantSlug.isNotEmpty() && !supportedSourceSets.contains(variantSlug)) return null
-
-    return if (variantSlug == Variant.TEST_NAME) {
+    return if (variantSlug.isEmpty()) {
```
autonomousapps, line 84: why `variantSlug.isEmpty()`?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -936,15 +912,8 @@ internal class ProjectPlugin(private val project: Project) {
...
-
-    val main: SourceSet? = sourceSets.getByName(SourceSet.MAIN_SOURCE_SET_NAME)
-
-    val test: SourceSet? = if (project.shouldAnalyzeTests()) {
-      sourceSets.getByName(SourceSet.TEST_SOURCE_SET_NAME)
-    } else {
-      null
-    }
+    val sourceSets = project.the<SourceSetContainer>().matching {
+      s -> project.shouldAnalyzeTests() || s.name != SourceSet.TEST_SOURCE_SET_NAME}
```
autonomousapps, line 916: nit: `}` on new line

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
```diff
@@ -64,6 +69,14 @@ internal class StandardTransform(
     androidTestUsages = if (isMainVisibleDownstream) mutableSetOf() else reduceUsages(androidTestUsages)
     computeAdvice(advice, androidTestUsages, androidTestDeclarations, androidTestUsages.size == 1)
 
+
+    /*
+     * Custom JVM source sets like 'testFixtures', 'integrationTest' or other custom source sets and feature variants
+     */
+
+    customJvmUsage = reduceUsages(customJvmUsage)
+    computeAdvice(advice, customJvmUsage, customJvmDeclarations, customJvmUsage.size == 1)
```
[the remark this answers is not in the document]
autonomousapps: If you were to change its value, I would expect that some tests in `StandardTransformTest` would fail.

autonomousapps (approved):

autonomousapps:
> I think it is debatable, if this is really the right behavior. I.e., if I use something directly in my tests, why not also have a 'testImplementation' dependency in addition to 'implementation'. But we probably don't want to change the current behavior. It could be reconsidered in the future though and then I think all source sets in JVM projects could be treated equally (have the same SourceSetKind).
> 
> For custom source sets, I think the 'stricter' behavior is right, as they usually can't automatically see dependencies of other source sets.

I agree with both these sentiments, that we should maintain current behavior with main/test, and be stricter with custom source sets.

autonomousapps:
Once you fix up the (pretty simple) merge conflicts, I'm happy to see this merged.

src/main/kotlin/com/autonomousapps/internal/analyzer/JvmProjectAnalyzer.kt
jjohannes: No. But source sets inside a library project (like integrationTest) also do not have an API. That's why I changed the name as it is used for **all** source sets without API now.

jjohannes: Well, maybe an application can have an `api` if the user decides so. E.g. you have something that can run, but you still want to use it somewhere else as a library. You could configure it like this.

I wonder if we really still need to distinguish between `application` and `library` cases or if checking it the `...Api` config exists is enough to make a distinction _per source set_.

jjohannes: One last comment 😄 

This PR adds the `val hasAbi = configurations.findByName(sourceSet.apiConfigurationName) != null` check to the "application" case as it is now (only not in the SPRING_BOOT_PLUGIN special case).

I think it is correct to do so, as I wrote in the second comment.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IntegrationTestProject.groovy
jjohannes: _adjective_  change to `withCorrectIntegrationTestDependencies`

src/main/kotlin/com/autonomousapps/model/declaration/Configurations.kt
jjohannes: The empty string represents the `main` variant (I added a comment now). Or is that not always the case? It is for standard Java projects (because everything with a "variantSlug" is either "test" or "custom").

I moved this check for the "main" variant first in this list of checks, because in the "custom source set case" it will default to that further down.

There is still the "else" in the end that also falls back to `main`, but I think that's a case we should never reach (unless this is something in Android I am not aware of).

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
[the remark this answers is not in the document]
jjohannes: Thanks for the pointer. If I understand correctly, this is related to the fact the the Android variants inherit from each other (e.g. the relationship between `implementation`, `releaseImplementation`, `debugImplementation`)

Then this should not concern the _standard_ JVM case we care about here.

In any case I noticed that there is inconsistency between "main usages" and the other usages types.

In "main usages" the `singleVariant` is computed for the original list, while for the "test usages" it is computed on the reduced list. I assume this might actually be wrong the test cases in Android (?).

I added a commit on top in the PR to make this consistent - 96cc3fe24812d465d78d5787d2e56f95e2919d2f. We can remove that gain if it is wrong (or move it to a separate PR).

jjohannes:
👍  Merge conflicts are fixed. If this passes tests and there are no concern in the open discussions, I am happy to see this merged and I will add more tests that also cover this as part of #854.

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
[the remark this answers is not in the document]
jjohannes: Ok that change fails tests related to "test" dependencies - https://scans.gradle.com/s/4irf2zne7vuhw/tests/overview?outcome=failed

In this case, I leave it as it is.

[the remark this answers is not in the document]
autonomousapps: Yeah, we can discuss that in a separate PR if you want. That whole very messy algorithm was developed with TDD. Every decision was made in order to get a test to pass.

autonomousapps (approved):

autonomousapps:
Thanks for squashing and providing a reasonable commit message!

Issue #342, opened 2020-12-02
Joda time and android joda time


ColtonIdle:
**Plugin version**
0.68.0

**Gradle version**
6.7

**Android Gradle Plugin (AGP) version**
4.1.1

**Describe the bug**
Added a dependency of `implementation "net.danlew:android.joda:2.10.7.2"` and then running this plugin recommended to add `implementation "joda-time:joda-time:2.10.7"`

After adding it, my project doesn't build due to this issue:

```
Duplicate class org.joda.time.Chronology found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateMidnight found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateMidnight$Property found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateTime found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateTime$Property found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateTimeComparator found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
Duplicate class org.joda.time.DateTimeConstants found in modules jetified-joda-time-2.10-no-tzdb.7-no-tzdb (joda-time:joda-time:2.10.7) and jetified-joda-time-2.10.7 (joda-time:joda-time:2.10.7)
[...]

PR #854, opened 2023-01-31
Support dependencies to other Feature Variants (including Test Fixtures)

### LINKED ISSUE #298: Plugin incorrectly(?) reporting unused dependency

### LINKED ISSUE #836: Incorrect advice when using a dependency with a capability
**Build scan link**
Private :/

**Plugin version**
`1.17.1-SNAPSHOT`

**Gradle version**
`7.4.2`

Not applicable.

```
dependencies {
    implementation(project(":bar")) {
        capabilities {
            requireCapability("com.foo:bar-<capability>")
        }
    }
}
```

```
> Transitively used dependencies that should be declared directly as indicated:
    implementation project(':bar')
```

**Describe the bug**
When computing the advice in `StandardTransform#computeAdvice` the `usages` are compared to the `declarations` and based on what is incorrect some advice is given. However the set of `declarations` have filtering applied to `special dependencies like test fixtures or platforms`, but this filtering does not happen for the `usages`. This result in incorrect advice where a dependency on a 'special dependency' is used and declared, the declaration is filtered out when computing the advice and then the plugin advises to declare the dependency. Unfortunately the advice prints the advice also incorrectly as the capability of the dependency is not taken into account.

**To Reproduce**
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/commit/35ebb6bb20d1aab8e8bbae6eacec9ea4f57e5b0d

**Expected behavior**
Either
* Ignore the usage of the 'special' dependency with a capability
* Include the capability in the dependency
```
> Transitively used dependencies that should be declared directly as indicated:
    implementation(project(':bar')) { capabilities { requireCapability("com.foo:bar-<capability>") }
```

**Additional context**
This relates to #298, but it's slightly different as 'special' dependencies are not supported and should be filtered out.

I've managed to trace this back to `com.autonomousapps.internal.utils.GradleStringsKt#toCoordinates(org.gradle.api.artifacts.result.ResolvedDependencyResult)` where the `ResolvedDependencyResult` is converted to the internal model `Coordinates`. The `ResolvedDependencyResult` might also have `requested.requestedCapabilities`, but this information is not saved to `Coordinates`. Something like the following could be added to determine whether a capability needs to be saved to the `Coordinates`. 
```
val requestedCapability = requested.requestedCapabilities.firstOrNull()
  val selectedCapability = selected.variants.firstOrNull()?.capabilities?.firstOrNull()
  if (requestedCapability == selectedCapability) {
    capability = selectedCapability?.name
  }
```

### LINKED ISSUE #841: False positive for testImplementation dependency with classifier
**Build scan link**
N/A

**Plugin version**
1.18.0

**Gradle version**
7.6

N/A

```
> Task :E84HSP-app:reason

You asked about the dependency 'se.micronic.bennu:kernel:6.2.0.0-46-16c1002'.
You have been advised to remove this dependency from 'testImplementation'.

Shortest path from :E84HSP-app to se.micronic.bennu:kernel:6.2.0.0-46-16c1002 for compileClasspath:
:E84HSP-app
\--- se.micronic.bennu:kernel:6.2.0.0-46-16c1002

Shortest path from :E84HSP-app to se.micronic.bennu:kernel:6.2.0.0-46-16c1002 for runtimeClasspath:
:E84HSP-app
\--- se.micronic.bennu:kernel:6.2.0.0-46-16c1002

Shortest path from :E84HSP-app to se.micronic.bennu:kernel:6.2.0.0-46-16c1002 for testCompileClasspath:
:E84HSP-app
\--- se.micronic.bennu:kernel:6.2.0.0-46-16c1002

Shortest path from :E84HSP-app to se.micronic.bennu:kernel:6.2.0.0-46-16c1002 for testRuntimeClasspath:
:E84HSP-app
\--- se.micronic.bennu:kernel:6.2.0.0-46-16c1002

Source: main
------------
* Exposes 27 classes, 5 of which are shown: se.micronic.bennu.kernel.gui.DisplayConditionsIfc, se.micronic.bennu.kernel.logevent.LogEventDeclaration, se.micronic.bennu.kernel.logevent.LogEventId, se.micronic.bennu.kernel.logevent.XMLLogEventSpecificationBase, se.micronic.bennu.kernel.message.Command (implies api).
* Imports 3 constants: se.micronic.bennu.kernel.message.*, se.micronic.bennu.kernel.message.CommandDeclaration.*, se.micronic.bennu.kernel.variable.* (implies implementation).

Source: test
\------------
\* Uses 1 class: se.micronic.bennu.kernel.message.threaded.ThreadedCommandTestExtension (implies testImplementation).

```
**Describe the bug**
I have this dependency, (please note the classifier = 'test'), without it my project do not compile.
```
  testImplementation("se.micronic.bennu:kernel:6.2.0.0-46-16c1002") { artifact { classifier = 'test' } }
```

So, the test code of my project depends on the test code of another project which is selected using the 'classifier = test', however the dependency analysis seems to ignore the classifier.

**To Reproduce**
N/A

**Expected behavior**
Should not give advise to remove the dependency since it is used for compilation of test code.

**Additional context**
N/A

jjohannes:
The main change in this PR is that it adds something I called `GradleVariantIdentification` to all `Coordinates`.

- On the declaration side (consumer) The `GradleVariantIdentification` collects all [capabilities](https://docs.gradle.org/current/userguide/feature_variants.html#sec::consuming_feature_variants) and [attributes](https://docs.gradle.org/current/userguide/variant_attributes.html) that are defined on a dependency. Shortcut notations like `platform(...)` and `testFixtures(...)` also add these under the hood.
- On the resolved graph side (producer) the `GradleVariantIdentification` contains all the _capabilities_ of the selected node. The attributes are ignored here as we do not need them (see also comments in code).

With this, the analysis plugin now recognizes the fact that it is possible to have multiple nodes with the same `identifier` but different `GradleVariantIdentification` in the graph. This is the case when the capabilities differ. Because it is allowed to have multiple variants of one component in the graph as long as they have different capabilities. For example, there can be two `org.example:foo` nodes: one with capability `org.example:foo` and one with capability `org.example:foo-test-fixtures`.

The most critical impact on the implementation is IMO that there are several places throughout the code base that assume/assumed that Coordinates are comparable for different reasons. We need to be careful now, as different coordinates can mean the same. For example, if I don't declare any capability, the "main" capability is automatically used. The "Declaration Coordinates" then do not have a capability, but the "Resolved Graph Coordinates" will have the "main" capability. 

Therefore, several places needed to be touched, but I am confident wit the implementation now. I tried several things before I ended up with this solution (can still be seen in the commit history). What we have now feels most sound and also future-proof. We now collect all information we need to identify a component+variant and then can use this later during advice computation to make decisions. This should cover all "Feature Variant" standard cases. But if we need more information in the future, we can extend what we put into `GradleVariantIdentification`. (For example, it could also contain 'classifier' information – #342 – if we could somehow get this for 'both sides'.)

I chose the name `GradleVariantIdentification` as the _main_ thing, because _capability_ already has a different meaning in the context of this plugin. But I am happy to rename that if there are better ideas. 😄 

The PR resolves and adds test coverage for:
- Resolves #298 🎉 
- Fixes #836 
- Gives a alternative solution for #859 (will comment there)
- Fixes #841 in so far that there are no more false positives (we still won't be able to suggest classifiers correctly, but that is #342) will comment there

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -2,6 +2,7 @@ package com.autonomousapps.internal.advice
 
 import com.autonomousapps.model.Advice
 import com.autonomousapps.model.Coordinates
+import com.autonomousapps.model.ModuleCoordinates
 import com.autonomousapps.model.ProjectCoordinates
 
 internal class AdvicePrinter(
@@ -10,21 +11,41 @@ internal class AdvicePrinter(
   private val dependencyMap: (String) -> String = { it },
 ) {
 
-  fun line(configuration: String, printableIdentifier: String, was: String = ""): String = when (dslKind) {
-    DslKind.KOTLIN -> "  $configuration($printableIdentifier)$was"
-    DslKind.GROOVY -> "  $configuration $printableIdentifier$was"
-  }
+  fun line(configuration: String, printableIdentifier: String, was: String = ""): String =
+    "  $configuration$printableIdentifier$was"
 
-  fun toDeclaration(advice: Advice): String = when (dslKind) {
-    DslKind.KOTLIN -> "  ${advice.toConfiguration}(${gav(advice.coordinates)})"
-    DslKind.GROOVY -> "  ${advice.toConfiguration} ${gav(advice.coordinates)}"
-  }
+  fun toDeclaration(advice: Advice): String =
+    "  ${advice.toConfiguration}${gav(advice.coordinates)}"
 
   fun gav(coordinates: Coordinates): String {
     val quotedDep = coordinates.mapped()
     return when (coordinates) {
       is ProjectCoordinates -> if (dslKind == DslKind.KOTLIN) "project(${quotedDep})" else "project(${quotedDep})"
       else -> if (dslKind == DslKind.KOTLIN) quotedDep else quotedDep
+    }.let {
+      if (coordinates.featureVariantName.isEmpty()) {
+        when (dslKind) {
+          DslKind.KOTLIN -> "($it)"
+          DslKind.GROOVY -> " $it"
+        }
+      } else if (coordinates.featureVariantName == "test-fixtures") {
+        when (dslKind) {
+          DslKind.KOTLIN -> "(testFixtures($it))"
+          DslKind.GROOVY -> " testFixtures($it)"
+        }
+      } else {
+        val capability = when (coordinates) {
+          // TODO could probably pass the group around in ProjectCoordinates
+          is ProjectCoordinates -> "[group]${coordinates.identifier}-${coordinates.featureVariantName}"
+          is ModuleCoordinates -> "${coordinates.identifier}-${coordinates.featureVariantName}"
+          else -> "[group]:${coordinates.identifier}-${coordinates.featureVariantName}"
+        }
+        val quote = when (dslKind) {
+          DslKind.KOTLIN -> "\""
+          DslKind.GROOVY -> "'"
+        }
+        "($it) { capabilities { requireCapabilities($quote$capability$quote) } }"
+      }
     }
   }
 
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -27,7 +27,7 @@ internal class AdvicePrinter(
           DslKind.KOTLIN -> "($id)"
           DslKind.GROOVY -> " $id"
         }
-      } else if (coordinates.gradleVariantIdentification.capabilities.any { it.endsWith(":test-fixtures") }) {
+      } else if (coordinates.gradleVariantIdentification.capabilities.any { it.endsWith("-test-fixtures") }) {
         when (dslKind) {
           DslKind.KOTLIN -> "(testFixtures($id))"
           DslKind.GROOVY -> " testFixtures($id)"
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -23,28 +23,23 @@ internal class AdvicePrinter(
       is ProjectCoordinates -> if (dslKind == DslKind.KOTLIN) "project(${quotedDep})" else "project(${quotedDep})"
       else -> if (dslKind == DslKind.KOTLIN) quotedDep else quotedDep
     }.let {
-      if (coordinates.featureVariantName.isEmpty()) {
+      // Used 'endsWith' instead of '==' as project coordinates do not include their group right now
+      if (coordinates.capability.endsWith(coordinates.identifier)) {
         when (dslKind) {
           DslKind.KOTLIN -> "($it)"
           DslKind.GROOVY -> " $it"
         }
-      } else if (coordinates.featureVariantName == "test-fixtures") {
+      } else if (coordinates.capability.endsWith(":test-fixtures")) {
         when (dslKind) {
           DslKind.KOTLIN -> "(testFixtures($it))"
           DslKind.GROOVY -> " testFixtures($it)"
         }
       } else {
-        val capability = when (coordinates) {
-          // TODO could probably pass the group around in ProjectCoordinates
-          is ProjectCoordinates -> "[group]${coordinates.identifier}-${coordinates.featureVariantName}"
-          is ModuleCoordinates -> "${coordinates.identifier}-${coordinates.featureVariantName}"
-          else -> "[group]:${coordinates.identifier}-${coordinates.featureVariantName}"
-        }
         val quote = when (dslKind) {
           DslKind.KOTLIN -> "\""
           DslKind.GROOVY -> "'"
         }
-        "($it) { capabilities { requireCapabilities($quote$capability$quote) } }"
+        "($it) { capabilities { requireCapabilities($quote${coordinates.capability}$quote) } }"
       }
     }
   }
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -2,7 +2,6 @@ package com.autonomousapps.internal.advice
 
 import com.autonomousapps.model.Advice
 import com.autonomousapps.model.Coordinates
-import com.autonomousapps.model.ModuleCoordinates
 import com.autonomousapps.model.ProjectCoordinates
 
 internal class AdvicePrinter(
@@ -22,27 +21,28 @@ internal class AdvicePrinter(
     return when (coordinates) {
       is ProjectCoordinates -> if (dslKind == DslKind.KOTLIN) "project(${quotedDep})" else "project(${quotedDep})"
       else -> if (dslKind == DslKind.KOTLIN) quotedDep else quotedDep
-    }.let {
-      // Used 'endsWith' instead of '==' as project coordinates do not include their group right now
-      if (coordinates.capability.endsWith(coordinates.identifier)) {
+    }.let { id ->
+      if (coordinates.gradleVariantIdentification.capabilities.isEmpty()) {
         when (dslKind) {
-          DslKind.KOTLIN -> "($it)"
-          DslKind.GROOVY -> " $it"
+          DslKind.KOTLIN -> "($id)"
+          DslKind.GROOVY -> " $id"
         }
-      } else if (coordinates.capability.endsWith(":test-fixtures")) {
+      } else if (coordinates.gradleVariantIdentification.capabilities.any { it.endsWith(":test-fixtures") }) {
         when (dslKind) {
-          DslKind.KOTLIN -> "(testFixtures($it))"
-          DslKind.GROOVY -> " testFixtures($it)"
+          DslKind.KOTLIN -> "(testFixtures($id))"
+          DslKind.GROOVY -> " testFixtures($id)"
         }
       } else {
         val quote = when (dslKind) {
           DslKind.KOTLIN -> "\""
           DslKind.GROOVY -> "'"
         }
-        "($it) { capabilities { requireCapabilities($quote${coordinates.capability}$quote) } }"
+        "($id) { capabilities {\n${coordinates.gradleVariantIdentification.capabilities.filter { !it.endsWith(":test-fixtures") }.joinToString("") { it.requireCapability(quote) }}  }}"
       }
     }
   }
+
+  private fun String.requireCapability(quote: String) = "    requireCapability($quote$this$quote)\n"
 
   private fun Coordinates.mapped(): String {
     val gav = gav()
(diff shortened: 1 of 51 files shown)


autonomousapps:
I'd prefer to wait on reviewing this till the other PR gets merged and this rebased.

src/main/kotlin/com/autonomousapps/tasks/SynthesizeDependenciesTask.kt
```diff
@@ -205,7 +205,16 @@ abstract class SynthesizeDependenciesTask @Inject constructor(
...
+      other.capabilities.forEach { otherCapability ->
+        val existing = capabilities.find { it.javaClass.canonicalName == otherCapability.javaClass.canonicalName }
+        if (existing != null) {
+          val merged = existing.merge(otherCapability)
+          capabilities.remove(existing)
+          capabilities.add(merged)
+        } else {
+          capabilities.add(otherCapability)
+        }
+      }
```
jjohannes, line 217: This + the `Capability.merge()` method is the fix for #841

src/main/kotlin/com/autonomousapps/model/Coordinates.kt
```diff
@@ -5,59 +5,111 @@ import dev.zacsweers.moshix.sealed.annotations.TypeLabel
...
+   * let to selecting that node. If we could find that declaration, we could use it for the ADD advice.
+   * Right now, we use the details from the node in the Graph which may contain more capabilities as you need
+   * to declare. In particular, it also contains the 'default capability', which makes it conceptually equal to
+   * Coordinates without capability.
+   * In order to correctly reduce advices (e.g. merge a REMOVE and an ADD to a CHANGE), we need the same Coordinates
+   * on both. So this method should be used to 'minify' the GradleVariantIdentification for ADD advices.
+   *
+   * @return A copy of this Coordinates without the 'default capability'
+   */
+  fun withoutDefaultCapability(): Coordinates {
```
jjohannes, line 55: With all the different solutions I considered, I always tried to reduce assumptions and guesswork on top of the information we get out of Gradle. This method is basically what is left and for which I don't know if and how it could be further improved.
This is what the user gets in case of an ADD advice where we do not have an existing declaration to "reuse". I think this is a good start and if users report cases with unexpected advices, this can be tuned further if we have real-world reproducers.

jjohannes:
I did some more manual testing and there still seems to be an issue when the **same** dependency is declared **from** different source sets in the same project. I'll see that I provide tests that reproduce this (+ a fix) soon.

jjohannes:
With this (actually with #852 already 😬) users might suddenly get more in their reports than before as the plugin is now doing more analysis as before by default. I wonder if there should at least be an opt-out. There were two things I was thinking about. (Which could be done later. But maybe one of them before this is released...)

- Should the `dependencyAnalysis.issues.only... {}` mechanism be extended to allow selective ignoring per source set? Right now, it works for all source sets, but you can not say for example "do everything for main, but ignore for testFixtures".
- Should the `dependency.analysis.test.analysis=false` flag be generalized to something that's configured in the `dependencyAnalysis` extension and where you with which you can ignore arbitrary source sets (not only test).

TimvdLippe:
Commenting here as I originally asked Jendrik about this.

While trying to roll out the new changes, we were indeed hit by new warnings (which indeed is already true with #852). However, we weren't able to selectively ignore the warnings, other than ignoring all warnings. However, we don't want to lose coverage for `main` while we are cleaning up `test` (or any other sourceset). So any mechanism that allows for more selective filtering would be great. I have no strong preference on how that mechanism should work.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -238,43 +240,50 @@ internal class DependencyAdviceBuilder(
...
+      .mapNotNull { (advice, originalCoordinates) ->
+        // Make sure to do all operations here based on the originalCoordinates used in the graph.
+        // The 'advice.coordinates' may be reduced - e.g. contain less capabilities in the GradleVariantIdentifier.
         // TODO could improve performance by merging has... with find...
         when {
-          advice.isAdd() && bundles.hasParentInBundle(advice.coordinates) -> {
-            val parent = bundles.findParentInBundle(advice.coordinates)!!
-            bundledTraces += BundleTrace.DeclaredParent(parent = parent, child = advice.coordinates)
+          // The user cannot change this one:
+          // https://github.com/gradle/gradle/blob/d9303339298e6206182fd1f5c7e51f11e4bdff30/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaTestFixturesPlugin.java#L68
```
jjohannes, line 261: Not sure if there is a better place to but this exception. I think it's too in individual to solve this with 'bundles' (as it's done in other cases).
For reference: https://github.com/gradle/gradle/blob/d9303339298e6206182fd1f5c7e51f11e4bdff30/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaTestFixturesPlugin.java#L68

Probably was a mistake to hardcode this into the `test-fixtures-plugin` ... 😐

jjohannes:
Regarding the "opt-out". I prototyped this, that would turn off the whole analysis for a given _source set_. Either for _all_ or selected projects:

```
dependencyAnalysis {
    issues {
        all {
            ignoreSourceSet("mockApi")
            ignoreSourceSet("endToEndTest")
        }
    }
}
```

Can try it out from here: https://github.com/jjohannes/dependency-analysis-android-gradle-plugin/tree/ignore-selected-source-sets

If this is wanted, I can add a few tests and do follow up PR, once this one is integrated.

src/main/kotlin/com/autonomousapps/model/Coordinates.kt
```diff
@@ -5,59 +5,118 @@ import dev.zacsweers.moshix.sealed.annotations.TypeLabel
 
 @JsonClass(generateAdapter = false, generator = "sealed:type")
 sealed class Coordinates(
-  open val identifier: String
+  open val identifier: String,
+  open val gradleVariantIdentification: GradleVariantIdentification
```
jjohannes, line 9: Should we call the field just `gvi` (or smth completely different) to make this more compact in the Json files?

ljacomet:
> Regarding the "opt-out".

It would be great if the opt out for _all_ projects was lenient: If a source set is referenced but does not exist in a specific project, simply ignore.

src/main/kotlin/com/autonomousapps/tasks/SynthesizeProjectViewTask.kt
```diff
@@ -190,9 +190,11 @@ abstract class SynthesizeProjectViewTask @Inject constructor(
         .map { it.build() }
         .toSet()
 
-      val projectCoordinates = ProjectCoordinates(parameters.projectPath.get())
+      val projectCoordinates = ProjectCoordinates(parameters.projectPath.get(),
+        GradleVariantIdentification(emptySet(), emptyMap()))
+      val ignoreSelfDependencies = parameters.buildType.isPresent // ignore on Android
```
autonomousapps, line 195: clever

src/test/kotlin/com/autonomousapps/transform/StandardTransformTest.kt
autonomousapps, line 1: It doesn't look like you added any new tests here, is that correct?

autonomousapps (commented):
Thanks for this! And sorry it took so long to review. Left a few minor comments. Just want to hear your response before I approve.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -10,24 +10,40 @@ internal class AdvicePrinter(
   private val dependencyMap: (String) -> String = { it },
 ) {
 
-  fun line(configuration: String, printableIdentifier: String, was: String = ""): String = when (dslKind) {
-    DslKind.KOTLIN -> "  $configuration($printableIdentifier)$was"
-    DslKind.GROOVY -> "  $configuration $printableIdentifier$was"
-  }
+  fun line(configuration: String, printableIdentifier: String, was: String = ""): String =
+    "  $configuration$printableIdentifier$was"
```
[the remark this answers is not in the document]
jjohannes: Then `when` is now done in `fun gav(coordinates: Coordinates)` already (see changes below).

This is because we might need to add a `{ ... }` configuration block as part of the "printableIdentifier" for a `requireCapability`. Because of that, we need to make the decision if we want to add `(` and `)` there (as `)` might be in the middle of the string before the `{`).

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -10,24 +10,40 @@ internal class AdvicePrinter(
...
 
-  fun toDeclaration(advice: Advice): String = when (dslKind) {
```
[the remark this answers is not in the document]
jjohannes: See https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/854/files#r1158053173

src/main/kotlin/com/autonomousapps/model/Coordinates.kt
jjohannes: I don't have a strong opinion either. I also don't have a concrete case where anyone "complained". I just thought that the size of the files were a concern in the past already. I remember you thinking about alternative formats (but maybe I remember that wrongly).

Let's leave it like this for now then.

src/main/kotlin/com/autonomousapps/tasks/SynthesizeProjectViewTask.kt
jjohannes: I am actually not 100% sure why it needs to be like this on Android. But there were tests showing cases that needed this.

In "plain JVM" projects this cannot be ignored to support all cases that are shown in the tests this PR adds. So I think this distinction here is fine for now.

src/test/kotlin/com/autonomousapps/transform/StandardTransformTest.kt
jjohannes: I did not add a test. I adjusted the exiting tests as they now also have to respect the distinction between "Declaration Coordinates" and "Resolved Graph Coordinates". See this part of the PR description:

> The most critical impact on the implementation is IMO that there are several places throughout the code base that assume/assumed that Coordinates are comparable for different reasons. We need to be careful now, as different coordinates can mean the same. For example, if I don't declare any capability, the "main" capability is automatically used. The "Declaration Coordinates" then do not have a capability, but the "Resolved Graph Coordinates" will have the "main" capability.

jjohannes:
Thanks @autonomousapps! I responded to the comments. Let me know if there are any any concerns left before integrating this.

After this is merged, it would be great if you could also integrate the (small) follow-up in #882 before the next release.

autonomousapps (approved):
Thanks! Sorry for the approval delay. I was expecting a github notification that you had responded to my comments, but it never arrived 😕

PR #880, opened 2023-03-10
Allow ignoring the incompatibility warning.

TWiStErRob:
```
> Configure project :
The Dependency Analysis plugin is only known to work with versions of AGP between 4.2.2 and 7.4.1. You are using 7.4.2. Proceed at your own risk.
If you would like to suppress this message, add dependency.analysis.android.ignored.version=7.4.2 to you gradle.properties file.
```
and if you add
```properties
dependency.analysis.android.ignored.version=7.4.2
```
then it prints nothing. ☮

### DIFF
--- a/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/DependencyAnalysisPlugin.kt
@@ -2,6 +2,7 @@
 
 package com.autonomousapps
 
+import com.autonomousapps.Flags.androidIgnoredVersion
 import com.autonomousapps.internal.android.AgpVersion
 import com.autonomousapps.subplugin.ProjectPlugin
 import com.autonomousapps.subplugin.RootPlugin
@@ -39,11 +40,12 @@ class DependencyAnalysisPlugin : Plugin<Project> {
     }
 
     logger.debug("AgpVersion = $current")
-    if (!current.isSupported() && this == rootProject) {
+    if (!current.isSupported() && this == rootProject && androidIgnoredVersion() != current.version) {
       logger.warn(
         "The Dependency Analysis plugin is only known to work with versions of AGP between " +
-          "${AgpVersion.AGP_MIN.version} and ${AgpVersion.AGP_MAX.version}. You are using ${current.version}. " +
-          "Proceed at your own risk."
+          "${AgpVersion.AGP_MIN.version} and ${AgpVersion.AGP_MAX.version}. " +
+          "You are using ${current.version}. Proceed at your own risk. \n" +
+          "If you would like to suppress this message, add ${Flags.FLAG_ANDROID_IGNORED_PLUGIN_VERSION}=${current.version} to you gradle.properties file."
       )
     }
   }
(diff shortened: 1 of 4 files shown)


autonomousapps:
We resolved this in https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/920.

autonomousapps:
Sorry, I didn't really like the version-specific suppression. Kind of don't understand why people dislike these warnings so much :D

TWiStErRob:
> Kind of don't understand why people dislike these warnings so much :D

_Here is one explanation based on years of experience of upgrading tools, reviewing PRs; and seeing how early (ignored) signals turn into blocked upgrades, delayed releases or hotfixes. Sorry it got a bit long-ish as I think it's a complex issue._

Since the beginning of Gradle a working good build only outputs "BUILD SUCCESSFUL". This means that anything else output during build  (unless explicitly asked for with `--info` or properties, etc.) signals there's something wrong with the build. This could be compiler warnings, Gradle nagging, plugin issues, Spotless output, etc. Compiler warnings are coming from the project and can be trivially fixed, Gradle nagging comes from plugins and can be raised as an issue, and usually can be solved by bumping a plugin version. Having too many of these warnings hides real, important warnings like (paraphrasing) "your app will crash at runtime if you don't fix this". As an end user (app/service/product developer) one really only cares about compilation and static analysis warnings and errors, because they likely caused those and can fix them. As a platform developer, one cares about plugin warnings, Gradle nagging, etc. as they're early signals that the build will break in the future, potentially soon-ish (months) and they might need to take action now to get the dependencies ready.

Adding on top of this: for open source projects where you run `gradlew build` and get 2-5 pages of warnings, it's really hard to contribute, because you have no idea if you broke something, missing env setup, etc. And if you introduce/break something that shows a warning, you are way less likely to find it because the pre-existing warnings overwhelm the output. Contributors are more likely to ignore all warnings and just keep adding more and more. I endearingly call this "warning blindness". After one point people will (apparently) stop seeing warnings in CLI and in IDE too.

Another side-effect of having too many warnings, is that it could contribute gigabytes of CI logs per day. Somewhere I calculated this for real and it was mind-blowing. Of course gigabytes are cheap, but the time it takes to backup CI, or upload logs for finishing a build can actually affect developer experience, IT work, or emergency responses.

Now, bringing all this back to this specific warning: this warning bears informational value to the person who updated your plugin or AGP. The information has value until someone read it, and accepted the risk. Because there's nothing else you can really do than accept it, unless one's willing to downgrade AGP or contribute compatibility PR to the plugin, or remove the plugin entirely. After these actions, it's just one of those warnings in the "3rd page". If there's no way of suppressing it, then it'll contribute to the noise of build logs. This why it's great that your merged the other PR 🙏. We now have recourse to hide an unactionable warning that shows up on every build, even when we're not actively using this plugin.

Coming to my comment about the version-specificity: whoever accepted the risk of having the wrong version combination accepted it for the versions that was listed in the warning, maybe further patch versions. With the current implementation, the warning is gone forever, even if someone updates to later AGP 8.x or AGP 9 where things could really break (potentially silently). With my proposed solution, the targeted user (platform engineer) can make the choice of when they want to see this warning again, when it'll be relevant. In my initial implementation I just added support for suppressing one specific version. Later I realized that I wouldn't want it to come back for any patch, yet it would be really useful to get reminded when bigger changes are happening. Prefixes allowed this as described in the KDoc of the last commit.

I would be really curious why @ZacSweers, @Jacks0N23, @ChristianKatzmann, @dsvoronin also wanted it gone.

PR #882, opened 2023-03-14
Ignore selected source sets

jjohannes:
**Note: #854 should be merged first, then this PR only contains the commits for the additional feature**

As follow up to #854, this adds as _opt-out_ that turns off the whole analysis for a given source set. Works for all or selected projects:

```
dependencyAnalysis {
    issues {
        all {
            ignoreSourceSet("mockApi")
            ignoreSourceSet("endToEndTest")
        }
    }
}
```


autonomousapps:
Is https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/854#issuecomment-1443818373 from @ljacomet relevant here?

jjohannes:
@autonomousapps I rebased this PR.

> Is https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/854#issuecomment-1443818373 from @ljacomet relevant here?

Yes the solution implemented here is "lenient" as described in that comment. It's also in the test that uses `issues { all { .. } }` although not all subprojects have the "extraFeature" source set.

autonomousapps (approved):
LGTM! Could you also propose an update to the wiki documentation for this feature? Unfortunately that isn't under source control, so maybe open an issue or just slack me.

jjohannes:
@autonomousapps I can share a proposal for the wiki here directly.

I would just extend the listing at https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Customizing-plugin-behavior#failure-conditions-and-filtering as follows

```

dependencyAnalysis {
  issues {
    // configure for all projects
    all {
      // set behavior for all issue types
      onAny {
        severity(<'fail'|'warn'|'ignore'>) // default is 'warn'
        exclude('an:external-dep', 'another:external-dep', ':a:project-dep')
      }
      // or configure per-type
      onUnusedDependencies { ... }
      onUsedTransitiveDependencies { ... }
      onIncorrectConfiguration { ... }
      onCompileOnly { ... }
      onRuntimeOnly { ... }
      onUnusedAnnotationProcessors { ... }
      onRedundantPlugins { ... }
      // and/or ignore the analysis for a complete source set (JVM projects)
      ignoreSourceSet("testFixtures")
    }
    // or configure per project
    project(':lib') {
      ...
    }
  }
}

```

autonomousapps:
@jjohannes thanks! I just realized that `ignoreSourceSet` doesn't take a vararg argument, but only a single string. I think it might make sense to change that, although not a big deal.

jjohannes:
@autonomousapps good catch. It's not consistent with the existing DSL constructs. Here is a small follow-up: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/890

PR #883, opened 2023-03-30
Fix badge links to point to actual things

TWiStErRob:
The badges on the top of the readme point to the images which are shown, which is not useful for the users. This PR replaces the links to point to actually useful URLs, where more details can be found.


autonomousapps (approved):
Thanks!

PR #890, opened 2023-04-21
Make 'ignoreSourceSet(...)' DSL construct consistent with others

jjohannes:
...for example 'exclude(...)'.
It is now 'vararg' and fills a Set (instead of a List).

Follow up to: https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/882#issuecomment-1516575385


autonomousapps (approved):
Thanks!

Issue #853, opened 2023-01-30
AnnotationProcessor detection fails to collect all supported annotation types when multiple annotation processors exist in the same artifact/coordinate


marklundbergcrd:
**Build scan link**
n/a

**Plugin version**
1.18.0

**Gradle version**
7.7

**(Optional) Android Gradle Plugin (AGP) version**
n/a

**`reason` output for bugs relating to incorrect advice**
n/a

**Describe the bug**
When annotation processor dependencies are examined, a list of supported annotation types is found and assigned to an AnnotationProcessorDependency object for a specific coordinate.  When multiple annotation processors are available in the artifact found for a given coordinate, only the supported annotations types from the last discovered annotation processor are saved.  All other annotations processors are then marked as not needed during evaluation because their lists of supported annotation types were accidentally discarded.

**To Reproduce**
1. Create a jar that contains two annotation processors and lists them in src\main\resources\META-INF\services\javax.annotation.processing.Processor
2. Create a project that uses annotations from the first processor listed in the processor metadata but do not include annotations from the second processor.
3. projectHealth indicates that the annotation processor is unused.

**Expected behavior**
If any of the supported annotation types from any of the annotation processors included in a dependency are used, then the dependency is considered as 'used'.

**Additional context**
n/a

autonomousapps:
Thanks for the issue. It would help a lot if you provided a minimal reproducer project, or at the very least indicated an existing library that has multiple annotation processors.

PR #896, opened 2023-05-15
Add test to showcase included build annotation processor issue

TimvdLippe:
We are seeing issues with using annotation processor results from included builds when retrieving the information from the cache. I mirrored our setup in this test, by introducing two new projects:
1. A new included build called `processor-build` that defines an annotation processor and uses the Google AutoService to define it
2. A new submodule in this project, which uses the processor to process one of its fields

The problem is the following (although I couldn't fully define that in the test): when we run `:projectHealth` on this, it stores the information in our build cache. When we then attempt to retrieve it in a subsequent run, the `:projectHealth` task fails with the following error:

```
> Transitively used dependencies that should be declared directly as indicated:
    annotationProcessor project('my.custom.processor:sub-processor1')
```

It seems like the information retrieved from the cache is incomplete, where it somehow deduces that the annotation processor is unused. Again, if we run it without the cache, it is all fine. Only when the `:user-of-processor:findDeclaredProcsMain` is `FROM-CACHE`, it fails.

This only happens when using included builds. If the subprojects are part of the same build, we don't observe any caching problems.

Next to that, we also noticed that the `usages-annotation-processors.json` includes fully qualified project references. However, the cache key for this task does not take into account the fully qualified names of dependencies. In the sense that: when we run the task in 1 build (where the `rootProject.group` is set to `Group1`) and we run it again in another build where the group is set to `Group2`, the task would also fail, as it doesn't match the group.

We fixed this issue by aligning the group names, but I think that's a problem that can be fixed in the plugin as well. To do so, we can mark the `rootProject.group` as `inputs.property()` to these tasks. (Note that the `rootProject.group` is of the dependency, not the project that defines the plugin task, which makes it slightly more complicated).

I created this PR as a request of @jjohannes who wanted to see the example. I am not sure Jendrik if I did this correctly and if these test expectations are correct. I had trouble trying to formulate what the advice should be, as the test is now failing with an assertion error I don't understand. Hopefully it is clear enough with this setup.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
@@ -72,4 +72,19 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     where:
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
   }
+
+  def "can handle annotation processors from cache in subsequent builds"() {
+    given:
+    def project = new IncludedBuildWithSubprojectsProject(true)
+    gradleProject = project.gradleProject
+
+    when: 'build does not fail'
+    build(gradleVersion, gradleProject.rootDir, ':buildHealth')
+
+    then: 'and there is no advice'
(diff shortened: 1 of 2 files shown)


TimvdLippe:
I think #853 is related to this issue

TimvdLippe:
This was included in #916

PR #910, opened 2023-05-31
Change text to clarify missing entries

### LINKED ISSUE #909: Recommend small rewrite to error message
Replace "Transitively used dependencies that should be declared directly as indicated:"

Use text like "This project requires all transitive dependencies to be declared directly.  Add these to your build.gradle:"

**Additional context**
Basically be a bit more explicit about why there was a failure, and how to address.  I think this would help non-native speakers as well.

pforhan:
fixes #909

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilder.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilder.kt
@@ -48,8 +48,7 @@ internal class ProjectHealthConsoleReportBuilder(
 
       if (addAdvice.isNotEmpty()) {
         maybeAppendTwoLines()
-        appendReproducibleNewLine("This project requires all transitive dependencies to be declared directly. " +
-          "Add these to your build.gradle:")
+        appendReproducibleNewLine("These transitive dependencies should be declared directly:")
 
         val toPrint = addAdvice.mapToOrderedSet {
           line(it.toConfiguration!!, printableIdentifier(it.coordinates))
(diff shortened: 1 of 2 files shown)


ZacSweers (approved):

src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilder.kt
```diff
@@ -48,7 +48,8 @@ internal class ProjectHealthConsoleReportBuilder(
 
       if (addAdvice.isNotEmpty()) {
         maybeAppendTwoLines()
-        appendReproducibleNewLine("Transitively used dependencies that should be declared directly as indicated:")
+        appendReproducibleNewLine("This project requires all transitive dependencies to be declared directly. " +
+          "Add these to your build.gradle:")
```
autonomousapps, line 52: What about simply:
```
```

pforhan: I think that's a good step.  Should it include the path the the build file or is that obvious from context?

autonomousapps: No other advice includes a path to the build file, and in fact that path is technically dynamic so you shouldn't hardcode it. So, I'd rather not mention it at all. I will assume for now that people know where to declare their dependencies.

pforhan: updated

autonomousapps (approved):
LGTM. Will merge when tests pass.

PR #914, opened 2023-06-15
Use one InMemoryCache for all builds in a composite build tree

jjohannes:
When using composite builds, currently there is one cache per build in the build tree. So for example if one build is composed of 300 included builds, there will be 300 caches – which makes the memory consumption explode.

Heap dump analysis from such a build:

cache

### DIFF
--- a/src/main/kotlin/com/autonomousapps/services/InMemoryCache.kt
+++ b/src/main/kotlin/com/autonomousapps/services/InMemoryCache.kt
@@ -9,6 +9,7 @@ import com.autonomousapps.model.intermediates.ExplodingJar
 import com.github.benmanes.caffeine.cache.Cache
 import com.github.benmanes.caffeine.cache.Caffeine
 import org.gradle.api.Project
+import org.gradle.api.invocation.Gradle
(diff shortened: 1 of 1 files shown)


jjohannes:
With this change, the issue in the large composite build is gone. The analysis plugin works as it did before in a setup where all builds were combined into one. In the memory analysis, the _BuildServiceProvider_ is no longer dominating.

cache2

jjohannes:
@autonomousapps can this be integrated for the next release? I didn't see any functional issues with this and I tested it on the huge build mentioned above.

autonomousapps (approved):
LGTM.

autonomousapps:
> @autonomousapps can this be integrated for the next release? I didn't see any functional issues with this and I tested it on the huge build mentioned above.

Are you asking for a release soon, or can this wait for us to finish our planned set of issues?

jjohannes:
> Are you asking for a release soon, or can this wait for us to finish our planned set of issues?

I don't need this to be released urgently. It can wait and be released with other upcoming improvements/fixes.

PR #916, opened 2023-06-22
Consistently use IncludedBuildCoordinates to refer to other projects

jjohannes:
There is a bug that leads to wrong/inconsistent advices if included builds are involved. I first thought this is probably an easy fix, but then discovered a more fundamental problem with the _included builds_ handling and went down that rabbit hole.

The assumption at the moment is that if a 'project path' matches the 'identity path', the dependency is not in an included build.

This is in:
[`gradleStrings.ResolvedArtifactResult.toCoordinates()`](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt#L66)

While this works in the case where we do not use included builds (much), it is generally wrong. There can also be references to other projects inside other included builds. And, more problematic, an included build can refer back to the "root" build. The plugin then assumes that a a project in the root build could be addressed via 'project(":path")' notation from that other build, which is not true. So depending on where you start the build (which of your builds is the "root") the plugin may yield different advices.

This PR proposed to use `IncludedBuildCoordinates` consistently throughout the analysis. Then we also do not need to handle both in the middle of the advice calculation. Then only in the end when an advice is given, we decide which representation of the coordinates is shown to the user. For 'add' advices, we need to track the build ID (build name), to decide if a project is in the same build and can hence be addressed by project coordinates.

Maybe #896 is related. It seems to be in the same area – issues with build/project identity in a composite.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
@@ -19,12 +19,31 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     gradleProject = project.gradleProject
 
     when: 'build does not fail'
-    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+    build(gradleVersion, gradleProject.rootDir, ':buildHealth')
+    build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
 
-    then: 'and there is no advice'
-    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+    then: 'and there is the expected advice'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth('second-build'))
+    assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild(':'))
 
     where: 'This new feature only works for Gradle 7.3+'
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "result of analysis does not change if root build of included build tree changes (only root projects)"() {
+    given:
+    def project = new IncludedBuildProject()
+    gradleProject = project.gradleProject
+
+    when: 'build running from second build root does not fail'
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':the-project:buildHealth')
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
+
+    then: 'and there is the expected advice'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth(':'))
+    assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild('the-project'))
+
+    where:
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
   }
 
@@ -52,7 +71,7 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
 
     then: 'and there is no advice'
-    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth)
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth('second-build'))
 
     where:
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
@@ -63,11 +82,26 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     def project = new IncludedBuildWithSubprojectsProject(true)
     gradleProject = project.gradleProject
 
-    when: 'build does not fail'
+    when:
     build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
 
-    then: 'and there is no advice'
-    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth)
+    then:
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth('second-build'))
+
+    where:
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "result of analysis does not change if root build of included build tree changes (with subprojects)"() {
+    given:
+    def project = new IncludedBuildWithSubprojectsProject(true)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
+
+    then:
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth(':'))
 
     where:
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
@@ -1,6 +1,7 @@
 package com.autonomousapps.jvm
 
 import com.autonomousapps.jvm.projects.IncludedBuildProject
+import com.autonomousapps.jvm.projects.IncludedBuildWithAnnotationProcessorProject
 import com.autonomousapps.jvm.projects.IncludedBuildWithSubprojectsProject
 
 import static com.autonomousapps.utils.Runner.build
@@ -102,6 +103,53 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
 
     then:
     assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth(':'))
+
+    where:
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "up-to-date check is correct when switching root builds"() {
+    shouldClean = false
+
+    given:
+    def project = new IncludedBuildWithSubprojectsProject(true)
+    gradleProject = project.gradleProject
+
+    when: 'The first build is the root'
+    build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
+
+    then:
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth('second-build'))
+
+    when: 'The second build is the root - the "buildName" attribute of ProjectCoordinates changes'
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
+
+    then:
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth(':'))
+
+    where:
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "can handle annotation processors from cache in subsequent builds"() {
+    given:
+    def project = new IncludedBuildWithAnnotationProcessorProject()
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, '--build-cache', ':buildHealth')
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    when: 'Running again - UP-TO-DATE'
+    build(gradleVersion, gradleProject.rootDir, '--build-cache', ':buildHealth')
+    then: 'Result is the same'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    when: 'Running again - FROM-CACHE'
+    build(gradleVersion, gradleProject.rootDir, '--build-cache', 'clean', ':buildHealth')
+    then: 'Result is the same'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
 
     where:
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
@@ -123,7 +123,7 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     then:
     assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth('second-build'))
 
-    when: 'The second build is the root - the "buildName" attribute of ProjectCoordinates changes'
+    when: 'The second build is the root - the "buildPath" attribute of ProjectCoordinates changes'
     build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
 
     then:
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
@@ -23,15 +23,17 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     build(gradleVersion, gradleProject.rootDir, ':buildHealth')
     build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
 
-    then: 'and there is the expected advice'
+    then: 'the build health of the first build is as expected'
     assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth('second-build'))
+
+    and: 'the build health of the second build is as expected'
     assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild(':'))
 
     where: 'This new feature only works for Gradle 7.3+'
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
   }
 
-  def "result of analysis does not change if root build of included build tree changes (only root projects)"() {
+  def "result of analysis does not change if root build of included build tree changes for projects without subprojects"() {
     given:
     def project = new IncludedBuildProject()
     gradleProject = project.gradleProject
@@ -40,8 +42,10 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':the-project:buildHealth')
     build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
 
-    then: 'and there is the expected advice'
+    then: 'the build health of the first build is the same as when running that build directly'
     assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth(':'))
+
+    then: 'the build health of the second build is the same as when running that build as included build'
     assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild('the-project'))
 
     where:
@@ -93,7 +97,7 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
   }
 
-  def "result of analysis does not change if root build of included build tree changes (with subprojects)"() {
+  def "result of analysis does not change if root build of included build tree changes for projects with subprojects"() {
     given:
     def project = new IncludedBuildWithSubprojectsProject(true)
     gradleProject = project.gradleProject
@@ -109,8 +113,6 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
   }
 
   def "up-to-date check is correct when switching root builds"() {
-    shouldClean = false
-
     given:
     def project = new IncludedBuildWithSubprojectsProject(true)
     gradleProject = project.gradleProject
@@ -138,16 +140,19 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
 
     when:
     build(gradleVersion, gradleProject.rootDir, '--build-cache', ':buildHealth')
+
     then:
     assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
 
     when: 'Running again - UP-TO-DATE'
     build(gradleVersion, gradleProject.rootDir, '--build-cache', ':buildHealth')
+
     then: 'Result is the same'
     assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
 
     when: 'Running again - FROM-CACHE'
     build(gradleVersion, gradleProject.rootDir, '--build-cache', 'clean', ':buildHealth')
+
     then: 'Result is the same'
     assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
 
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithAnnotationProcessorProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithAnnotationProcessorProject.groovy
@@ -1,0 +1,126 @@
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.*
+import com.autonomousapps.model.Advice
+import com.autonomousapps.model.ProjectAdvice
+
+import static com.autonomousapps.AdviceHelper.*
+
+final class IncludedBuildWithAnnotationProcessorProject extends AbstractProject {
+
+  final GradleProject gradleProject
+
+  IncludedBuildWithAnnotationProcessorProject() {
+    this.gradleProject = build()
+  }
+
+  private GradleProject build() {
+    def builder = newGradleProjectBuilder()
+    builder.withRootProject { root ->
+      root.settingsScript.additions = """\
+        includeBuild 'processor-build'
+        include 'user-of-processor'
+      """.stripIndent()
+    }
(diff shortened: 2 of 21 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithSubprojectsProject.groovy
```diff
@@ -107,8 +109,12 @@ final class IncludedBuildWithSubprojectsProject extends AbstractProject {
     return actualProjectAdvice(project)
   }
 
-  final Set<ProjectAdvice> expectedIncludedBuildHealth = [
-    projectAdviceForDependencies(':second-sub1', [] as Set<Advice>),
+  final Set<ProjectAdvice> expectedIncludedBuildHealth(String buildNameInAdvice) {[
+    projectAdviceForDependencies(':second-sub1', [
+      useProjectDependencyWherePossible
+        ? Advice.ofChange(projectCoordinates(':second-sub2', null, buildNameInAdvice), 'api', 'implementation')
+        : Advice.ofChange(includedBuildCoordinates('second:second-sub2', projectCoordinates(':second-sub2', 'second:second-sub2', buildNameInAdvice)), 'api', 'implementation')
```
jjohannes, line 116: This was not covered by the tests: Inside an included build, I may refer to my own projects in two different ways:
- `project(':path')`
- `group:name`

Before this PR, it depended on if the project is in the _root_ or one of the other builds, which notation would be suggested. Now the used notation is preserved (for change/remove advices). For add advices, the  `project(':path')` notation is suggested always.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -87,13 +82,13 @@ internal fun ResolvedArtifactResult.toCoordinates(): Coordinates {
 /** Returns the [coordinates][Coordinates] of the root of [this][Configuration]. */
 internal fun Configuration.rootCoordinates(): Coordinates = incoming.resolutionResult.root.id
   // For the root, the 'GradleVariantIdentification' is always empty as there is only one root (which we match later)
-  .toCoordinates(GradleVariantIdentification(emptySet(), emptyMap()))
+  .toCoordinates(GradleVariantIdentification(setOf("ROOT"), emptyMap()))
```
jjohannes, line 85: In some places, the ROOT node is part of the graph/classpath although dependencies never point at it. For the root, we do not know which "variant" it is – e.g. are we looking at the "main" or the "test-fixtures" root? Since there might be different nodes in the graph where dependencies point back at a variant of the "root" the root itself should not be confused with any of these. Hence we give it a special capability "ROOT" here. Otherwise, the plugin thinks that the ROOT is always equal to the "main" variant.

This ROOT could probably be removed in some places where we do analysis. It is not relevant as a target of a dependency, when the nodes are compared to compute advices.

src/main/kotlin/com/autonomousapps/model/Coordinates.kt
```diff
@@ -115,7 +115,8 @@ sealed class Coordinates(
 @JsonClass(generateAdapter = false)
 data class ProjectCoordinates(
   override val identifier: String,
-  override val gradleVariantIdentification: GradleVariantIdentification
+  override val gradleVariantIdentification: GradleVariantIdentification,
+  val buildName: String? = null // Name of the build in a composite for which the project coordinates are valid
```
jjohannes, line 119: We track the **build name**, so that we can decide, for **add** advices, if a `project(:path)` dependency is possible – which is only if **source** and **target** are in the same build.

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -15,4 +15,26 @@ data class GradleVariantIdentification(
 
   private fun toSingleString() =
     capabilities.sorted().joinToString() + attributes.map { (k, v) -> "$k=$v" }.sorted().joinToString()
+
+  /**
+   * Check that all the requested capabilities are declared in the 'target'. Otherwise, the target can't be a target
+   * of the given declarations. The requested capabilities ALWAYS have to exist in a target to be selected.
+   */
+  fun variantMatches(target: Coordinates): Boolean = when(target) {
```
jjohannes, line 23: Moved this method to be reused in two places

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -50,30 +50,25 @@ private fun ProjectComponentIdentifier.identityPath(): String {
...
   if (resolved !is ProjectCoordinates) return resolved
 
-  // may be a composite substitution
-  val identity = ProjectCoordinates(
-    (id.componentIdentifier as ProjectComponentIdentifier).identityPath(),
-    gradleVariantIdentification
-  )
-
-  // Identity path matches project path, so we assume this isn't resolved from an included build, and return as-is.
-  if (resolved == identity) return resolved
```
jjohannes, line 66: Not doing this now leads to all **ProjectCoordinates** being wrapped into **IncludedBuildCoordinates**. Overall, I think this is a good change to move a more consistent setup.

We could add some other check here. We could check, if we are in an included build setup at all. That is, for builds that do not include anything, we could still do the traditional approach which would keep this behavior for most builds.

However, to do that we would need more context passed into the method. At lease some `runningWithIncludedBuilds: Boolean` parameter. I thought about this, but it would ass more complexity without any additional functionality. It would just be a safeguard to avoid using the "new" approach in "traditional" builds.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -50,30 +50,25 @@ private fun ProjectComponentIdentifier.identityPath(): String {
...
-
+  // Module may have been resolved from an included build. Construct IncludedBuildCoordinates if possible.
   // This is a very naive heuristic. Doesn't work for Gradle < 7.2, where capabilities is empty.
-  val requested = variant.capabilities.firstOrNull()?.let { c ->
+  val projectName = (owner as ProjectComponentIdentifier).projectName
+  val requested = capabilities.find { it.name.startsWith(projectName) }?.let { c ->
     c.version?.let { v ->
       ModuleCoordinates(
-        identifier = "${c.group}:${c.name}",
+        identifier = "${c.group}:$projectName",
```
jjohannes, line 69: In a sane setup, the "projectName" corresponds to the second element of the GA coordinate.

Using the capability here is not correct. We missed it before. For test fixtures for example, this is wrong because then `project-name-test-fixtures` would be used as identifier. We just did not hit this bug before because most tests were done from the "root" build. The existing tests captured this after the change.

Only the group is still guesswork then. But in normal project structures, it should be correct like this. I will open an issue at Gradle that the coordinates should be provided in the result here.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -259,7 +266,11 @@ internal class DependencyAdviceBuilder(
         when {
           // The user cannot change this one:
           // https://github.com/gradle/gradle/blob/d9303339298e6206182fd1f5c7e51f11e4bdff30/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaTestFixturesPlugin.java#L68
-          advice.coordinates.identifier == projectPath && advice.fromConfiguration?.endsWith("testFixturesApi") ?: false -> {
+          advice.coordinates.identifier == projectPath && advice.fromConfiguration?.equals("testFixturesApi") ?: false -> {
+            null
+          }
+          // https://github.com/gradle/gradle/blob/d9303339298e6206182fd1f5c7e51f11e4bdff30/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaTestFixturesPlugin.java#L70
+          advice.coordinates.identifier == projectPath && advice.fromConfiguration?.lowercase()?.endsWith("testimplementation") ?: false -> {
```
jjohannes, line 273: This was also missed due to the buggy situation of "root vs included build". The existing tests now showed that the also needs special handling.

(Also I am blind or something - this line in the Gradle code base is just two lines below the other one above and I missed it before 🙈 )

src/main/kotlin/com/autonomousapps/tasks/SynthesizeDependenciesTask.kt
```diff
@@ -180,10 +183,6 @@ abstract class SynthesizeDependenciesTask @Inject constructor(
         .forEach { dependency ->
           val coordinates = dependency.coordinates
           outputDir.file(coordinates.toFileName()).get().asFile.bufferWriteJson(dependency)
-          if (coordinates is IncludedBuildCoordinates) {
-            // Write the same information using the 'project' coordinates as later processing steps rely on it
-            outputDir.file(coordinates.resolvedProject.toFileName()).get().asFile.bufferWriteJson(dependency)
```
jjohannes, line 185: This was bound to cause issues – writing the same information twice with different file names. This is now no longer necessary as "inside" the analysis the `IncludedBuildCoordinates` notation is used consistently.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
jjohannes: Rediscovered: https://github.com/gradle/gradle/issues/20048 again. Will comment on that one.

autonomousapps: > In a sane setup, the "projectName" corresponds to the second element of the GA coordinate.

Can you elaborate on this? Because most Android projects I've seen are not "sane" in this way. They almost always have some random project name and custom artifact ID because (1) Android devs do not care about sane builds and (2) Gradle docs are confusing.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -19,15 +20,34 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
     gradleProject = project.gradleProject
 
     when: 'build does not fail'
-    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+    build(gradleVersion, gradleProject.rootDir, ':buildHealth')
```
autonomousapps, line 23: why did you add the `:` prefix here?

one of the constraints of the way this plugin is currently architected is that we generally need to invoke the task as `buildHealth` to force gradle to configure all of the subprojects, which is how results get aggregated into the `buildHealth` report.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -19,15 +20,34 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
+    build(gradleVersion, gradleProject.rootDir, ':second-build:buildHealth')
 
-    then: 'and there is no advice'
-    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+    then: 'and there is the expected advice'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth('second-build'))
+    assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild(':'))
```
autonomousapps, line 28: Given that we have two assertions, can we use `and:` / `then:` with more concrete string descriptions to help explain what we're asserting?

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -19,15 +20,34 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
 
     where: 'This new feature only works for Gradle 7.3+'
     gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
   }
 
+  def "result of analysis does not change if root build of included build tree changes (only root projects)"() {
```
autonomousapps, line 34: by `(only root projects)`, do you mean that we're only analyzing the root projects? that would be in accord with invoking the task as `:buildHealth` instead of `buildHealth`.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -19,15 +20,34 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
+    def project = new IncludedBuildProject()
+    gradleProject = project.gradleProject
+
+    when: 'build running from second build root does not fail'
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':the-project:buildHealth')
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
+
+    then: 'and there is the expected advice'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth(':'))
+    assertThat(project.actualBuildHealthOfSecondBuild()).containsExactlyElementsIn(project.expectedBuildHealthOfIncludedBuild('the-project'))
```
autonomousapps, line 45: same comment as above.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -63,11 +83,73 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "result of analysis does not change if root build of included build tree changes (with subprojects)"() {
+    given:
+    def project = new IncludedBuildWithSubprojectsProject(true)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, new File(gradleProject.rootDir, 'second-build'), ':buildHealth')
```
autonomousapps, line 102: etc

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -63,11 +83,73 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
+
+    then:
+    assertThat(project.actualIncludedBuildHealth()).containsExactlyElementsIn(project.expectedIncludedBuildHealth(':'))
+
+    where:
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "up-to-date check is correct when switching root builds"() {
+    shouldClean = false
```
autonomousapps, line 112: how is this being used?

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
```diff
@@ -63,11 +83,73 @@ final class IncludedBuildSpec extends AbstractJvmSpec {
...
+    where:
+    gradleVersion << INCLUDED_BUILD_SUPPORT_GRADLE_VERSIONS
+  }
+
+  def "can handle annotation processors from cache in subsequent builds"() {
+    given:
+    def project = new IncludedBuildWithAnnotationProcessorProject()
+    gradleProject = project.gradleProject
+
+    when:
```
autonomousapps, line 139: nit: missing vertical whitespace

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithAnnotationProcessorProject.groovy
```diff
@@ -0,0 +1,126 @@
...
+  IncludedBuildWithAnnotationProcessorProject() {
+    this.gradleProject = build()
+  }
+
+  private GradleProject build() {
+    def builder = newGradleProjectBuilder()
+    builder.withRootProject { root ->
+      root.settingsScript.additions = """\
+        includeBuild 'processor-build'
+        include 'user-of-processor'
```
autonomousapps, line 23: good names, thanks

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithSubprojectsProject.groovy
autonomousapps: I actually didn't know you could refer to your own projects using `group:name` coordinates.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -92,6 +98,7 @@ internal class Bundles(private val dependencyUsages: Map<Coordinates, Set<Usage>
 
       // Handle bundles that don't have a primary entry point
       dependencyGraph.forEach { (_, view) ->
+        val projectNode = view.graph.nodes().find { it.identifier == projectPath }!!
```
autonomousapps, line 101: and the `!!` is appropriate because we know that the graph must have a node in it where the node's identifier matches the `projectPath`.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -144,5 +151,12 @@ internal class Bundles(private val dependencyUsages: Map<Coordinates, Set<Usage>
 
       return bundles
     }
+
+    private fun coordinatesOrPathEquals(coordinates: Coordinates, primaryId: String) =
+      coordinates.identifier == primaryId || (coordinates is IncludedBuildCoordinates) && coordinates.resolvedProject.identifier == primaryId
```
autonomousapps, line 156: the parentheses look redundant here. did you mean the `)` to include the third boolean as well, after the `&&`?

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -144,5 +151,12 @@ internal class Bundles(private val dependencyUsages: Map<Coordinates, Set<Usage>
...
+
+    private fun coordinatesOrPathMatch(coordinates: Coordinates, regex: Regex) =
+      regex.matches(coordinates.identifier) || (coordinates is IncludedBuildCoordinates) && regex.matches(coordinates.resolvedProject.identifier)
```
autonomousapps, line 159: same as above

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -15,4 +15,26 @@ data class GradleVariantIdentification(
...
+    is FlatCoordinates -> true
+    is ProjectCoordinates -> if (capabilities.isEmpty()) {
+      target.gradleVariantIdentification.capabilities.any {
+        it.endsWith(target.identifier.substring(target.identifier.lastIndexOf(":"))) // If empty, needs to contain the 'default' capability (for projects we need to check with endsWith)
```
autonomousapps, line 27: nit: move comment above `it.endsWith...` for line length reasons.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -835,6 +840,8 @@ internal class ProjectPlugin(private val project: Project) {
     )
   }
 
+  private fun Project.buildName(): String = serviceOf<BuildState>().buildIdentifier.name
```
autonomousapps, line 843: how safe / future-proof is this?

and since it's called multiple times, is there benefit in caching the value after first lookup?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
autonomousapps: happens to the best of us

autonomousapps (commented):
Thank you! Please respond to my questions when you get a chance. Really appreciate the work here.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithSubprojectsProject.groovy
jjohannes: This is possible as soon as you "opt-in" by including at least one build.

If you want it for a "singe build" you can do `includeBuild(".")` - https://docs.gradle.org/current/userguide/composite_builds.html#included_build_declaring_substitutions

It is good to have a consistent notation if you work with included builds to split up your own product. Then you don't need different notations to refer to the same component in different places.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
jjohannes: Sorry for the sloppy wording.

If you change the artifact ID for publishing (which I believe is what you refer to) it has no influence here. The identity will be different, when you use the published version of the component. But as long as the component is a project in one of the builds, such a reconfiguration has no effect on what this plugin is looking at.

The only situation where the project name would not correspond to the "artifact name" is when the included build substitution is changed. This is a feature of Gradle I am not very happy with that it exists at all. :) But that problem already exists in the current plugin release. And adding more information to the resolution result for these cases is what https://github.com/gradle/gradle/issues/20048 is about.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
jjohannes: Background: When you call a task in an included build, the fuzzy matching (without `:`) is not supported. Although I never saw anything official about that, it was discussed that it is generally better to be precise when invoking Gradle (i.e. it is idiomatic to always use `:` notation). And if you want an entry point that calls a number of tasks at once, you should create a _lifecycle task_ with multiple _dependsOn_. (This is rather cumbersome right now, but that's how things currently are if you work with a larger included build setup.)

In these tests I wanted to be consistent with calling `:buildHealth`, and for the included build there is no other way to call it:

- `:buildHealth` build health of the root build (could also call `buildHealth`
- `:second:buildHealth` build health of the _second_ build (no other way to call it)

Generally, I think doing `:buildHealth` is fine as there is only one `buildHealth` task per build. Gradle always configures all the subprojects of a build independent of which task you call. Such considerations might only be important for the old _configure-on-demand_ flag, which is half broken anyway in my opinion (and will be removed I think soonish).

If you prefer to change this back to `buildHealth` in the tests wherever possible, I can do that. I have no strong opinion on that.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
jjohannes: No. It's about the projects not having subprojects that reference each other. I changed the test name.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
jjohannes: whups. Missed that from debugging. Removed.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithAnnotationProcessorProject.groovy
jjohannes: Wasn't me ;) - this is taken (and adjusted) from #896

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
jjohannes: Before, we were creating a Coordinates object and assumed that that exact object – according to `equals` – is in the graph. Now there is more information attached to the Coordinates – in particular the build name. So that did not work anymore.

I added a comment.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
jjohannes: It's redundant. Good catch. Removed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
jjohannes: Very good point. I asked in the Gradle Community Slack and Stefan pointed me at a better solution. This is now updated.

I also discovered that with Gradle 8.2, `buildName` is replaced by `buildPath`. We should do a version check and use that for new Gradle versions. I added a FIXME to do this, once the plugin build has been updated to the latest Gradle version.

See: https://docs.gradle.org/8.2-rc-2/userguide/upgrading_version_8.html#build_identifier_name_and_current_deprecation

jjohannes:
Thank you for the thorough review @autonomousapps. I addressed your comments.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
autonomousapps: Ah, I meant in part publishing, but also the included build substitution thing. I've had to walk folks through that so many times. There's a good (bad) example of that in the build where I work. All the modules are named `:feature1:public`, `:feature2:public`, and so on. So if went with the default, then every projectName is just "public", which of course doesn't work.

src/functionalTest/groovy/com/autonomousapps/jvm/IncludedBuildSpec.groovy
autonomousapps: > Gradle always configures all the subprojects of a build independent of which task you call.

This was not my experience writing these tests. At one point I tried to be more idiomatic and added the `:` prefix, and the tests started failing. Maybe there was a bug somewhere. And I am also certain that I ran into it at work, where we use do CoD.

If your tests are passing (they are), and the assertions make sense to you, then you can leave it as it is.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -835,6 +839,13 @@ internal class ProjectPlugin(private val project: Project) {
     )
   }
 
+  /** Get the buildPath of the current build from the root component of the resolution result. */
+  private fun Project.buildPath(configuration: String) = project.provider {
+    // Note: starting with Gradle 7.4, we can replace 'project.provider' with 'resolutionResult.rootComponent.map'
```
autonomousapps, line 844: what's this note about? does that relate to Gradle's CC support for accessing the dependency graph?

autonomousapps (approved):
LGTM. I did leave one new question. It's an important one because I would like to get around to supporting CC in this plugin this year.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
jjohannes: The new Provider bases API – resolutionResult.rootComponent (as alternative to resolutionResult.root) – was added in that context I believe. But  doing it like we do now also works is you want to support older Gradle versions.

I use the current solution in another plugin as well and it is CC compatible.
[ExtraJavaModuleInfoPlugin.java#L131-L134](https://github.com/gradlex-org/extra-java-module-info/blob/main/src/main/java/org/gradlex/javamodule/moduleinfo/ExtraJavaModuleInfoPlugin.java#L131-L134)

Something like that is probably also what you would need to do here in the places where you now pass "Configurations" into tasks.

PR #918, opened 2023-06-30
Add test: nested projects with same name have different identities

jjohannes:
Because by default, the 'group' is derived from the project path, the identity differs even if if GA from the IncludeBuildCoordinates is used as identity.

Confirms that there is no problem here after the changes from #916.

See:
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/916#discussion_r1244109793

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/NestedProjectsSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/NestedProjectsSpec.groovy
@@ -1,0 +1,50 @@
+package com.autonomousapps.jvm
+
+
+import com.autonomousapps.jvm.projects.NestedSubprojectsProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.autonomousapps.utils.Runner.buildAndFail
+import static com.google.common.truth.Truth.assertThat
+
+final class NestedProjectsSpec extends AbstractJvmSpec {
+
+  def "handles projects with same name but different nested path"() {
+    given:
+    def project = new NestedSubprojectsProject()
+    gradleProject = project.gradleProject
+
+    when: 'build does not fail'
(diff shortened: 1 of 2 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/NestedProjectsSpec.groovy
```diff
@@ -0,0 +1,50 @@
...
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then: 'the build health is as expected'
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  def "handles projects with same name and group but different nested path"() {
```
autonomousapps, line 27: can we change the test name to "does not handle..."?

src/functionalTest/groovy/com/autonomousapps/jvm/NestedProjectsSpec.groovy
```diff
@@ -0,0 +1,50 @@
...
+    def result = buildAndFail(gradleVersion, gradleProject.rootDir, 'buildHealth')
+
+    then: 'build fails with circular dependency issue'
+    result.output.contains '''
+      Circular dependency between the following tasks:
+      :featureC:public:compileJava
+      \\--- :featureC:public:compileJava (*)
+    '''.stripIndent()
+
+    where: 'This new feature only works for Gradle 7.3+'
```
autonomousapps, line 46: this string looks like copy-paste error?

autonomousapps (approved):
LGTM, but left some minor comments I'd like addressed pre-merge.

jjohannes:
Thanks for the feedback. Fixed.

PR #919, opened 2023-07-03
Support for correctly resolving KMP artifacts

### LINKED ISSUE #907: DAGP should be aware of KMP (multiplatform) artifacts and not advise downgrading from multiplatform facade to "-android"-specfic artifact (androidx is one example)
Bump from alphas to betas causing issues
https://github.com/google/horologist/pull/1312/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df

Failures https://github.com/google/horologist/actions/runs/5088063599/jobs/9144041802

```
Caused by: com.autonomousapps.exception.BuildHealthException: Unused dependencies which should be removed:
  api("androidx.compose.foundation:foundation-layout:1.5.0-beta01")
  api("androidx.compose.foundation:foundation:1.5.0-beta01")
  api("androidx.compose.runtime:runtime:1.5.0-beta01")
  api("androidx.compose.ui:ui-graphics:1.5.0-beta01")
  api("androidx.compose.ui:ui:1.5.0-beta01")
  debugImplementation("androidx.compose.material:material-icons-extended:1.5.0-beta01")
  debugImplementation("androidx.compose.ui:ui-tooling-preview:1.5.0-beta01")
  implementation("androidx.compose.material:material-icons-core:1.5.0-beta01")
  implementation("androidx.compose.ui:ui-text:1.5.0-beta01")
  implementation("androidx.compose.ui:ui-unit:1.5.0-beta01")
  testImplementation("androidx.compose.material:material-icons-extended:1.5.0-beta01")
  testImplementation("androidx.compose.ui:ui-test-junit4:1.5.0-beta01")
  testImplementation("androidx.compose.ui:ui-test:1.5.0-beta01")

Transitively used dependencies that should be declared directly as indicated:
  api("androidx.compose.foundation:foundation-android:1.5.0-beta01")
  api("androidx.compose.foundation:foundation-layout-android:1.5.0-beta01")
  api("androidx.compose.runtime:runtime-android:1.5.0-beta01")
  api("androidx.compose.ui:ui-android:1.5.0-beta01")
  api("androidx.compose.ui:ui-graphics-android:1.5.0-beta01")
  debugImplementation("androidx.compose.material:material-icons-extended-android:1.5.0-beta01")
  debugImplementation("androidx.compose.ui:ui-tooling-preview-android:1.5.0-beta01")
  implementation("androidx.compose.material:material-icons-core-android:1.5.0-beta01")
  implementation("androidx.compose.ui:ui-text-android:1.5.0-beta01")
  implementation("androidx.compose.ui:ui-unit-android:1.5.0-beta01")
  testImplementation("androidx.compose.material:material-icons-extended-android:1.5.0-beta01")
  testImplementation("androidx.compose.ui:ui-test-android:1.5.0-beta01")
  testImplementation("androidx.compose.ui:ui-test-junit4-android:1.5.0-beta01")
```

**Build scan link**

**Plugin version**
1.20.0

**Gradle version**
8.1 (there is a warning it is unsupported)

8.0.2

It feels like apps should just depend on androidx.compose.ui:ui instead of androidx.compose.ui:ui-android

ZacSweers:
This implements support for gracefully handling Kotlin multiplatform dependencies.

## Nomenclature

For lack of a better word, I settled on disambiguating KMP dependencies between "canonical" (i.e. `androidx.compose.material:material`) and "targets" (i.e. `androidx.compose.material:material-android`).

## Design

There are three specific categories we want to gracefully handle in this case
1. If a canonical KMP dep is used, _don't_ try to remove it if a specific target is intended instead. This defeats the intended design of how these dependencies are intended to work in a KGP-handled project.
2. If a target dependency is added, defer to the canonical one IFF it's a kotlin project and we had previously declared the canonical one.
3. If a target dependency is added as a transitive misused dep and IFF it's a kotlin project, swap it with the canonical dependency instead.

The way we define a KMP canonical dependency is if it has a kotlin platform type attribute and _does_ have an `externalVariant` defined. The `externalVariant` presence indicates that this variant is essentially just a proxy to another variant, which is how KMP artifacts work. A target dependency is if it has a kotlin platform type attribute and does _not_ have an `externalVariant` defined, indicating it is the proxied artifact. It's not enough to look at the value of the KotlinPlatformType attribute alone as it will say that it's the target type even on the original canonical dependency.

This tweaks the `ComputeAdviceTask` to run in essentially two passes. A first pass is taken read information about the KMP advice and defer the rest. Then the KMP advice are filtered with the above criteria. Finally, the second pass runs through both the deferred and filtered KMP advice together. It's important to do the KMP advice filtering _first_ so that bundle rules can still operate with canonical artifacts in mind.

A couple small API changes were adding resolved attribute data to usage discovery (only including the kotlin platform type for now to cut down on JSON data size) and the external variant metadata to `GradleVariantIdentification`. I also added a convenience `GradleVariantIdentification.EMPTY` to cover a common empty case I was seeing in the project.

Resolves #907

Using https://github.com/ZacSweers/CatchUp/pull/789 as a test bed


ZacSweers:
Extremely open to suggestions on where to start on testing this!

ZacSweers:
So an interesting thing I observed while testing this is that there's no way to handle a misused dependency that's linked from a different project I think.

Consider this case

1. Depend on `org.jetbrains.compose.foundation:foundation` in an android project, which depends on `org.jetbrains.compose.runtime:runtime`.
2. Use only a class from `androidx.compose.runtime`, which is in the `androidx.compose.runtime:runtime` for android and `org.jetbrains.compose.runtime:runtime` for any other target. The JB artifact just depends on the android one for its `androidJvm` artifact.
3. DAGP will advice just using `androidx.compose.runtime:runtime`. There doesn't seem to be a way to reconnect to the JB artifact as it's just declaring a dependency.

I think, at least for now, this case should probably just stand and consumers should manually handle this on their side if they care. We could also program in special-cased logic for compose multiplatform if we want to smoothen this specific case, but it'd be touchy on how to do this in a fully correct way.

ZacSweers:
Got a basic test set up but the current advice won't try to swap a _used_ `-jvm` target out for a canonical dep, which I think we should do but ok leaving as a toe-hold for a later pass. I've left a toe-hold for this in the test as well.

I think this could be another change-advice that comes from `StandardTransform`, but I don't fully understand the code logic in there and would be open to pointers on where to start. Basically we want to create a new transform that says "hey, if we have a used KMP declaration that's pointing at a specific target but we're in a kotlin project, just switch it for the canonical one".

autonomousapps:
> So an interesting thing I observed while testing this is that there's no way to handle a misused dependency that's linked from a different project I think.
> 
> Consider this case
> 
> 1. Depend on `org.jetbrains.compose.foundation:foundation` in an android project, which depends on `org.jetbrains.compose.runtime:runtime`.
> 2. Use only a class from `androidx.compose.runtime`, which is in the `androidx.compose.runtime:runtime` for android and `org.jetbrains.compose.runtime:runtime` for any other target. The JB artifact just depends on the android one for its `androidJvm` artifact.
> 3. DAGP will advice just using `androidx.compose.runtime:runtime`. There doesn't seem to be a way to reconnect to the JB artifact as it's just declaring a dependency.
> 
> I think, at least for now, this case should probably just stand and consumers should manually handle this on their side if they care. We could also program in special-cased logic for compose multiplatform if we want to smoothen this specific case, but it'd be touchy on how to do this in a fully correct way.

In my opinion, this is exactly the use-case that bundles are designed for. I've been thinking that it might be worthwhile to add a way to define bundles in a file (similar to how Gradle core does it with version catalogs). That would eliminate or at least move the verbosity people don't like in their build scripts to a purpose-built file definition. We might even go a step further and provide some default definitions for common use-cases like this KMP stuff and the massive Compose BOM, as a companion.

autonomousapps:
> Got a basic test set up but the current advice won't try to swap a _used_ `-jvm` target out for a canonical dep, which I think we should do but ok leaving as a toe-hold for a later pass. I've left a toe-hold for this in the test as well.
> 
> I think this could be another change-advice that comes from `StandardTransform`, but I don't fully understand the code logic in there and would be open to pointers on where to start. Basically we want to create a new transform that says "hey, if we have a used KMP declaration that's pointing at a specific target but we're in a kotlin project, just switch it for the canonical one".

I'm also okay with this as-is. If folks declare a specific dependency, I'd like to assume they mean it.

As far as `StandardTransform` is concerned, that algorithm... evolved in a TDD way to meet all the requirements I had for good, non-destructive, stable advice. It's very challenging to describe -- maybe take a look at the test suite. Furthermore, the name is an indication that I originally meant for there to be multiple kinds of `Transform`s to meet different needs, but that has not (yet) come to pass. I suppose if we had a valid use-case (and I'm not convinced that what you described is something I'd want to do, but I could perhaps be convinced), we could chain another transform in, or add some logic to use different transforms for different conditions.

autonomousapps:
Two failed tests https://scans.gradle.com/s/xpyotyswcmff4.
Screenshot 2023-07-05 at 11 10 22 AM

src/functionalTest/groovy/com/autonomousapps/android/projects/AndroidProjectWithKmpDependencies.groovy
```diff
@@ -0,0 +1,141 @@
...
+    builder.withAndroidSubproject('app') { s ->
+      s.manifest = AndroidManifest.app('com.example.MainApplication')
+      s.sources = sources
+      s.withBuildScript { bs ->
+        bs.plugins = [Plugin.androidAppPlugin, Plugin.kotlinAndroidPlugin]
+        bs.dependencies = [
+          kotlinStdLib('implementation'),
+          appcompat('implementation'),
+
+          // Immutable collections JVM dep that should be corrected to the canonical target
```
autonomousapps, line 44: Thanks for the comments here 👍

src/functionalTest/groovy/com/autonomousapps/android/projects/AndroidProjectWithKmpDependencies.groovy
```diff
@@ -0,0 +1,141 @@
...
+          kotlinxImmutable('implementation', "-jvm"),
+
+          // Coroutines Test JVM dep that should be
+          // - Swapped with the core dep
+          // - Core dep should use the canonical target
+          kotlinxCoroutinesTest('implementation', "-jvm"),
+
+          // A foundation compose dependency but we only use runtime APIs
```
autonomousapps, line 52: when you say "runtime", do you literally mean it could be a `runtimeOnly` dependency? Just want to clarify that for myself.

src/main/kotlin/com/autonomousapps/internal/kotlin/KotlinPlatformType.kt
```diff
@@ -0,0 +1,20 @@
+package com.autonomousapps.internal.kotlin
+
+import org.gradle.api.Named
+import org.gradle.api.attributes.*
+import java.io.Serializable
+
+/** Shaded copy of the real KGP [org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType] for classpath reasons. */
+enum class KotlinPlatformType: Named, Serializable {
```
autonomousapps, line 8: should this be public or `internal`?

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -20,6 +21,11 @@ import org.gradle.api.internal.artifacts.DefaultProjectComponentIdentifier
 import org.gradle.internal.component.local.model.OpaqueComponentArtifactIdentifier
 import org.gradle.internal.component.local.model.OpaqueComponentIdentifier
 
+/** To avoid bloating intermediate JSON, we only capture a subset of attributes DAGP is interested in. */
```
autonomousapps, line 24: Very valid concern.

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -39,4 +43,39 @@ data class GradleVariantIdentification(
       target.gradleVariantIdentification.capabilities.containsAll(capabilities)
     }
   }
+
+  companion object {
+    val EMPTY = GradleVariantIdentification(emptySet(), emptyMap())
```
autonomousapps, line 48: Thanks for this 👍

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -39,4 +43,39 @@ data class GradleVariantIdentification(
...
+
+/**
+ * KMP artifacts have a [KotlinPlatformType] attribute defined for which target it is.
+ *
+ * At a general level, these can be thought of as a canonical proxy dependency and "the rest", which
+ * covers specific platform targets like [KotlinPlatformType.jvm], [KotlinPlatformType.native], etc.
+ *
+ * When inspecting attributes for the context of DAGP, we want to know if a given dependency is a canonical
+ * dependency or a specific target. When recommending advice, we want to defer to the canonical deps as
+ * a sort of implicit bundle of its other targets and will resolve to the correct target.
```
autonomousapps, line 60: Thanks for the consistently useful kdoc 👍

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -39,4 +43,39 @@ data class GradleVariantIdentification(
...
+/** Returns whether this is a specific target KMP dependency, such as [KotlinPlatformType.jvm]. */
+internal val Coordinates.isKmpTargetTarget: Boolean
+  get() = kmpAttribute != null && gradleVariantIdentification.externalVariant == null
+
+/**
+ * Returns a [Coordinates.gav] string that represents the parent [Coordinates.identifier] of the given KMP target's
+ * `common` parent.
+ */
+internal fun ModuleCoordinates.kmpCommonParentIdentifier(): String {
+  return identifier.substringBeforeLast('-')
```
autonomousapps, line 80: How safe is this string manipulation? Do we "know" that they'll always use a suffix that starts with a `-`?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +262,79 @@ internal class DependencyAdviceBuilder(
...
-    return dependencyUsages.asSequence()
+
+    /*
+     * To support KMP properly, we need to do an initial pass at deps to clean up any KMP deps advice.
+     *
+     * The primary case we are trying to catch and avoid here is one where we "fix" canonical KMP deps by trying to replace
+     * them with added deps on their specific targets ("-jvm", "-android", etc). To solve this, we save off information
+     * about both in the below mappings and then cross-reference them in the second pass farther down.
+     *
+     * Finally, in the third pass, we do our usual processing of advice but with the cleaned KMP advice. It's
```
autonomousapps, line 273: Three passes 😬 I haven't looked at the code below yet, but do you have a sense for the performance costs here? Also, should we just skip this stuff if we're not in a Kotlin project?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +262,79 @@ internal class DependencyAdviceBuilder(
...
+     * important we clean up the KMP advice first so that the usual processing can continue to account for bundles
+     * in a simple way.
+     */
+
+    // Deferred KMP advice.
+    val deferredKmpAdvice = mutableListOf<Pair<Advice, Coordinates>>()
```
autonomousapps, line 279: - [ ] TODO for me: open this up in an IDE and make sure I understand the changes 😅

autonomousapps (commented):
Thanks for this! The only complex part I need to study more is the core of the changes in `ComputeAdviceTask`. Plus we need to resolve the test failures.

src/functionalTest/groovy/com/autonomousapps/android/projects/AndroidProjectWithKmpDependencies.groovy
ZacSweers: ah no, I mean it's from the `androidx.compose.runtime:runtime` artifact

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
ZacSweers: h/t to @jjohannes for pointing it out

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
ZacSweers: as far as I can tell, yes. I'd love a more programmatic way but I haven't been able to decipher where in Gradle's APIs one could find this information and think it's just implicit.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: it's more like 2.1 :). The middle pass is just looking at the KMP deps we collected in the first pass

ZacSweers: To answer your question, I'm not super fussed about the performance cost in this case. Still linear complexity. The absolute worst case is O(2.5n) (i.e. 50% of deps are KMP) if I'm remembering my algos and complexity right

src/main/kotlin/com/autonomousapps/internal/kotlin/KotlinPlatformType.kt
ZacSweers: [5f179ec](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/5f179ecc69505a94aef187b704ad44ca3ec872e6)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
autonomousapps: Alrighty, I'm willing to accept that analysis.

src/main/kotlin/com/autonomousapps/tasks/ComputeUsagesTask.kt
```diff
@@ -130,7 +130,15 @@ private class GraphVisitor(
     var hasSecurityProvider = false
     var hasNativeLib = false
 
-    dependency.capabilities.values.forEach { capability ->
+    var capabilities = dependency.capabilities
+
+    dependencyCoordinates.gradleVariantIdentification.externalVariant?.let { externalVariant ->
+      // If this is just a shim to an external variant, defer to its capabilities instead
+      val externalIdentifier = externalVariant.capabilities.single()
```
autonomousapps, line 137: wrt to `single()`, do we know that there's only one capability here?

src/main/kotlin/com/autonomousapps/tasks/ComputeUsagesTask.kt
```diff
@@ -191,10 +199,7 @@ private class GraphVisitor(
       }
     }
 
-    // TODO KMP dependencies only contain metadata (pom/module files). This is good evidence of a facade and could be
-    //  used for smarter detection of same.
-    // An example are KMP facades that resolve to -jvm artifacts
-    if (dependency.capabilities.isEmpty()) {
+    if (capabilities.isEmpty()) {
```
autonomousapps, line 202: 👍

autonomousapps (commented):
Left a question, and I still need to look in detail at the new algorithm (that's next).

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -310,6 +401,27 @@ internal class DependencyAdviceBuilder(
         StandardTransform(coordinates, declarations, supportedSourceSets, buildPath, isKaptApplied).reduce(usages)
       }
   }
+
+  /**
+   * Returns whether this is a KMP non-common target that can just defer to a parent common declaration in
+   * [kmpCommonDeclarations].
+   *
+   * @see GradleVariantIdentification.kmpAttribute
```
autonomousapps, line 409: My IDE cannot resolve this. Looks like it should be `Coordinates.kmpAttribute`.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
...
+      }
+
+    val modifiedKmpAdvice = deferredKmpAdvice
+      // "null" removes the advice
+      .mapNotNull { (advice, originalCoordinates) ->
+        when {
+          // KMP target dependencies ("<dep>-android", "<dep>-jvm", etc) should defer to using their parent comment dep
+          // If the parent _isn't_ present in the original declarations though, then accept it and assume they're
+          // intentionally only depending on that specific target's artifact.
+          advice.isAdd() && originalCoordinates.isKmpTargetThatShouldDeferToParent(advice.toConfiguration, removedKmpCanonicalDeps) -> {
```
autonomousapps, line 316: You could use `advice.toConfiguration!!` since we know that's non-null here, and then simplify the method itself. If you're worried about it being _mis-used_, we could add an `error()` check. I'd prefer that for clarity vs the current null-check. As a possible alternative, you could pass in the advice itself, in which case the `!!` would be even more clear.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -310,6 +401,27 @@ internal class DependencyAdviceBuilder(
...
+   *
+   * @see GradleVariantIdentification.kmpAttribute
+   */
+  private fun Coordinates.isKmpTargetThatShouldDeferToParent(
+    targetConfiguration: String?,
+    kmpCommonDeclarations: Map<String, Set<String>>
+  ): Boolean {
+    // This only applies to module coordinates.
+    if (this !is ModuleCoordinates) return false
+    if (targetConfiguration == null) return false
```
autonomousapps, line 417: As I indicated in a comment above, I find this null check a bit confusing out of context.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -310,6 +401,27 @@ internal class DependencyAdviceBuilder(
...
+    if (isKmpTargetTarget) {
```
autonomousapps, line 418: Is this a typo that we're using `Target` twice in a row?

autonomousapps (changes requested):
Left some fairly minor comments on areas I think could be improved.

src/main/kotlin/com/autonomousapps/tasks/ComputeUsagesTask.kt
ZacSweers: We do yeah 👍

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
...
+
+    dependencyUsages.asSequence()
       .flatMap { (coordinates, usages) ->
         StandardTransform(coordinates, declarations, supportedSourceSets, buildPath).reduce(usages).map { it to coordinates }
       }
+      .forEach { (advice, originalCoordinates) ->
+        when {
+          advice.isAdd() && originalCoordinates is ModuleCoordinates && originalCoordinates.isKmpTargetTarget -> {
+            deferredKmpAdvice += advice to originalCoordinates
+            addedKmpTargets.getOrPut(advice.toConfiguration!!, ::mutableSetOf).add(originalCoordinates.kmpCommonParentIdentifier())
```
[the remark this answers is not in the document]
ZacSweers: We do care about the return value, we're using it (it's a mutable set) as we call `.add()` on it right after.

It's not exactly the same as `map.merge()`. You could use it to accomplish the same thing, but it's more verbose and allocates a new intermediary collection unnecessarily

```kotlin
addedKmpTargets.merge(config, mutableSetOf(originalCoordinates.kmpCommonParentIdentifier())) { existing, new ->
    existing.addAll(new)
}
```

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
...
+            deferredNonKmpAdvice += advice to originalCoordinates
+          }
+        }
+      }
+
+    val modifiedKmpAdvice = deferredKmpAdvice
+      // "null" removes the advice
+      .mapNotNull { (advice, originalCoordinates) ->
+        when {
+          // KMP target dependencies ("<dep>-android", "<dep>-jvm", etc) should defer to using their parent comment dep
```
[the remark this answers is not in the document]
ZacSweers: oops should be common

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: I'm not sure what you mean?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
 
   private fun computeDependencyAdvice(projectPath: String): Sequence<Advice> {
     val declarations = declarations.filterToSet { Configurations.isForRegularDependency(it.configurationName) }
-    return dependencyUsages.asSequence()
+
+    /*
+     * To support KMP properly, we need to do an initial pass at deps to clean up any KMP deps advice.
```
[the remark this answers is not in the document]
ZacSweers: [96e7e6d](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/96e7e6d4afd97fee45ab02c748d5bd3a1a52c062)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
...
+     *
+     * The primary case we are trying to catch and avoid here is one where we "fix" canonical KMP deps by trying to replace
```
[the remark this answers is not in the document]
ZacSweers: [8e36341](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/8e36341cbea3b5e207d61ad47e1c7aa1852da2a6)

src/main/kotlin/com/autonomousapps/model/GradleVariantIdentification.kt
```diff
@@ -39,4 +43,39 @@ data class GradleVariantIdentification(
...
+ * KMP artifacts have a [KotlinPlatformType] attribute defined for which target it is.
+ *
+ * At a general level, these can be thought of as a canonical proxy dependency and "the rest", which
+ * covers specific platform targets like [KotlinPlatformType.jvm], [KotlinPlatformType.native], etc.
+ *
+ * When inspecting attributes for the context of DAGP, we want to know if a given dependency is a canonical
+ * dependency or a specific target. When recommending advice, we want to defer to the canonical deps as
+ * a sort of implicit bundle of its other targets and will resolve to the correct target.
+ */
+internal val Coordinates.kmpAttribute get() = gradleVariantIdentification.attributes[KotlinPlatformType.attribute.name]
```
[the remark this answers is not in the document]
ZacSweers: [db44960](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/db44960ae60ffb8c62132609b41fc8dc8ba43e0e)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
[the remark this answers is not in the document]
ZacSweers: [6ad22cb](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/6ad22cbaf8f99a04931f9fa1c5734b8b87b7b092)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -254,10 +263,91 @@ internal class DependencyAdviceBuilder(
...
+          // If the parent _isn't_ present in the original declarations though, then accept it and assume they're
+          // intentionally only depending on that specific target's artifact.
+          advice.isAdd() && originalCoordinates.isKmpTargetThatShouldDeferToParent(advice.toConfiguration, removedKmpCanonicalDeps) -> {
+            null
+          }
+          // This is a "misused" dep, but we still want it to use the KMP parent type rather than the targeted subtype
+          advice.isAdd() && isKotlinPluginApplied && originalCoordinates is ModuleCoordinates && originalCoordinates.isKmpTargetTarget -> {
+            val newIdentifier = originalCoordinates.kmpCommonParentIdentifier()
+            val newCoordinates = originalCoordinates.copy(
+              identifier = originalCoordinates.kmpCommonParentIdentifier(),
```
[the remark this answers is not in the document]
ZacSweers: [0bfac35](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/0bfac35ee1c048455431fd44c983cc9b0cc4eced)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: [9580098](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/958009802ed7360f0ef74239fd32c9d0b45bc603)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: [044613e](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/044613e44adcfd86a6f57abdb010bdbeaf596896)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: [044613e](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/pull/919/commits/044613e44adcfd86a6f57abdb010bdbeaf596896)

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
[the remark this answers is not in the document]
autonomousapps: I see, I missed the `.add(...)` call in the earlier review 👍

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
autonomousapps: the name of this variable is `isKmpTargetTarget`. The word "Target" is repeated.

ZacSweers: image

autonomousapps (approved):

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
ZacSweers: [7a31c83](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/919/commits/7a31c832b6239b17feb0b3a1bee27b61835be5d5)

PR #920, opened 2023-07-04
Add gradle property to disable compatibility mismatch

### LINKED ISSUE #801: Move AGP check warning to task execution time instead of configuration time.
```
> Configure project :
The Dependency Analysis plugin is only known to work with versions of AGP between 4.2.2 and 7.4.0-alpha07. You are using 8.0.0-alpha05. Proceed at your own risk.
```
is printed during the configuration phase of the project.

Please move this to task runtime, so we can apply this plugin to our project without everyone getting these warnings.

If you are not OK with that, I'd be happy with a suppression gradle property or DSL method. WDYT?

ZacSweers:
This adds a new property to control the compatibility warning log. There are three levels: `NONE`, `DEBUG`, `WARN`, `ERROR`. These correspond to their logger visibilities.

```properties
dependency.analysis.compatibility=NONE
```

Resolves #801

### DIFF
--- a/src/main/kotlin/com/autonomousapps/Flags.kt
+++ b/src/main/kotlin/com/autonomousapps/Flags.kt
@@ -50,7 +50,8 @@ object Flags {
 
   internal fun Project.compatibility(): Compatibility {
     return getGradlePropForConfiguration(FLAG_DISABLE_COMPATIBILITY, Compatibility.WARN.name).let {
-      Compatibility.valueOf(it.toUpperCase(Locale.US))
+      val value = it.toUpperCase(Locale.US)
+      Compatibility.values().find { it.name == value } ?: error("Unrecognized value '$it' for 'dependency.analysis.compatibility' property. Allowed values are ${Compatibility.values()}")
     }
   }
 
--- a/src/main/kotlin/com/autonomousapps/Flags.kt
+++ b/src/main/kotlin/com/autonomousapps/Flags.kt
@@ -75,6 +75,6 @@ object Flags {
       .toBoolean()
 
   internal enum class Compatibility {
-    NONE, WARN, ERROR
+    NONE, DEBUG, WARN, ERROR
   }
 }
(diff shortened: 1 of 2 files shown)


autonomousapps (changes requested):
PTAL at https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/Flags.kt and use the pattern there. Thanks!

ZacSweers:
Updated and made it more configurable than just a boolean on/off switch

ZacSweers:
Adding a DEBUG and disambiguation of NONE. Planning to make NONE work as a total no-op.

src/main/kotlin/com/autonomousapps/Flags.kt
```diff
@@ -45,6 +48,12 @@ object Flags {
       .getOrElse(default)
   }
 
+  internal fun Project.compatibility(): Compatibility {
+    return getGradlePropForConfiguration(FLAG_DISABLE_COMPATIBILITY, Compatibility.WARN.name).let {
+      Compatibility.valueOf(it.toUpperCase(Locale.US))
```
autonomousapps, line 53: Are we ok just throwing an `IllegalArgumentException` if the user passes a bad value or do you think this warrants a nicer error message?

autonomousapps (commented):
I'm feeling much better about this vs the original version. Left a question about exception handling. Also, what do you feel about adding a test for this? Or is it not important enough to verify? I could go either way, since it's not critical functionality and the code is pretty straightforward.

src/main/kotlin/com/autonomousapps/Flags.kt
ZacSweers:
```suggestion
      val value = it.toUpperCase(Locale.US)
      Compatibility.values().find { it.name == value } ?: error("Unrecognized value '$it' for 'dependency.analysis.compatibility' property. Allowed values are ${Compatibility.values()}")
```

ZacSweers:
I don't think this necessarily warrants a test :)

autonomousapps (approved):
LGTM. Thanks!

PR #922, opened 2023-07-05
Deprecate dependencyAnalysis.dependencies for structure

### LINKED ISSUE #891: Candidate resolution DSL issue configuring dependencies
**Build scan link**
`gradlew --scan buildHealth`: https://scans.gradle.com/s/nl5ukhfciwirw

**Plugin version**
1.20.0

**Gradle version**
8.1.1

**IDE version**
IntelliJ Ultimate 2023.1

**Verified platforms**
Mac and Windows

**Describe the bug**
I'm getting an IDE warning and subsequent errors trying to [configure dependency bundles](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/wiki/Customizing-plugin-behavior#logical-dependency-groups-aka-dependency-bundles):

```kotlin
dependencyAnalysis {
	dependencies { // Candidate resolution warning (full text below)
		bundle("kotlinx-coroutines-core") { // Unresolved reference: bundle
			includeDependency("org.jetbrains.kotlinx:kotlinx-coroutines-core") // Unresolved reference: includeDependency
			includeDependency("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm") // Same unresolved reference
		}
	}
}
```

> Candidate resolution will be changed soon, please use fully qualified name to invoke the following closer candidate explicitly:
public final fun dependencies(action: Action<DependenciesHandler>): Unit defined in com.autonomousapps.DependencyAnalysisExtension

dagp-dependencies-dsl

**To Reproduce**
Reproducer: https://github.com/sdkotlin/sd-kotlin-spring-talks/tree/dagp-dependencies-configuration
Attempted change: https://github.com/sdkotlin/sd-kotlin-spring-talks/commit/5546c9f3f0178cdf2881ebd4b8a3b398b5c8df96

**Expected behavior**
There are no IDE warnings, and "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm" isn't reported as an unused dependency given the bundle configuration.

**Actual behavior**
Running `buildHealth` does succeed, but the bundle configuration appears to have no effect:

> Unused dependencies which should be removed:
>  integrationTestImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.6.4")

**Additional context**
Perhaps this is just another Gradle 8 incompatibility, #851.

ZacSweers:
Resolves #891. Open to suggestions on the name, this felt the most semantically similar

### DIFF
--- a/sandbox-sample/build.gradle
+++ b/sandbox-sample/build.gradle
@@ -28,7 +28,7 @@ dependencyAnalysis {
       }
     }
   }
-  dependencies {
+  structure {
     bundle('hilt') {
       include('com.google.dagger:hilt-android')
       include('com.google.dagger:hilt-core')
--- a/src/functionalTest/groovy/com/autonomousapps/android/DependenciesSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/DependenciesSpec.groovy
@@ -32,7 +32,7 @@ final class DependenciesSpec extends AbstractAndroidSpec {
     given:
     def additions = """\
       dependencyAnalysis {
-        dependencies {
+        structure {
           bundle('three-ten') {
             includeDependency('org.threeten:threetenbp')
             includeDependency('com.jakewharton.threetenabp:threetenabp')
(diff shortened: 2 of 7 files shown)


autonomousapps:
Is this really necessary? Would it also fix the issue if we used `@DslMarker`? Or is that even possible given that it's Gradle that generates the DSL at runtime from its public types?

ZacSweers:
I think it is necessary, and don't think DslMarker would help as this is a resolution order in groovy afaik

PR #924, opened 2023-07-07
Only share InMemoryCache across builds if the same classloader was used to load the plugin

jjohannes:
Follow up to #914 

Users can run into https://github.com/gradle/gradle/issues/17559 is the plugin is loaded with different classloaders in different builds. In these cases we now keep the old behavior of having separate caches for separate builds.

There is now a test checking the two different cases.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithDivergingPluginClasspathsProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithDivergingPluginClasspathsProject.groovy
@@ -1,0 +1,43 @@
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Plugin
+
+final class IncludedBuildWithDivergingPluginClasspathsProject extends AbstractProject {
+
+  final GradleProject gradleProject
+
+  IncludedBuildWithDivergingPluginClasspathsProject(boolean divergingPluginClasspaths) {
+    this.gradleProject = build(divergingPluginClasspaths)
+  }
+
+  private GradleProject build(boolean divergingPluginClasspaths) {
+    def builder = newGradleProjectBuilder()
+    def printServiceObject = """\
+      afterEvaluate { // needs 'afterEvaluate' because plugin code also uses 'afterEvaluate'
+        tasks.named('explodeJarMain') {
+          doLast { println(inMemoryCache.get()) }
+        }
+      }
+    """
+    builder.withRootProject { root ->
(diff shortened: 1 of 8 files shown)


src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -58,8 +58,6 @@ internal class ProjectPlugin(private val project: Project) {
     }
   }
 
-  private lateinit var inMemoryCacheProvider: Provider<InMemoryCache>
```
jjohannes, line 61: Removed to not register too early (i.e. before all builds got the chance to apply the plugin).

Also, this additional state is not necessary. The `registerIfAbsent` of Gradle keeps care of holding the already registered service and should be optimized for being called many times.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/IncludedBuildWithDivergingPluginClasspathsProject.groovy
```diff
@@ -0,0 +1,43 @@
...
+  IncludedBuildWithDivergingPluginClasspathsProject(boolean divergingPluginClasspaths) {
+    this.gradleProject = build(divergingPluginClasspaths)
+  }
+
+  private GradleProject build(boolean divergingPluginClasspaths) {
+    def builder = newGradleProjectBuilder()
+    def printServiceObject = """\
+      afterEvaluate { // needs 'afterEvaluate' because plugin code also uses 'afterEvaluate'
+        tasks.named('explodeJarMain') {
+          doLast { println(inMemoryCache.get()) }
```
autonomousapps, line 20: clever 😅

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
```diff
@@ -18,6 +18,8 @@ import org.gradle.api.artifacts.Configuration
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.register
 
+const val DEPENDENCY_ANALYSIS_PLUGIN = "com.autonomousapps.dependency-analysis"
```
autonomousapps, line 21: Do we want this public, or is `internal` better?

autonomousapps (approved):
Left one question about visibility.

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
jjohannes: Yes `internal` is good. Thanks.

It's not in my repertoire as "doing a lot of Java" programmer. I tend to forget about it. 😅

autonomousapps (approved):

PR #926, opened 2023-07-11
Allow custom severity and filtering by source set.

autonomousapps:
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/901

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/CustomSourceSetSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/CustomSourceSetSpec.groovy
@@ -1,12 +1,12 @@
+//file:noinspection GroovyAssignabilityCheck
 package com.autonomousapps.jvm
 
-import com.autonomousapps.jvm.projects.FeatureVariantInSameProjectTestProject
-import com.autonomousapps.jvm.projects.FeatureVariantTestProject
-import com.autonomousapps.jvm.projects.IntegrationTestProject
-import com.autonomousapps.jvm.projects.TestFixturesTestProject
-import com.autonomousapps.jvm.projects.TestFixturesTestProject2
+import com.autonomousapps.jvm.projects.*
+import org.gradle.util.GradleVersion
 
+import static com.autonomousapps.jvm.projects.SourceSetFilteringProject.Severity.*
 import static com.autonomousapps.utils.Runner.build
+import static com.autonomousapps.utils.Runner.buildAndFail
 import static com.google.common.truth.Truth.assertThat
 
 final class CustomSourceSetSpec extends AbstractJvmSpec {
@@ -130,5 +130,41 @@ final class CustomSourceSetSpec extends AbstractJvmSpec {
 
     where:
     gradleVersion << gradleVersions()
+  }
+
+  def "custom source set analysis can be fine-filtered for severity=#severity (#gradleVersion)"() {
+    given:
+    def project = new SourceSetFilteringProject.Filtering(severity)
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'buildHealth', '-Pdependency.analysis.print.build.health=true')
+
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth())
+
+    where:
+    [gradleVersion, severity] << multivariableDataPipe(
+      gradleVersions(),
+      [IGNORE, WARN]
+    )
+  }
+
+  def "custom source set analysis can be layered: severity=#severity (#gradleVersion)"() {
+    given:
+    def project = new SourceSetFilteringProject.Layering(severity)
+    gradleProject = project.gradleProject
+
+    when:
+    buildAndFail(gradleVersion, gradleProject.rootDir, 'buildHealth', '-Pdependency.analysis.print.build.health=true')
+
+    then:
+    assertThat(project.actualBuildHealth()).containsExactlyElementsIn(project.expectedBuildHealth())
+
+    where:
+    [gradleVersion, severity] << multivariableDataPipe(
+      gradleVersions(),
+      [IGNORE, FAIL]
+    )
   }
 }
--- a/src/main/kotlin/com/autonomousapps/extension/Behavior.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/Behavior.kt
@@ -2,7 +2,11 @@ package com.autonomousapps.extension
 
 import java.io.Serializable
 
-sealed class Behavior(val filter: Set<String> = setOf()) : Serializable, Comparable<Behavior> {
+sealed class Behavior(
+  val filter: Set<String> = setOf(),
+  val sourceSetName: String = Issue.ALL_SOURCE_SETS
+) : Serializable, Comparable<Behavior> {
+
   /**
    * [Fail] > [Ignore] > [Warn] > [Undefined].
    */
@@ -27,7 +31,8 @@ sealed class Behavior(val filter: Set<String> = setOf()) : Serializable, Compara
 
       is Warn -> {
         when (this) {
-          is Fail, Ignore -> 1
+          is Fail -> 1
+          is Ignore -> 1
           is Warn -> 0
           is Undefined -> -1
         }
@@ -36,7 +41,32 @@ sealed class Behavior(val filter: Set<String> = setOf()) : Serializable, Compara
   }
 }
 
-class Fail(filter: Set<String> = mutableSetOf()) : Behavior(filter)
-class Warn(filter: Set<String> = mutableSetOf()) : Behavior(filter)
-object Ignore : Behavior()
-class Undefined(filter: Set<String> = mutableSetOf()) : Behavior(filter)
+class Fail(
+  filter: Set<String> = mutableSetOf(),
+  sourceSetName: String = Issue.ALL_SOURCE_SETS
+) : Behavior(
+  filter = filter,
+  sourceSetName = sourceSetName
+)
+
+class Warn(
+  filter: Set<String> = mutableSetOf(),
+  sourceSetName: String = Issue.ALL_SOURCE_SETS
+) : Behavior(
+  filter = filter,
+  sourceSetName = sourceSetName
+)
+
+class Ignore(
+  sourceSetName: String = Issue.ALL_SOURCE_SETS
+) : Behavior(
+  sourceSetName = sourceSetName
+)
+
+class Undefined(
+  filter: Set<String> = mutableSetOf(),
+  sourceSetName: String = Issue.ALL_SOURCE_SETS
+) : Behavior(
+  filter = filter,
+  sourceSetName = sourceSetName
+)
--- a/src/main/kotlin/com/autonomousapps/extension/Issue.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/Issue.kt
@@ -2,10 +2,11 @@
 
 package com.autonomousapps.extension
 
-import org.gradle.api.GradleException
+import org.gradle.api.InvalidUserDataException
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
 import org.gradle.api.provider.SetProperty
+import org.gradle.kotlin.dsl.property
 import org.gradle.kotlin.dsl.setProperty
 import javax.inject.Inject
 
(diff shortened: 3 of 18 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/CustomSourceSetSpec.groovy
```diff
@@ -131,4 +128,22 @@ final class CustomSourceSetSpec extends AbstractJvmSpec {
     where:
     gradleVersion << gradleVersions()
   }
+
+  def "custom source set analysis can be fine-filtered for severity=#severity (#gradleVersion)"() {
```
autonomousapps, line 132: More thorough testing might be warranted.

src/main/kotlin/com/autonomousapps/extension/Behavior.kt
```diff
@@ -2,7 +2,11 @@ package com.autonomousapps.extension
 
 import java.io.Serializable
 
-sealed class Behavior(val filter: Set<String> = setOf()) : Serializable, Comparable<Behavior> {
+sealed class Behavior(
+  val filter: Set<String> = setOf(),
+  val sourceSetName: String = Issue.ALL_SOURCE_SETS
```
autonomousapps, line 7: The core of the new implementation is embedding source set information in the model here.

autonomousapps:
Build failed

src/main/kotlin/com/autonomousapps/extension/Issue.kt
```diff
@@ -27,26 +28,28 @@ import javax.inject.Inject
...
   fun severity(value: String) {
     when (value) {
       "warn" -> severity.set(Warn())
       "fail" -> severity.set(Fail())
-      "ignore" -> severity.set(Ignore)
-      else -> throw GradleException(
-        "'value' is not a recognized behavior. Must be one of 'warn', 'fail', or 'ignore'"
+      "ignore" -> severity.set(Ignore())
+      else -> throw InvalidUserDataException(
+        "'$value' is not a recognized behavior. Must be one of 'warn', 'fail', or 'ignore'"
```
autonomousapps, line 52: This was a longstanding error to not use `$`.

src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
```diff
@@ -65,66 +69,104 @@ open class IssueHandler @Inject constructor(objects: ObjectFactory) {
...
-  internal fun unusedAnnotationProcessorsIssueFor(path: String): Provider<Behavior> {
-    val global = all.unusedAnnotationProcessorsIssue
-    val proj = projects.findByName(path)?.unusedAnnotationProcessorsIssue
-    return overlay(global, proj)
+  internal fun moduleStructureIssueFor(projectPath: String): Provider<Behavior> {
+    return overlay(all.moduleStructureIssue, projects.findByName(projectPath)?.moduleStructureIssue)
+  }
+
+  private fun issuesFor(projectPath: String, mapper: (ProjectIssueHandler) -> Issue): List<Provider<Behavior>> {
+    val projectHandler = projects.findByName(projectPath)
```
autonomousapps, line 109: The algorithm here is so ugly.

src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
```diff
@@ -23,15 +25,17 @@ import javax.inject.Inject
  */
 open class IssueHandler @Inject constructor(objects: ObjectFactory) {
 
+  private val undefined = objects.newInstance<Issue>()
```
autonomousapps, line 28: Semantically this is a constant, but because we need the object factory to instantiate it, shouldn't use SHOUT_CASE.

src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
```diff
@@ -65,66 +69,104 @@ open class IssueHandler @Inject constructor(objects: ObjectFactory) {
...
-    return overlay(global, proj)
+  private fun issuesBySourceSetFor(project: ProjectIssueHandler?, mapper: (ProjectIssueHandler) -> Issue): List<Issue> {
+    return all.issuesBySourceSet(mapper) + project.issuesBySourceSet(mapper)
   }
 
-  /** Project severity wins over global severity. Excludes are unioned. */
+  /**
+   * Project severity wins over global severity. Excludes are unioned.
+   *
+   * TODO: update kdoc for sourceSets behavior
```
autonomousapps, line 163: TODO.

src/main/kotlin/com/autonomousapps/extension/ProjectIssueHandler.kt
```diff
@@ -54,21 +55,26 @@ import javax.inject.Inject
...
-  internal val anyIssue = objects.newInstance(Issue::class.java)
-  internal val unusedDependenciesIssue = objects.newInstance(Issue::class.java)
-  internal val usedTransitiveDependenciesIssue = objects.newInstance(Issue::class.java)
-  internal val incorrectConfigurationIssue = objects.newInstance(Issue::class.java)
-  internal val unusedAnnotationProcessorsIssue = objects.newInstance(Issue::class.java)
-  internal val compileOnlyIssue = objects.newInstance(Issue::class.java)
-  internal val runtimeOnlyIssue = objects.newInstance(Issue::class.java)
-  internal val redundantPluginsIssue = objects.newInstance(Issue::class.java)
-  internal val moduleStructureIssue = objects.newInstance(Issue::class.java)
+  internal val sourceSets = objects.domainObjectContainer(
```
autonomousapps, line 64: I'm deliberately not documenting this because I don't want to encourage the general population to use the new API.

src/main/kotlin/com/autonomousapps/extension/ProjectIssueHandler.kt
```diff
@@ -86,6 +92,17 @@ open class ProjectIssueHandler @Inject constructor(
     ignoreSourceSets.addAll(ignore.toSet())
   }
 
+  /**
+   * Specify custom behavior for [sourceSetName] using `this` [ProjectIssueHandler].
+   *
+   * TODO: fix kdoc.
```
autonomousapps, line 98: TODO.

src/main/kotlin/com/autonomousapps/extension/SourceSetsHandler.kt
```diff
@@ -0,0 +1,41 @@
+package com.autonomousapps.extension
+
+import org.gradle.api.Named
+import org.gradle.api.NamedDomainObjectFactory
+import org.gradle.api.model.ObjectFactory
+import javax.inject.Inject
+
+/**
+ * TODO: kdoc.
```
autonomousapps, line 9: TODO.

src/main/kotlin/com/autonomousapps/extension/SourceSetsHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+   * Gets the specified issue out of [project], and sets [sourceSetName] on it for later filtering by
+   * [FilterAdviceTask][com.autonomousapps.tasks.FilterAdviceTask].
+   */
+  internal fun issueOf(mapper: (ProjectIssueHandler) -> Issue): Issue {
+    return mapper(project).apply {
+      sourceSet.set(sourceSetName)
+    }
+  }
+
+  internal class Factory(
```
autonomousapps, line 31: Not sure I actually need a custom factory, but thinking about how Gradle instantiates custom objects via reflection was giving me a headache.

src/main/kotlin/com/autonomousapps/internal/advice/SeverityHandler.kt
```diff
@@ -3,35 +3,84 @@ package com.autonomousapps.internal.advice
...
 import com.autonomousapps.extension.Behavior
 import com.autonomousapps.extension.Fail
+import com.autonomousapps.internal.utils.filterToSet
+import com.autonomousapps.internal.utils.lowercase
+import com.autonomousapps.model.Advice
 import com.autonomousapps.model.ModuleAdvice
 import com.autonomousapps.model.Advice as DependencyAdvice
 
+/**
+ * TODO: this class needs unit tests.
```
autonomousapps, line 13: todo (little case)

src/main/kotlin/com/autonomousapps/extension/SourceSetsHandler.kt
jjohannes: Generally, I would make all types to be instantiated by Gradle `abstract` instead of `open`. Small practical advantage is that you won't accidentally instantiate them without Gradle's factories.

Then oftentimes you won't need `objects.*()` calls. But you can have `abstract val` that Gradle fills as we do in the tasks here.

However, this automatic stuff does not work with _domain object containers_ and custom types in all cases. I am also not sure when and why. I tend to avoid _domain object containers_ as these are so overloaded in their API (right now). I think there is also no clear statement from Gradle on this. For usability in a DSL, I prefer a clear interface with as little methods as possible. But that's a a different fundamental/philosophical discussion. :)

Anyway, just some background which might already know.

I think in this case, you have to do it like this to pass the `projectPath` to all objects. I don't know of a better way to do it.

src/main/kotlin/com/autonomousapps/extension/SourceSetsHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val projectPath: String,
+    private val objects: ObjectFactory
+  ) : NamedDomainObjectFactory<SourceSetsHandler> {
+    override fun create(name: String): SourceSetsHandler = SourceSetsHandler(
+      sourceSetName = name,
+      projectPath = projectPath,
+      objects = objects
+    )
```
jjohannes, line 59:
```suggestion
    override fun create(name: String): SourceSetsHandler = objects.newInstance(
      SourceSetsHandler::class.java,
      name,
      projectPath
    )
```

Could do it like this, and have Gradle inject `objects` for SourceSetsHandler (which could then also be `abstract`)

src/main/kotlin/com/autonomousapps/extension/IssueHandler.kt
jjohannes: 🤷 if it works. I think as you said, it could be a topic for `2.x` of the plugin to rework this whole part of the DSL so that it also becomes easier to process.

Still I think the PR as ist has now has value.

src/main/kotlin/com/autonomousapps/tasks/FilterAdviceTask.kt
```diff
@@ -177,13 +196,57 @@ abstract class FilterAdviceTask @Inject constructor(
...
+      }
+
+      val bySourceSets: (Advice) -> Boolean = { a ->
+        // These are the source sets represented in this advice. Might be empty if it is for the main source set.
+        val adviceSourceSets = supportedSourceSets
+          .map { it.lowercase() }
+          .filter { s ->
+            val from = a.fromConfiguration?.lowercase()?.startsWith(s) == true
+            val to = a.toConfiguration?.lowercase()?.startsWith(s) == true
+            from || to
```
jjohannes, line 230: Maybe I miss something, but does this work correctly for the `main` source set? Because if `s == ""` it matches everything, also things that **are not in main**.

If you want this to be 100% accurate I think you would have to split the sourceSetName out of the configuration name, by also looking for all the possible suffixes ("Implementation", "Api", "CompileOnly", ...):
1. Match one of the suffixes
2. Strip the suffix
3. The result is the _source set name_ or if `""` it is `main`

I did that here. If you ignore the Java Modules Directive part:

- Have a list of the possible suffixes: https://github.com/gradlex-org/java-module-dependencies/blob/main/src/main/java/org/gradlex/javamodule/dependencies/tasks/ModuleDirectivesScopeCheck.java#L48-L54
- Compute the source set name by first determining which of the the suffixes ("Scope Configuration") is used and then strip that to get the source set name https://github.com/gradlex-org/java-module-dependencies/blob/main/src/main/java/org/gradlex/javamodule/dependencies/tasks/ModuleDirectivesScopeCheck.java#L139-L154

src/main/kotlin/com/autonomousapps/extension/SourceSetsHandler.kt
autonomousapps: I'll give it a try. At the time of writing that code, I remembered being a little worried about the fact that both `name` and `projectPath` are Strings, Gradle would rely on the order of parameters which strikes me as fragile.

src/main/kotlin/com/autonomousapps/tasks/FilterAdviceTask.kt
autonomousapps: How could `s == ""`, though? `s` is an element of the collection `supportedSourceSets`, which in a "normal" case might have values like "main" and "test." The comment above that says "Might be empty if it is for the main source set" means that the _result_ of the map+filter might be empty. (is that comment what you're referring to?)

I considered your approach that you linked, but for some reason having a concrete list of scopes was bothering me :) Any chance Gradle will add yet another possibility? That said, I'll continue changing the implementation to match your preference because it would probably be more readable.

jjohannes: Sorry then I misunderstood. I thought for the "main" source set that list would contain the empty string. But if it contains "main", does it work then? Because the configurations do not start with "main".

Anyway, if this works and I just miss something, then you can also keep the approach you have.

Regarding the list of "scopes": I don't see how that would ever change again. All combinations that makes sense in Java are now covered (since the addition of compileOnlyApi).

jjohannes: Maybe this list already exists somewhere in the Plugin code? In the end, these are also the configurations/scopes for which advices are created in the first place.

autonomousapps: I just pushed a commit that follows your advice and I think the code is much cleaner and easier to follow. And the tests still pass! So thanks :)

autonomousapps:
Rebased and force-pushed. If the tests still pass, will merge on green.

PR #930, opened 2023-07-20
Build and test against Gradle 8.2.1.

autonomousapps:


autonomousapps:
Build failed

autonomousapps:
Tempted to merge this even though I have to ignore `AntlrSpec` to do so. I think that the `antlr` plugin (or Gradle) has a bug that is causing its source directory information to not be properly tracked for task dependency purposes.

PR #938, opened 2023-08-03
Fix addressing a project by path in the reason '--id' argument

jjohannes:
This broke with #916

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
@@ -19,7 +19,7 @@ final class JvmReasonSpec extends AbstractFunctionalSpec {
     def result = build(gradleVersion, gradleProject.rootDir, ':featureC:public:reason', '--id', id)
 
     then:
-    outputMatchesForProject(result)
+    outputMatchesForProject(result, id)
 
     where:
     gradleVersion << gradleVersions()
@@ -34,18 +34,18 @@ final class JvmReasonSpec extends AbstractFunctionalSpec {
     def result = build(gradleVersion, gradleProject.rootDir, ':featureC:public:reason', '--id', id)
 
     then:
-    outputMatchesForProject(result)
+    outputMatchesForProject(result, id)
 
     where:
     gradleVersion << gradleVersions()
   }
 
-  private static void outputMatchesForProject(BuildResult result) {
+  private static void outputMatchesForProject(BuildResult result, String id) {
     def lines = Colors.decolorize(result.output).readLines()
     def asked = lines.find { it.startsWith('You asked about') }
     def advised = lines.find { it.startsWith('There is no advice') }
 
-    assertThat(asked).isEqualTo("You asked about the dependency 'the-project.featureA:public'.")
+    assertThat(asked).isEqualTo("You asked about the dependency '$id'.".toString())
     assertThat(advised).isEqualTo("There is no advice regarding this dependency.")
   }
 
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
@@ -1,0 +1,52 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.AbstractFunctionalSpec
+import com.autonomousapps.jvm.projects.NestedSubprojectsProject
+import com.autonomousapps.utils.Colors
+import org.gradle.testkit.runner.BuildResult
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class JvmReasonSpec extends AbstractFunctionalSpec {
+
+  def "can discover reason for project dependency defined by project path (#gradleVersion)"() {
+    given:
+    gradleProject = new NestedSubprojectsProject().gradleProject
+    def id = ':featureA:public'
+
+    when:
+    def result = build(gradleVersion, gradleProject.rootDir, ':featureC:public:reason', '--id', id)
+
+    then:
+    outputMatchesForProject(result)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  def "can discover reason for project dependency defined by coordinates (#gradleVersion)"() {
+    given:
+    gradleProject = new NestedSubprojectsProject().gradleProject
+    def id = 'the-project.featureA:public'
+
+    when:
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -190,15 +190,27 @@ abstract class ComputeAdviceTask @Inject constructor(
...
+      }.toMap()
+
+    private fun toCoordinatesKey(coordinates: Coordinates, buildPath: String) =
+      coordinates.gav() + when {
+        // If the coordinates point at a project in the current build, there are two notations to address it:
+        // 1) 'group:name' coordinates, which is the main representation we aim to use everywhere (see IncludedBuildCoordinates)
+        // 2) The project path notation starting with a ':' (:project-name)
+        // To allow notation 2) to address a project in the 'reasons' task, we add this coordinate to the key as well if applicable (separated by '|')
+        coordinates is IncludedBuildCoordinates && coordinates.resolvedProject.buildPath == buildPath -> "|${coordinates.resolvedProject.gav()}"
+        else -> ""
```
jjohannes, line 211: In the `usages-*.json` files, the 'id' is the String key. We somehow need to pass along the alternative ID here. Since I did not want to change the whole model there, I now append the alternative 'id' – the project path – to the key separated by `|`.

src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+    outputMatchesForProject(result)
+
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  def "can discover reason for project dependency defined by coordinates (#gradleVersion)"() {
+    given:
+    gradleProject = new NestedSubprojectsProject().gradleProject
+    def id = 'the-project.featureA:public'
```
autonomousapps, line 31: oh neat, that's cool that this works for local projects

src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+    where:
+    gradleVersion << gradleVersions()
+  }
+
+  private static void outputMatchesForProject(BuildResult result) {
+    def lines = Colors.decolorize(result.output).readLines()
+    def asked = lines.find { it.startsWith('You asked about') }
+    def advised = lines.find { it.startsWith('There is no advice') }
+
+    assertThat(asked).isEqualTo("You asked about the dependency 'the-project.featureA:public'.")
```
autonomousapps, line 48: Hm will it always say this? Even if we query using the project path? I'd prefer if the output matched the input (if the user queries for `:featureA:public` then it should say that here).

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -190,15 +190,27 @@ abstract class ComputeAdviceTask @Inject constructor(
...
-    }.toMap()
+    private fun Map<Coordinates, Set<Usage>>.toStringCoordinates(buildPath: String): Map<String, Set<Usage>> =
+      map { (key, value) ->
+        toCoordinatesKey(key, buildPath) to value
+      }.toMap()
+
+    private fun toCoordinatesKey(coordinates: Coordinates, buildPath: String) =
+      coordinates.gav() + when {
+        // If the coordinates point at a project in the current build, there are two notations to address it:
+        // 1) 'group:name' coordinates, which is the main representation we aim to use everywhere (see IncludedBuildCoordinates)
```
autonomousapps, line 207: Do we aim to use it everywhere?

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
autonomousapps: Can we extract this code into a utility class for better encapsulation and also use by `ReasonTask`? Then it would be clear how the `|` works and who uses it.

autonomousapps (commented):
Thanks for addressing this! Left some questions and comments.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
jjohannes: I removed that part of the comment. I am also not sure what I meant to say with that. Thanks.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
jjohannes: Moved all functions related to this into `utils/coordinatesStrings.kt`.

src/functionalTest/groovy/com/autonomousapps/jvm/JvmReasonSpec.groovy
jjohannes: Yes that's probably better. We need to trace what we requested then to use that information when we create the report. I added that.

jjohannes:
Thanks @autonomousapps. I addressed your feedback.

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -174,13 +174,15 @@ abstract class ReasonTask @Inject constructor(
 
     // Derived from the above
     private val finalAdvice by unsafeLazy { findAdviceIn(finalProjectAdvice) }
-    private val coord by unsafeLazy { getRequestedCoordinates() }
+    private val requestedCoord by unsafeLazy { getRequestedCoordinates(false) }
+    private val coord by unsafeLazy { getRequestedCoordinates(true) }
```
autonomousapps, line 178: Interesting, thanks!

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -205,12 +211,19 @@ abstract class ReasonTask @Inject constructor(
...
         ?: annotationProcessorUsages.entries.find(requestedId::equalsKey)?.key
         ?: annotationProcessorUsages.entries.find(requestedId::startsWithKey)?.key
         ?: findInGraph()
         ?: throw InvalidUserDataException("There is no dependency with coordinates '$requestedId' in this project.")
+
+      val gav = if (requestedViaProjectPath && !normalize) {
+        gavKey.secondCoordinatesKeySegment() ?: gavKey
+      } else {
+        gavKey.firstCoordinatesKeySegment()
+      }
```
autonomousapps, line 225: 👍

autonomousapps (approved):
LGTM!

PR #943, opened 2023-08-18
Test against AGP 8.1.0 as latest supported stable version

joshfriend:
I had some test failures running this locally that looked like flakes (failed running all of `functionalTest` but pass when running individually in IDE), lets see if CI agrees


autonomousapps:
Thanks!

PR #959, opened 2023-09-13
Failing test.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/952


src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -78,6 +80,7 @@ internal class Bundles(private val dependencyUsages: Map<Coordinates, Set<Usage>
     ): Bundles {
       val bundles = Bundles(dependencyUsages)
 
+      // TODO verify that primaries still work
```
autonomousapps, line 83: TODO. It's possible that `primary()` still doesn't work for projects.

PR #961, opened 2023-09-15
Add exclude method for version catalog dependencies

Vichy97:
This pr adds a version of `exclude` that accepts version catalog dependencies. I didn't see any existing tests for this file, but let me know if there is somewhere I should be adding tests for this.

https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/946

### DIFF
--- a/src/main/kotlin/com/autonomousapps/extension/Issue.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/Issue.kt
@@ -3,9 +3,9 @@
 package com.autonomousapps.extension
 
 import org.gradle.api.InvalidUserDataException
+import org.gradle.api.artifacts.MinimalExternalModuleDependency
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
-import org.gradle.api.provider.SetProperty
 import org.gradle.kotlin.dsl.property
 import org.gradle.kotlin.dsl.setProperty
 import javax.inject.Inject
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/extension/Issue.kt
```diff
@@ -63,6 +64,18 @@ open class Issue @Inject constructor(objects: ObjectFactory) {
...
   }
 
+  /**
+   * All provided elements will be filtered out of the final advice. For example:
+   * ```
+   * exclude(libs.example, libs.some.thing)
+   * ```
+   * tells the plugin to exclude those dependencies in the final advice.
+   */
+  fun exclude(vararg ignore: Provider<MinimalExternalModuleDependency>) {
```
autonomousapps, line 74: could we have this call the existing `exclude()` method just above instead? Might make maintenance slightly easier if we ever have to change the above method.

Vichy97: Done!

autonomousapps:
Sorry, apparently there's now a merge conflict. Resolve that and I can merge this.

autonomousapps (approved):

src/main/kotlin/com/autonomousapps/extension/Issue.kt
```diff
@@ -3,9 +3,9 @@
 package com.autonomousapps.extension
 
 import org.gradle.api.InvalidUserDataException
+import org.gradle.api.artifacts.MinimalExternalModuleDependency
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
-import org.gradle.api.provider.SetProperty
```
Vichy97, line 8: After rebasing this import was unused. I think maybe it was missed in a previous refactor?

autonomousapps (approved):

PR #964, opened 2023-09-26
Fix issues with sourceset-based analysis.

autonomousapps:
Addresses multiple issues around sourceset-based analysis:
1. When analyzing the `test` sourceset, don't suggest "upgrading" dependency declarations to something `api`-like. That sourceset has no ABI; it is not published.
2. When analyzing sourcesets that extend other sourcesets, don't redundantly suggest adding dependencies to them that are inherited from their upstream sourcesets.

### DIFF
--- a/graph-support/src/main/kotlin/com/autonomousapps/graph/Graphs.kt
+++ b/graph-support/src/main/kotlin/com/autonomousapps/graph/Graphs.kt
@@ -15,6 +15,14 @@ public object Graphs {
     }
   }
 
+  /**
+   * Returns the first node it finds that has an in-degree of 0. This is the root node if this DAG contains only one
+   * such node.
+   */
+  public fun <N : Any> Graph<N>.root(): N = nodes().first {
+    inDegree(it) == 0
+  }
+
   public fun <N : Any> Graph<N>.parents(node: N): Set<N> = predecessors(node)
 
   public fun <N : Any> Graph<N>.children(node: N): Set<N> = successors(node)
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -24,7 +24,7 @@ pluginManagement {
     }
   }
   plugins {
-    id("com.autonomousapps.dependency-analysis") version "1.10.0"//latestSnapshot
+    id("com.autonomousapps.dependency-analysis") version "1.22.0"//latestSnapshot
     id("com.github.johnrengelman.shadow") version "7.1.2"
     id("com.gradle.enterprise") version "3.10.2"
     id("com.gradle.plugin-publish") version "1.1.0"
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -25,7 +25,7 @@ pluginManagement {
   }
   plugins {
     id("com.autonomousapps.dependency-analysis") version "1.22.0"//latestSnapshot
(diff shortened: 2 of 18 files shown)


graph-support/src/main/kotlin/com/autonomousapps/graph/Graphs.kt
```diff
@@ -15,6 +15,14 @@ public object Graphs {
     }
   }
 
+  /**
+   * Returns the first node it finds that has an in-degree of 0. This is the root node if this DAG contains only one
+   * such node.
+   */
+  public fun <N : Any> Graph<N>.root(): N = nodes().first {
```
autonomousapps, line 22: Required publishing a new version of graph-support (at 0.2).

settings.gradle.kts
```diff
@@ -24,8 +24,8 @@ pluginManagement {
     }
   }
   plugins {
-    id("com.autonomousapps.dependency-analysis") version "1.10.0"//latestSnapshot
-    id("com.github.johnrengelman.shadow") version "7.1.2"
+    id("com.autonomousapps.dependency-analysis") version latestSnapshot//"1.22.0"//latestSnapshot
```
autonomousapps, line 27: - [ ] Cleanup.

src/functionalTest/groovy/com/autonomousapps/jvm/AbstractJvmSpec.groovy
```diff
@@ -12,7 +12,7 @@ abstract class AbstractJvmSpec extends AbstractFunctionalSpec {
   @SuppressWarnings('unused')
   def cleanup() {
     if (gradleProject != null && shouldClean) {
-      clean(gradleProject.rootDir)
+//      clean(gradleProject.rootDir)
```
autonomousapps, line 15: - [ ] TODO

testkit/src/main/kotlin/com/autonomousapps/kit/BuildScript.kt
```diff
@@ -49,12 +49,12 @@ class BuildScript(
...
     val add =
       if (additions.isNotEmpty()) {
-        "\n$additions"
+        "\n$additions\n"
       } else {
         ""
       }
 
-    return buildscriptBlock + pluginsBlock + reposBlock + androidBlock + sourceSetsBlock + featureVariantsBlock + dependenciesBlock + add
+    return buildscriptBlock + pluginsBlock + add + reposBlock + androidBlock + sourceSetsBlock + featureVariantsBlock + dependenciesBlock
```
autonomousapps, line 57: sometimes build script additions should come before other stuff. This might break some tests, won't be sure till I run them. Might need to make this more configurable.

autonomousapps:
Build failed

testkit/src/main/kotlin/com/autonomousapps/kit/BuildScript.kt
autonomousapps: I think what I actually need to do is at least a combination of the following:
1. Expose the ability to just write out raw text as a build script for full and transparent configurability
2. Enhance the existing model rather than shoving a bunch of new things into `BuildScript.additions` as raw text.

PR #967, opened 2023-09-28
Opt-out of configuration cache in `FindDeclaredProcsTask.kt`

SimonMarquis:
This should fix the corresponding issue:

```
> Task :Foo:findDeclaredProcsMain FAILED

* What went wrong:
Execution failed for task ':Foo:findDeclaredProcsMain'.
> Cannot invoke "kotlin.Lazy.getValue()" because "<local1>" is null
```

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
@@ -2,6 +2,7 @@ package com.autonomousapps.tasks
 
 import com.autonomousapps.TASK_GROUP_DEP_INTERNAL
 import com.autonomousapps.internal.ANNOTATION_PROCESSOR_PATH
+import com.autonomousapps.internal.GradleVersions
 import com.autonomousapps.internal.unsafeLazy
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/tasks/FindDeclaredProcsTask.kt
```diff
@@ -55,6 +56,11 @@ abstract class FindDeclaredProcsTask : DefaultTask() {
   init {
     group = TASK_GROUP_DEP_INTERNAL
     description = "Produces a report of all supported annotation types and their annotation processors"
+
+    if (GradleVersions.isAtLeastGradle74) {
+      @Suppress("LeakingThis")
+      notCompatibleWithConfigurationCache("Cannot serialize unsafeLazy")
```
SimonMarquis, line 62: Or could be renamed to `InMemoryCache`
```suggestion
      notCompatibleWithConfigurationCache("Cannot serialize InMemoryCache")
```

autonomousapps:
We're currently working towards making the plugin CC-compatible, so I don't think I want to go in this direction.

SimonMarquis:
Good news 😃

PR #985, opened 2023-10-05
Lower min supported version of Gradle to 7.4 from 7.5.

autonomousapps:
One non-critical task still requires 7.5, but it is not part of buildHealth.


autonomousapps:
Build failed

PR #995, opened 2023-10-10
Use BuildIdentifier.getBuildPath() starting with Gradle 8.2 (second PR)

jjohannes:
Follow up to #949

I totally missed these other usages. They did not show up in the builds I tested with and I forgot about them. Sorry for updating this inconsistently in #949.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
@@ -46,8 +46,9 @@ private fun ResolvedDependencyResult.compositeRequest(): IncludedBuildCoordinate
   val resolved = ProjectCoordinates(
     identifier = (selected.id as ProjectComponentIdentifier).identityPath(),
     gradleVariantIdentification = gradleVariantIdentification,
-    // FIXME use 'buildState.buildIdentifier.buildPath' with Gradle 8.2+
-    buildPath = (selected.id as ProjectComponentIdentifier).build.name
+    buildPath = (selected.id as ProjectComponentIdentifier).build.let {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -46,8 +46,9 @@ private fun ResolvedDependencyResult.compositeRequest(): IncludedBuildCoordinate
   val resolved = ProjectCoordinates(
     identifier = (selected.id as ProjectComponentIdentifier).identityPath(),
     gradleVariantIdentification = gradleVariantIdentification,
-    // FIXME use 'buildState.buildIdentifier.buildPath' with Gradle 8.2+
-    buildPath = (selected.id as ProjectComponentIdentifier).build.name
+    buildPath = (selected.id as ProjectComponentIdentifier).build.let {
+      if (GradleVersions.isAtLeastGradle82) it.buildPath else @Suppress("DEPRECATION") it.name
```
autonomousapps, line 50: Are we ok with the fact that Gradle 8.2+ uses `:second-build`, and otherwise it's `second-build` (no `:` prefix)?

autonomousapps (commented):
Left a question. If the answer is "yes", then LGTM.

src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
jjohannes: Yes. It is fine as long as we consistently use the same API in one Gradle version. We do not care much about the concrete values, but only about comparing the values to figure out if a (sub)project belongs to a certain build – can be depended on via `project(...)` – or not.

autonomousapps (approved):

PR #1011, opened 2023-10-30
Better API for creating included builds.

autonomousapps:


testkit/src/main/kotlin/com/autonomousapps/kit/gradle/BuildScript.kt
```diff
@@ -28,11 +28,22 @@ public class BuildScript(
       appendLine(scribe.use { s -> plugins.render(s) })
     }
 
+    // TODO push this into a new model type
```
autonomousapps, line 31: Sooner than later, this is gross.

PR #1039, opened 2023-11-17
Make graph view task cc compatible

cobexer:
* Update GraphViewTask to use the CC compatible CCGraphViewBuilder
* promote CCGraphViewBuilder to be the only GraphViewBuilder

Hey @autonomousapps,
@ljacomet and I made `projectHealth` CC compatible and tested it with gradle/gradle and we confirmed that we get CC hits with the changes from this PR.


cobexer:
Hey @autonomousapps is there anything holding this back from getting merged and shipped?

autonomousapps:
> Hey @autonomousapps is there anything holding this back from getting merged and shipped?

I think we also need to bump the minimum version of Gradle required to use this plugin to 7.5, right? It's currently at 7.4 (to account for a big user that has that as their current version), and there's also at least one check in `ProjectPlugin` for Gradle version. In addition to changing those things, we'd need to ensure that the functional tests all run against 7.5+ (and not 7.4+).

autonomousapps:
I thought that one of the APIs for CC-safe dependency graph input required Gradle 7.5? if I'm wrong, that's great.

autonomousapps (approved):

autonomousapps:
Note that there is still at least one issue with using the configuration cache. In the `GenerateBuildHealthTask`, there's this task input:
```kotlin
  @Transient
  @get:PathSensitive(PathSensitivity.RELATIVE)
  @get:InputFiles
  lateinit var projectHealthReports: Configuration
```
That task is marked with `notCompatibleWithConfigurationCache("Cannot serialize Configurations")`, as is `ComputeDuplicateDependenciesTask` (although this latter is not actually part of `buildhealth`).

autonomousapps:
I'm enabling CC in all functional tests [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1063). It will at least show if there are issues with enabling it, even if the cache entries can't currently be used.

ljacomet:
Hey Tony,

Indeed, this work was targeted at unblocking `projectHealth`.
It did not cover all the cases to make this plugin fully CC compatible.

That might come in a future PR though.

ljacomet:
And mostly: thanks for the merge! 🎉

Issue #873, opened 2023-03-06
testInstrumentationRunner dependencies are not recognized


Vichy97:
**Plugin version**
1.19.0

**Gradle version**
7.5

**`reason` output for bugs relating to incorrect advice**
`UnusedDepdencies`

**Describe the bug**
This plugin does not recognize dependencies where the classes used are referenced solely in the build.gradle. For example test runner classes used for `testInstrumentationRunner` will not be recognized and will be removed as unused, but this will cause tests to fail.

autonomousapps:
Thanks for the issue. Can you provide a minimal reproducer?

Vichy97:
https://github.com/Vichy97/Dependency-Analysis-Minimal-Repro/tree/issue_873 If you checkout `issue_873` branch and run `./gradlew fixDependencies` you will see it remove the dependency on `module_b` which causes tests to fail

autonomousapps:
Cloning the repo and running `buildHealth`, the report says
```
> Task :buildHealth
Advice for :module_a
Unused dependencies which should be removed:
  androidTestImplementation(project(":module_b"))
```
I'm making this comment just to note that the issue appears before `fixDependencies` -- the advice is to remove the dependency, which `fixDependencies` simply follows. With this reproducer, I ought to be able to fix the issue. Thanks!

autonomousapps:
@Vichy97 should your `:module_b` actually be `androidTestRuntimeOnly`? Since it's not used at compile time, I think that makes sense. Could you double-check in your project?

PR #1040, opened 2023-11-19
Track usage of testInstrumentationRunner in Android projects.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/873.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/model/intermediates/Reason.kt
+++ b/src/main/kotlin/com/autonomousapps/model/intermediates/Reason.kt
@@ -104,6 +104,16 @@ internal sealed class Reason(open val reason: String) {
     override val configurationName: String = "implementation"
   }
 
+  @TypeLabel("testInstrumentationRunner")
+  @JsonClass(generateAdapter = false)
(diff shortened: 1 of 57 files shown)


autonomousapps:
Build failed

src/main/kotlin/com/autonomousapps/model/intermediates/Reason.kt
```diff
@@ -104,6 +104,16 @@ internal sealed class Reason(open val reason: String) {
...
   }
 
+  @TypeLabel("testInstrumentationRunner")
+  @JsonClass(generateAdapter = false)
+  data class TestInstrumentationRunner(override val reason: String) : Reason(reason) {
+    init {
+      buildReason(setOf(reason), "Declares", Kind.AndroidTestInstrumentationRunner)
+    }
+
+    override val configurationName: String = "runtimeOnly"
```
autonomousapps, line 114: Open question if this should be runtimeOnly or implementation, but I think this makes sense.

PR #1042, opened 2023-11-21
Add support for dominator tree calculation for runtime classpath

mmollaverdi:


src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -487,9 +487,13 @@ internal class ProjectPlugin(private val project: Project) {
...
         configurations[dependencyAnalyzer.compileConfigurationName].artifactsFor(dependencyAnalyzer.attributeValueJar)
       )
+      setRuntimeClasspath(
+        configurations[dependencyAnalyzer.runtimeConfigurationName].artifactsFor(dependencyAnalyzer.attributeValueJar)
+      )
       buildPath.set(buildPath(dependencyAnalyzer.compileConfigurationName))
 
-      output.set(outputPaths.artifactsPath)
+      output.set(outputPaths.compileArtifactsPath)
+      outputRuntime.set(outputPaths.runtimeArtifactsPath)
```
mpeyper, line 496: Just as suggestion, if `setCompileClasspath` was renamed `setClasspath` you could use the same task to register `ArtifactsReportTask` twice with different classpaths and output paths so each configuration could be run independently and/or in parallel:

```kt
val artifactsCompileReportTask = tasks.register<ArtifactsReportTask>("artifactsReportCompile$taskNameSuffix") {
      setClasspath(
        configurations[dependencyAnalyzer.compileConfigurationName].artifactsFor(dependencyAnalyzer.attributeValueJar)
      )
      buildPath.set(buildPath(dependencyAnalyzer.compileConfigurationName))
      output.set(outputPaths.compileArtifactsPath)
}

val artifactsRuntimeReportTask = tasks.register<ArtifactsReportTask>("artifactsReportRuntime$taskNameSuffix") {
      setClasspath(
        configurations[dependencyAnalyzer.runtimeConfigurationName].artifactsFor(dependencyAnalyzer.attributeValueJar)
      )
      buildPath.set(buildPath(dependencyAnalyzer.runtimeConfigurationName))
      output.set(outputPaths.runtimeArtifactsPath)
}

// parent task that runs both
val artifactsReportTask = tasks.register("artifactsReport$taskNameSuffix") {
    dependsOn(artifactsCompileReportTask, artifactsRuntimeReportTask)
}
```

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -531,9 +535,13 @@ internal class ProjectPlugin(private val project: Project) {
...
+      runtimePhysicalArtifacts.set(artifactsReportTask.flatMap { it.outputRuntime })
       graphView.set(graphViewTask.flatMap { it.output })
-      outputTxt.set(outputPaths.dominatorConsolePath)
-      outputDot.set(outputPaths.dominatorGraphPath)
+      runtimeGraphView.set(graphViewTask.flatMap { it.outputRuntime })
+      outputTxt.set(outputPaths.compileDominatorConsolePath)
+      runtimeOutputTxt.set(outputPaths.runtimeDominatorConsolePath)
+      outputDot.set(outputPaths.compileDominatorGraphPath)
+      runtimeOutputDot.set(outputPaths.runtimeDominatorGraphPath)
     }
```
mpeyper, line 545: Same suggestion as [above](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1042/files#r1400248269), perhaps this could be simpler to register seperate tasks for `computeCompileDominatorTree$taskNameSuffix` and `computeRuntimeDominatorTree$taskNameSuffix` that use the output from their respective `(compile|runtime)ArtifactsReportTask` tasks, then have a parent task for `computeDominatorTree$taskNameSuffix` that runs both.

mmollaverdi: There was precedence for in the existing code for supporting both compile and runtime classpath in a single tasks (e.g. [ here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/a22aa2fa499fcb814b2295ee0d6a08efffeebca5/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt#L497-L498)), so I opted to do something similar.

src/functionalTest/groovy/com/autonomousapps/jvm/ComputeDominatorTreeSpec.groovy
```diff
@@ -0,0 +1,25 @@
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.ComputeDominatorTreeProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class ComputeDominatorTreeSpec extends AbstractJvmSpec {
```
mmollaverdi, line 8: I know there is an existing [test](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/a22aa2fa499fcb814b2295ee0d6a08efffeebca5/src/functionalTest/groovy/com/autonomousapps/android/DominanceTreeSpec.groovy#L11) for dominator tree, but I couldn't run that one locally, because it depends on some Android config which I didn't have. So I created a new test. Happy to merge them if there are easy instructions for getting the existing test running locally, so that I can update it.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/ComputeDominatorTreeProject.groovy
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  String actualRuntimeDominatorTree() {
+    return gradleProject.rootDir.toPath()
+      .resolve('app/build/reports/dependency-analysis/main/graph/graph-dominator-runtime.txt')
+      .readLines()
+  }
+
+  String expectedRuntimeDominatorTreeTotalSize() {
+    return "61.49 MiB :app"
```
mmollaverdi, line 47: Note the larger total size for the runtime classpath compared to compile classpath.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -531,9 +535,13 @@ internal class ProjectPlugin(private val project: Project) {
...
 
     tasks.register<PrintDominatorTreeTask>("printDominatorTree$taskNameSuffix") {
```
mmollaverdi, line 547: I didn't modify this one. Figured we'd be relying on files written to by the compute task, so including runtime sizes in that one is enough, but happy to add runtime output to this one too.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
mpeyper: That's fair. I'm not very familiar with this codebase either so please don't consider my suggestion as requirements. Perhaps @autonomousapps has an opinion on this?

src/functionalTest/groovy/com/autonomousapps/jvm/ComputeDominatorTreeSpec.groovy
autonomousapps: Totally fine, I think it makes sense to have a JVM spec for this feature too 👍

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
autonomousapps: I could go either way. This particular task is not part of the primary analysis of this plugin. I basically added it because it was convenient to reuse all the machinery around graph analysis I already had set up here, and I thought maybe I'd find a good use for it eventually. At this point, _you_ are the main users of this, as far as I know. So as the users, you should tell me what you'd find most convenient / useful, and that works for me. What you have here, or separate tasks.

src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
autonomousapps, line 1: This is an interesting exception to what I wrote earlier. This task _is_ part of the primary analysis of the plugin, so I don't want to make it "heavier" than it needs to be, to support a secondary task. I think we should split this into two task registrations, one for compile time and one for runtime. That might lend itself to splitting up the other task as well -- you tell me.

autonomousapps (changes requested):
This probably warrants having runtime/compile time separation of tasks, for simplicity. WDYT?

autonomousapps (approved):

PR #1055, opened 2023-12-01
Use the 'shadowed' subprojects from source

jjohannes:
The reason why I would like to have this working is that currently you can't use the plugin via:

```
pluginManagement {
    includedBuild("/path/to/dependency-analysis-android-gradle-plugin")
}
```

Because the moment it becomes an included build, the composite build dependency substitution starts working and then the 'shadowed' components are used from source - which does not work without this change.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -41,6 +41,9 @@ plugins {
 
 // Yes, this is also in pluginManagement above. This is required for normal dependencies.
 includeBuild("testkit")
+// Address subprojects of this build (e.g. 'relocated.asm') by their coordinates.
+// https://docs.gradle.org/current/userguide/composite_builds.html#included_build_declaring_substitutions
+includeBuild(".")
 
 dependencyResolutionManagement {
   repositories {
--- a/shadowed/README.asciidoc
+++ b/shadowed/README.asciidoc
@@ -1,4 +1,0 @@
-The projects in this directory are not used as project dependencies,
-primarily due to limitations with IDEA consuming the shadowed variants of
-projects. Also they're very stable. Therefore, we publish the shadowed jars
-and consume those as binary dependencies for simplicity.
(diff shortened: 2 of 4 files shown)


autonomousapps:
Build failed

jjohannes:
Configuring the Shadow plugin is such a pain. Right now you need to be a super expert in Gradle to get it right...

autonomousapps:
I use this plugin as an included build all the time. What error are you seeing specifically? Does it have to do with trying to update those shadowed components?

jjohannes:
Alright. I was assuming this is a general problem. Feel free to discard this PR if this is not an issue in other "include" scenarios.

I get the following error, without these fixes: https://scans.gradle.com/s/fkqwok5fyqc3o/console-log?anchor=384&page=1
It's trying to compile the plugin, but does not find the shaded classes.

You can reproduce it in a setup like this one: https://github.com/jjohannes/gradle-project-setup-howto/tree/__rework (branch **__rework** and change the two places where it says `includeBuild("/Users/jendrik/projects/gradle/plugins/dependency-analysis-android-gradle-plugin")` to your local path)

autonomousapps:
Huh interesting. I'll take a look. One of the reasons for my current setup is that those shadowed components don't change much, and I think one at least is quite large, so I was trying to reduce the size of the DAGP jar. Would your change impact normal consumers in any way? Would they still use the pre published binaries?

jjohannes:
This should not change anything about how the plugin is published and used by "normal" consumers.

The only difference now is that when you build locally, it **always** takes the locally built version of the shadowed dependencies to compile and test against. Right now, it sometime takes the published ones (if you just build the plugin or include it like you do) and sometimes takes the local ones (if you include it like I did).

In my mind, the changes to the `shadowed/*/build.gradle.kts` files are valid fixes in the setup. Without these, using the shadowed libs from source is broken. You could also just take this part of the changes in – and **not** the `includeBuild(".")`. Then my way of including works, and the normal plugin building keeps working as before.

In any case, this is not meant to change anything about how the plugin is published or used. It should just make it easier for others to use the plugin from source if they want to test or contribute something. I can make these adjustments locally, if you feel uncomfortable with integrating this.

autonomousapps (approved):

PR #1057, opened 2023-12-02
Properly linkify report file in console

SimonMarquis:
On Windows the current implementation would result in non-linkified output:

```
There were dependency violations. See report at E:\AndroidStudioProjects\nowinandroid\build\reports\dependency-analysis\build-health-report.txt 
```

But using the `java.net.URI` formatting will properly add the necessary triple slashes and replace backslack with regular forward slashes, making it supported on Windows as well:

```
There were dependency violations. See report at file:///E:/AndroidStudioProjects/nowinandroid/build/reports/dependency-analysis/build-health-report.txt 
```

(partially solved by https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/669 initially)

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt
@@ -102,7 +102,7 @@ abstract class FindInlineMembersTask @Inject constructor(
       if (!finder.didWriteErrors) {
         errorsReport.writeText("")
       } else {
-        logger.warn("There were errors during inline member analysis. See ${errorsReport.toUri()}")
+        logger.warn("There were errors during inline member analysis. See ${errorsReport.toPath().toUri()}")
       }
     }
   }


autonomousapps (commented):
Thanks! Should we do this for ProjectHealth as well?

SimonMarquis:
Unfortunately I don't see the `project-health-report.txt` file path beeing logged.

SimonMarquis:
Maybe this one could be refactored as well?

https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/9967d87d0d2ae4499985f6b30507cacad88e83d5/src/main/kotlin/com/autonomousapps/tasks/FindInlineMembersTask.kt#L105

autonomousapps:
> Unfortunately I don't see the `project-health-report.txt` file path beeing logged.

oh right, that one always just prints to console.

> Maybe this one could be refactored as well?

sure! thanks

autonomousapps (approved):

PR #1066, opened 2023-12-10
GenerateBuildHealthTask is compatible with the configuration cache.

autonomousapps:
- [x] There is commented out code to delete.


autonomousapps:
Build failed

PR #1070, opened 2023-12-11
Fix plugin, repository, and settings script rendering for Kotlin Dsl

tcrawford-figure:
I did take note that the `Dependency` rendering for Kotlin Dsl also requires some work and will likely not generate a valid Kotlin Dsl, but chose to forego adjusting any of that since it seems to require a larger change.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/RewriteDependenciesProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/RewriteDependenciesProject.groovy
@@ -134,7 +134,7 @@ final class RewriteDependenciesProject extends AbstractProject {
 
   final String expectedBuildFile = '''\
     plugins {
-      id 'java-library'
+      id('java-library')
     }
     
     dependencies {
@@ -147,7 +147,7 @@ final class RewriteDependenciesProject extends AbstractProject {
   /** This build script has only been upgrade. Downgrades have been ignored. */
   final String expectedBuildFileUpgraded = '''\
     plugins {
-      id 'java-library'
+      id('java-library')
     }
     
     dependencies {
--- a/testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitPlugin.kt
+++ b/testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitPlugin.kt
@@ -53,8 +53,14 @@ public class GradleTestKitPlugin : Plugin<Project> {
           // Gradle tests generally require more metaspace
           jvmArgs("-XX:+HeapDumpOnOutOfMemoryError", "-XX:MaxMetaspaceSize=1g")
 
-          systemProperty("com.autonomousapps.plugin-under-test.repo", extension.funcTestRepo.absolutePath)
-          systemProperty("com.autonomousapps.plugin-under-test.version", version.toString())
+          systemProperty(
+            "com.autonomousapps.plugin-under-test.repo",
+            target.provider { extension.funcTestRepo.absolutePath }.get()
+          )
+          systemProperty(
+            "com.autonomousapps.plugin-under-test.version",
+            target.provider { version.toString() }.get()
+          )
         }
       }
       extension.setTestTask(gradleTest)
--- a/testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitPlugin.kt
+++ b/testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitPlugin.kt
@@ -53,14 +53,8 @@ public class GradleTestKitPlugin : Plugin<Project> {
           // Gradle tests generally require more metaspace
           jvmArgs("-XX:+HeapDumpOnOutOfMemoryError", "-XX:MaxMetaspaceSize=1g")
 
-          systemProperty(
-            "com.autonomousapps.plugin-under-test.repo",
-            target.provider { extension.funcTestRepo.absolutePath }.get()
-          )
-          systemProperty(
-            "com.autonomousapps.plugin-under-test.version",
-            target.provider { version.toString() }.get()
-          )
+          systemProperty("com.autonomousapps.plugin-under-test.repo", extension.funcTestRepo.absolutePath)
+          systemProperty("com.autonomousapps.plugin-under-test.version", version.toString())
         }
       }
       extension.setTestTask(gradleTest)
--- a/testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugin.kt
+++ b/testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugin.kt
@@ -10,8 +10,9 @@ public class Plugin @JvmOverloads constructor(
 ) : Element.Line {
 
   override fun render(scribe: Scribe): String = scribe.line { s ->
-    s.append("id ")
+    s.append("id(")
     s.appendQuoted(id)
+    s.append(")")
     version?.let { v ->
       s.append(" version ")
       s.appendQuoted(v)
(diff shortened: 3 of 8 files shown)


testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitPlugin.kt
```diff
@@ -53,8 +53,14 @@ public class GradleTestKitPlugin : Plugin<Project> {
...
-          systemProperty("com.autonomousapps.plugin-under-test.repo", extension.funcTestRepo.absolutePath)
-          systemProperty("com.autonomousapps.plugin-under-test.version", version.toString())
+          systemProperty(
+            "com.autonomousapps.plugin-under-test.repo",
+            target.provider { extension.funcTestRepo.absolutePath }.get()
+          )
+          systemProperty(
+            "com.autonomousapps.plugin-under-test.version",
+            target.provider { version.toString() }.get()
+          )
```
tcrawford-figure, line 63: I found that setting the system properties was too eager in a project I'm working on. The version hasn't been set yet when the plugin is applied and the extension hasn't been configured by the time this task is registered.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/RewriteDependenciesProject.groovy
```diff
@@ -134,7 +134,7 @@ final class RewriteDependenciesProject extends AbstractProject {
 
   final String expectedBuildFile = '''\
     plugins {
-      id 'java-library'
+      id('java-library')
```
autonomousapps, line 137: You probably want to use `"` instead of `'` here, for Kotlin DSL compatibility.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/RewriteDependenciesProject.groovy
```diff
@@ -147,7 +147,7 @@ final class RewriteDependenciesProject extends AbstractProject {
   /** This build script has only been upgrade. Downgrades have been ignored. */
   final String expectedBuildFileUpgraded = '''\
     plugins {
-      id 'java-library'
+      id('java-library')
```
autonomousapps, line 150: here too

testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugin.kt
```diff
@@ -10,8 +10,9 @@ public class Plugin @JvmOverloads constructor(
 ) : Element.Line {
 
   override fun render(scribe: Scribe): String = scribe.line { s ->
-    s.append("id ")
+    s.append("id(")
     s.appendQuoted(id)
+    s.append(")")
```
autonomousapps, line 15: I'm not going to be able to accept this PR as-is. The whole point of `Scribe` is that you pass in either GROOVY or KOTLIN in its constructor and then it uses that to modify its behavior. Support for Kotlin DSL is simply not ready yet. If you want to attempt to do that in my intended way, I'm open to that.

testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/render/Scribe.kt
```diff
@@ -85,4 +85,8 @@ public class Scribe @JvmOverloads constructor(
   }
 
   private fun quote(): String = if (dslKind == GradleProject.DslKind.GROOVY) "'" else "\""
+
+  internal fun build(): String {
+    return buffer.toString()
+  }
```
autonomousapps, line 91: what is this for?

autonomousapps (changes requested):
Thanks for the PR, but I can't accept it as-is.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/RewriteDependenciesProject.groovy
tcrawford-figure: You are correct. The Kotlin DSL that gets generated here would be quoted with `"` instead of `'`. However, this test is specifically for the groovy DSL. If I were to write a test for the Kotlin DSL I would use `"`. I can modify this so that the same syntax works regardless of DSL kind if you would like? That is, I would forego using Scribe to quote what needs quoted and we would just end up with `id("my-plugin")` regardless of DSL.

testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/render/Scribe.kt
tcrawford-figure: I used this in the SettingScript in order to build an unindented line and would enable using the quoting type that is currently tied to the DslKind. 

That is:
- Groovy = `'`
- Kotlin = `"`

I'll admit that this is probably a less desired solution to knowing what kind of quote to use based on DslKind 🤦🏼‍♂️

However, if for portability between languages, `"` would be agreed on as the quoting type to use universally, then this is completely unnecessary.

testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugin.kt
tcrawford-figure: I completely understand! My PR intent was to get most (not all) of the current rendering functionality of Kotlin into a sort of minimal viable product state with minimal changes to the current infrastructure to just get Kotlin projects building at least. 

Based on your feedback, I guess my question is whether you want Groovy Gradle to look like how you currently have it being rendered or if you're ok with Groovy Gradle rendering in a form that also works with Kotlin Gradle (in many cases)?

I can attempt to make updates that would make the generated code between DslKinds more similar, but also understand if you'd rather take that on since it's probably a larger change.

autonomousapps: I don't want consumers of `Scribe` to have to care about this. Scribe itself should handle it. That likely would involve enhancements to Scribe -- thinks like `methodCall` which would emit parens for Kotlin and no parens for Groovy. And yeah -- I want the Groovy to look idiomatic.

Why do you _need_ Kotlin projects building? This project generates test fixtures, not "real" production projects. Note that I also want support for Kotlin DSL, but just haven't had time to work on it. I'm curious about your use-case. I appreciate the PR for discussion purposes.

tcrawford-figure: Ah! Something like `methodCall` definitely makes sense. I don't have a strong use case for using the Kotlin DSL format here. It's primarily what I've been comfortable working with. So far, I have just been using this to generate projects for test purposes, much like you have in this repository. I just chose to use the Kotlin DSL since I saw it was available. 

I kind of figured that you might want Groovy and Kotlin to take on idiomatic forms. 

Given your feedback I'll see what I can think through for updating `Scribe` when I have some time. I think the important piece that you touched on was to add methods for creating abstract language constructs like `methodCall`.

autonomousapps: I might take this up myself soon. I very much appreciate the interest and the strong signal that being able to write out Kotlin DSL is desirable 👍

autonomousapps:
@tcrawford-figure would you be willing to push a separate PR with the documentation changes? I think those would be useful for lots of people <3

tcrawford-figure:
> @tcrawford-figure would you be willing to push a separate PR with the documentation changes? I think those would be useful for lots of people <3

Yes! Absolutely! Hopefully I can get a new PR up today.

testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugin.kt
tcrawford-figure: Sweet! Looking forward to it! For now, I'll fall back to letting everything get generated using Groovy Gradle in my project instead.

tcrawford-figure:
Closing this PR based on feedback discussed above. Will open a new PR with documentation changes.

PR #1077, opened 2023-12-15
Improve truth tasks

### LINKED ISSUE #1068: [gradle-testkit-truth] More useful error messages
I'm adding `gradle-testkit-truth` to my project and it works fine when the tests passes but when the things go wrong the message is not that useful.

With this code `assertThat(buildResult).task(":lightsaberChecka").failed()` I get this message error:

```
value of                : buildResult.task(:lightsaberChecka)
expected to have a value: FAILED
but was                 : null
buildResult was         : org.gradle.testkit.runner.internal.FeatureCheckBuildResult@3670f00
```

`org.gradle.testkit.runner.internal.FeatureCheckBuildResult@3670f00` gives no information about what's going on.

I would print the `output` as the "actual" value to provide this information.

Also `task(":lightsaberChecka")` provides a Subject of a `null` instance that is not that useful. I think that something like this would improve the error message:

https://github.com/SchwarzIT/dagger-lightsaber/blob/98686ec190168c7f14fed377651c4024af8b9788/gradle-plugin/src/functionalTest/kotlin/schwarz/it/lightsaber/gradle/truth/BuildResultSubject.kt#L17-L23

With this I get error messages like this:

```
value of:
    buildResult.getTasks()
expected to contain:
    :lightsaberChecka
but was:
    [:compileJava, :processResources, :classes, :compileTestJava, :lightsaberCheck]
buildResult was:
    > Task :compileJava
    > Task :processResources NO-SOURCE
    > Task :classes
    > Task :compileTestJava NO-SOURCE
...
```

As you can see I have something similar implemented already but I like the idea of having a truth library for this so I can contribute to this library if you think that things like this are valuable.

BraisGabin:
Close #1068

This PR changes a bit the idea behind `BuildResultSubject.task`. It doesn't return a `BuildTaskSubject` with an `actual == null` anymore. Now BuildResultSubject.task fails if the task doesn't exist with a better error message.

Before:
```
Caused by: value of:
    buildResult.task(:proj:compileJava2)
expected to have a value:
    FAILED
but was:
    null
buildResult was:
    Calculating task graph as no cached configuration is available for tasks: proj:assemble
    > Task :proj:processResources NO-SOURCE
    > Task :proj:compileJava
    > Task :proj:classes
    > Task :proj:jar
    > Task :proj:assemble
    
    BUILD SUCCESSFUL in 2s
    2 actionable tasks: 2 executed
    Configuration cache entry stored.
```

After:
```
Caused by: value of:
    buildResult.getTasks()
expected to contain:
    :proj:compileJava2
but was:
    [:proj:compileJava, :proj:processResources, :proj:classes, :proj:jar, :proj:assemble]
buildResult was:
    Calculating task graph as no cached configuration is available for tasks: proj:assemble
    > Task :proj:processResources NO-SOURCE
    > Task :proj:compileJava
    > Task :proj:classes
    > Task :proj:jar
    > Task :proj:assemble
    
    BUILD SUCCESSFUL in 2s
    2 actionable tasks: 2 executed
    Configuration cache entry stored.
```

With only that change it would be impossible to check that a task does not exist. For that reason I added `hasNotTask`. I'm not very happy with the name, any suggestion is more than welcome.

And this function I think that also improves the error message:

Before:
```
Caused by: value of:
    buildResult.task(:proj:compileJava)
expected:
    null
but was:
    :proj:compileJava=SUCCESS
buildResult was:
    Calculating task graph as no cached configuration is available for tasks: proj:assemble
    > Task :proj:processResources NO-SOURCE
    > Task :proj:compileJava
    > Task :proj:classes
    > Task :proj:jar
    > Task :proj:assemble
    
    BUILD SUCCESSFUL in 2s
    2 actionable tasks: 2 executed
    Configuration cache entry stored.
```

After:
```
Caused by: value of:
    buildResult.getTasks()
expected not to contain:
    :proj:compileJava
but was:
    [:proj:processResources, :proj:compileJava, :proj:classes, :proj:jar, :proj:assemble]
buildResult was:
    Calculating task graph as no cached configuration is available for tasks: proj:assemble
    > Task :proj:processResources NO-SOURCE
    > Task :proj:compileJava
    > Task :proj:classes
    > Task :proj:jar
    > Task :proj:assemble
    
    BUILD SUCCESSFUL in 2s
    2 actionable tasks: 2 executed
    Configuration cache entry stored.
```

### DIFF
--- a/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
+++ b/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
@@ -9,6 +9,7 @@ import com.google.common.truth.StringSubject
 import com.google.common.truth.Subject
 import com.google.common.truth.Subject.Factory
 import com.google.common.truth.Truth.assertAbout
+import com.google.errorprone.annotations.CanIgnoreReturnValue
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.TaskOutcome
 
@@ -37,11 +38,16 @@ public class BuildResultSubject private constructor(
     return check("getOutput()").that(actual!!.output)
   }
 
+  @CanIgnoreReturnValue
   public fun task(path: String): BuildTaskSubject {
     if (actual == null) {
       failWithActual(simpleFact("build result was null"))
     }
-    return check("task(%s)", path).about(buildTasks()).that(actual!!.task(path))
+    val tasks = actual!!.tasks.map { it.path }
+
+    check("getTasks()").that(tasks).contains(path)
+
+    return check("task(%s)", path).about(buildTasks()).that(actual.task(path))
   }
 
   public fun doesNotHaveTask(path: String) {
--- a/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
+++ b/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
@@ -44,6 +44,15 @@ public class BuildResultSubject private constructor(
     return check("task(%s)", path).about(buildTasks()).that(actual!!.task(path))
   }
 
+  public fun doesNotHaveTask(path: String) {
+    if (actual == null) {
+      failWithActual(simpleFact("build result was null"))
+    }
+    val tasks = actual!!.tasks.map { it.path }
+
+    check("getTasks()").that(tasks).doesNotContain(path)
+  }
+
   public fun getTasks(): BuildTaskListSubject {
     if (actual == null) {
       failWithActual(simpleFact("build result was null"))
--- a/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
+++ b/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
@@ -37,11 +37,11 @@ public class BuildResultSubject private constructor(
     return check("getOutput()").that(actual!!.output)
   }
 
-  public fun task(name: String): BuildTaskSubject {
+  public fun task(path: String): BuildTaskSubject {
     if (actual == null) {
       failWithActual(simpleFact("build result was null"))
     }
-    return check("task(%s)", name).about(buildTasks()).that(actual!!.task(name))
+    return check("task(%s)", path).about(buildTasks()).that(actual!!.task(path))
   }
 
   public fun getTasks(): BuildTaskListSubject {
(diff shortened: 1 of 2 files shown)


testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
```diff
@@ -37,11 +38,25 @@ public class BuildResultSubject private constructor(
...
     }
-    return check("task(%s)", name).about(buildTasks()).that(actual!!.task(name))
+    val tasks = actual!!.tasks.map { it.path }
+
+    check("getTasks()").that(tasks).contains(path)
+
+    return check("task(%s)", path).about(buildTasks()).that(actual.task(path))
+  }
+
+  public fun hasNotTask(path: String) {
```
BraisGabin, line 53: I'm not too happy with this function name. Any idea is more than welcome

jenni-arias: `checkHasTask`?

jenni-arias (approved):

testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
autonomousapps: Is there anything in the core Truth library that is similar that we can use for inspiration? I'm not too happy with it either, but I could live with it....

autonomousapps (approved):
Thanks! This looks good, other than the wonky name :) Looks like you might need to rebase.

testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/BuildResultSubject.kt
BraisGabin: `doesNotContainTask`? Similar to `doesNotContain`. The method that I use inside this function.

autonomousapps: how about `doesNotHaveTask`, since it uses the same verb as `hasTask`? you could also add `doesNotContainTask` since people are probably used to the verb "contain", and it could do the same thing (one function simply calling the other).

BraisGabin: Renamed to `doesNotHaveTask`. I think that, for now, one method is enough.

The thing I'm not sure about is this:

> since it uses the same verb as `hasTask`

There is not a `hasTask` function right now. Should I rename `task` to `hasTask`?

- `task` is more aligned with the gradle api.
- `hasTask` is more aligned with the truth api and with the fact that we have `doesNotHaveTask`.

autonomousapps: Oh huh, I'm not sure why I thought there was. Anyway, this is good enough. Thanks!

autonomousapps (approved):

PR #1081, opened 2023-12-19
Use `assertThat` instead of `assertAbout`

BraisGabin:
There are a lot more places where `assertAbout` is used but groovy doesn't allow to import multiple functions with the same name. I didn't change those because probable `assertThat` is better than alias imports. Other option it to don't import `Truth.assertThat` and in the places where `Truth` is used call them like `Truth.assertThat(...)`.

Anyway, these changes are the easier ones and it should make the code easier to read and provide better error messages.


autonomousapps (commented):
Honestly I prefer to keep using `assertAbout` here. Thanks though!

PR #1086, opened 2023-12-27
chore: bump DAGP to 1.28.0.

autonomousapps:


build.gradle.kts
```diff
@@ -317,16 +317,35 @@ dependencyAnalysis {
...
+          "com.autonomousapps:asm-relocated",
+        )
+      }
+      onUsedTransitiveDependencies {
+        exclude(
+          // TODO(tsr):
+          //  1. version catalog ref (libs.relocated.asm) is not working for some reason
+          //  2. This fatjar is bundling Kotlin stdlib classes, which is what the analysis is detecting
+          //  3. The exclusion only works when I use the project reference, not the Maven coordinates
+          ":asm-relocated",
```
autonomousapps, line 348: @jjohannes see point 3. This bug only appeared after I accepted your changes regarding how DAGP references internal projects (and I'm not blaming that change). Any idea why the exclude would work with `:asm-relocated` but not the full Maven coordinates?

jjohannes: I am not 100% sure without diving deep into the code again... but I think it is like this:

We have the bevior that a component can have two represenations if it is a local project. In this case:
1. `project(":asm-relocated")`
2. `com.autonomousapps:asm-relocated`

Both are tracked throughout the analysis. But at the end, a decisions is made which is the "primary" representation.
We decide that based on what was used when the dependency was declared if possible, that's why it should be right to use (2) here. We declare it like that in the Version Catalog.

Maybe here, we do not have the declaration tracked? (either on purpose, or it's a bug). In that case, I think we decide as follows: If we are inside the multi-project to which the corresponding project belongs, we decide for (1), otherwise for (2). In a single build, this is the case for every project. In a composite build, this is the case for any project in the root build.

Now if the "exclusion" runs late on the analysis results (which I think is the case) it only looks at this "primary" representation because that decision has been made.

Issue #980, opened 2023-10-02
Redundant advises starting from 1.21.0


Jacks0N23:
**Plugin version**
Starting from 1.21.0 (including lates 1.24.0)

**Gradle version**
8.3

**JDK version**
17

**(Optional) Android Gradle Plugin (AGP) version**
8.1.2

**`reason` output for bugs relating to incorrect advice**
Shortly
```
Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseRuntimeClasspath:
:feature-counterparty
\--- androidx.recyclerview:recyclerview:1.3.1
      \--- androidx.customview:customview:1.1.0

```
<details>
    <summary> Full Logs</summary>
    
    ```
    > Task :feature-counterparty:reason
    
    ----------------------------------------
    You asked about the dependency 'androidx.customview:customview:1.1.0'.
    You have been advised to add this dependency to 'implementation'.
    ----------------------------------------
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseCompileClasspath:
    :feature-counterparty
    \--- androidx.recyclerview:recyclerview:1.3.1
    \--- androidx.customview:customview:1.1.0
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseRuntimeClasspath:
    :feature-counterparty
    \--- androidx.recyclerview:recyclerview:1.3.1
    \--- androidx.customview:customview:1.1.0
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseUnitTestCompileClasspath:
    :feature-counterparty
    \--- androidx.recyclerview:recyclerview:1.3.1
    \--- androidx.customview:customview:1.1.0
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseUnitTestRuntimeClasspath:
    :feature-counterparty
    \--- androidx.recyclerview:recyclerview:1.3.1
    \--- androidx.customview:customview:1.1.0
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseAndroidTestCompileClasspath:
    :feature-counterparty
    \--- androidx.recyclerview:recyclerview:1.3.1
    \--- androidx.customview:customview:1.1.0
    
    Shortest path from :feature-counterparty to androidx.customview:customview:1.1.0 for releaseAndroidTestRuntimeClasspath:
    :feature-counterparty
[...]

PR #1100, opened 2024-01-11
fix: don't suggest unnecessary dependency relating to Android res and new IDs.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/980.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/parse/explodeAndroidResSource.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/parse/explodeAndroidResSource.kt
@@ -12,7 +12,7 @@ import java.io.File
 
 internal class AndroidLayoutParser(
   private val projectDir: File,
-  private val layouts: Set<File>
+  private val layouts: Set<File>,
 ) {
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/autonomousapps/internal/parse/explodeAndroidResSource.kt
```diff
@@ -35,16 +35,21 @@ internal class AndroidLayoutParser(
...
+  resources: Iterable<File>,
 ) {
 
   val androidResSource: Set<ExplodedRes> = resources
     .map { it to buildDocument(it) }
     .mapToSet { (file, doc) ->
+      // TODO(tsr): this could be more performant by not parsing every XML document twice
+      val attrRefs = extractAttrsFromResourceXml(doc) + extractContentReferencesFromResourceXml(doc)
+      val newIds = extractAttrsFromResourceXml(doc, true) + extractContentReferencesFromResourceXml(doc, true)
+      val nonLocalAttrRefs = attrRefs - newIds
```
autonomousapps, line 47: TODO

PR #1107, opened 2024-01-22
Make sure all input files are using path sensitivity of relative to avoid cache misses in CI builds.

NikolayMetchev:


autonomousapps (changes requested):
See associated issue comments for why this PR cannot be accepted.

PR #1117, opened 2024-01-30
fix: Reason was failing to give correct results for project dependencies.

autonomousapps:


src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -238,9 +237,12 @@ abstract class ReasonTask @Inject constructor(
         ?: emptySet()
     }
 
+    /** Returns null if there is no advice for the given id. */
     private fun findAdviceIn(projectAdvice: ProjectAdvice): Advice? {
-      // Would be null if there is no advice for the given id.
-      return projectAdvice.dependencyAdvice.find { it.coordinates.gav() == coord.gav() }
+      return projectAdvice.dependencyAdvice.find { advice ->
+        val adviceGav = advice.coordinates.gav()
+        adviceGav == coord.gav() || adviceGav == requestedCoord.gav()
```
autonomousapps, line 244: @jjohannes another area where the Coordinates implementation is problematic.

autonomousapps: Ugh I should write a test for this, shouldn't I.

jjohannes: Yeah this could be simplified throughout the code base I believe. 
I think the multiple "complicated" places are a result from changing assumptions: In the original implementation it's assumed that there is always only one coordinates/identifier and that the declaration (requested) looks the same as the result.

Taking a fresh look at the "Coordinates" with taking the new assumptions/situation into account, I believe this can be reworked in a way that these kinds of bugs become less likely.

Issue #1124, opened 2024-02-06
Reason id dependency filter ambiguity


seregamorph:
**Plugin version**
1.29.0

**Gradle version**
8.6

**JDK version**
17

**Describe the bug**
The `reason` task receives `--id` argument and filters dependency in a pretty lenient way. E.g. if the `module` has these dependencies:
* `project("list")`
* `project("list-A")`
* `project("list-B")`

and we call
```
./gradlew :module:reason --id list
```
if can pick any of those three modules (e.g. `list-A`). This depends on the combination of dependency scopes and order of definition.

**Expected behavior**
1. If the name of the dependency is specified fully, it should be preferred.
2. If there is am ambiguity and several dependencies match, it should throw exception instead of picking random first matching one

**Additional context**
The problem is around these two places in the code:
* https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt#L219
```
// picks first if several matches
val gavKey = dependencyUsages.entries.find(requestedId::matchesKey)?.key ?:
```
* https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/internal/utils/coordinatesStrings.kt#L32
```
// too lenient
  // finally the most lenient check
  if (startsWithKey(mapEntry)) return true
```

**To Reproduce**
Let me know if you need sample project to reproduce.

seregamorph:
The solution proposed here: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1125

autonomousapps:
Thanks for the issue! This problem has annoyed me for a while, but I haven't had the bandwidth to focus on it. I appreciate the PR 👍

seregamorph:
Addressed in https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1125

PR #1125, opened 2024-02-06
#1124 Reason explanation ambiguity

seregamorph:
The solution proposal for https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1124


seregamorph:
For unit testing it looks trivial, but perhaps you mean another type of test. This is my first submission here, so I'm not quite familiar with test approach in this project - please give a good reference what you mean :)

seregamorph:
Well, I see failed test and probably functionalTest `jvm.BundleSpec` can be a good candidate for enhancement?

autonomousapps:
I think you could update `JvmReasonSpec` maybe. Most of the time, new tests can be written by copy/pasting other related tests and tweaking slightly.

seregamorph:
@autonomousapps can you please recheck the PR?

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -216,7 +216,8 @@ abstract class ReasonTask @Inject constructor(
         }?.gav()
 
       // Guaranteed to find full GAV or throw
-      val gavKey = dependencyUsages.entries.find(requestedId::matchesKey)?.key
+      val filteredDependencyKey = findFilteredDependencyKey(requestedId)
+      val gavKey = filteredDependencyKey
         ?: annotationProcessorUsages.entries.find(requestedId::matchesKey)?.key
```
seregamorph, line 221: @autonomousapps it seems that this lenient filter can have the same problem. WDYT we can use the same logic here? Or let's keep it out of scope?

autonomousapps: No, let's fix both.

src/functionalTest/groovy/com/autonomousapps/jvm/BundleSpec.groovy
```diff
@@ -71,7 +72,22 @@ final class BundleSpec extends AbstractJvmSpec {
 
     // TODO reason needs to be updated to show that the -jvm variant is used
     expect:
-    build(gradleVersion, gradleProject.rootDir, ':consumer:reason', '--id', 'com.squareup.okio:okio')
+    build(gradleVersion, gradleProject.rootDir, ':consumer:reason', '--id', 'com.squareup.okio:okio-jvm')
```
autonomousapps, line 75: Why did you make this change?

src/functionalTest/groovy/com/autonomousapps/jvm/BundleSpec.groovy
```diff
@@ -71,7 +72,22 @@ final class BundleSpec extends AbstractJvmSpec {
...
+  def "reason fails when there is dependency filtering ambiguity"() {
+    given:
+    def project = new BundleKmpProject2()
+    gradleProject = project.gradleProject
+
+    when:
+    def result = buildAndFail(gradleVersion, gradleProject.rootDir, ':consumer:reason', '--id', 'com.squareup.okio:okio')
+
+    then:
+    assertThat(result.output).contains("> Coordinates 'com.squareup.okio:okio' matches more than 1 dependency [com.squareup.okio:okio-jvm:3.0.0, com.squareup.okio:okio:3.0.0]")
```
autonomousapps, line 91: We don't want this. We want to be able to reference a dependency by its identifier alone. If the user passes in `com.squareup.okio:okio`, it should unambiguously match `com.squareup.okio:okio:3.0.0`.

autonomousapps (changes requested):

src/functionalTest/groovy/com/autonomousapps/jvm/BundleSpec.groovy
seregamorph: You are totally right. I've implemented it this way for project modules, but forgot about non-project 🤦‍♂️.
BTW thanks for integration testing - it highlighted the problem (but I still missed it 🤦‍♂️).
Will address all comments in a couple of days - it seems I've caught flu and need to restore.

src/main/kotlin/com/autonomousapps/internal/utils/coordinatesStrings.kt
```diff
@@ -34,8 +34,16 @@ internal fun <T> String.matchesKey(mapEntry: Map.Entry<String, T>): Boolean {
   return false
 }
 
-internal fun <T> String.equalsKey(mapEntry: Map.Entry<String, T>) =
-  mapEntry.key.firstCoordinatesKeySegment() == this || mapEntry.key.secondCoordinatesKeySegment() == this
+internal fun <T> String.equalsKey(mapEntry: Map.Entry<String, T>): Boolean {
```
seregamorph, line 37: please note, that this method is also used is `ReasonTask.getUsagesFor`. Can it affect the logic here?

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
seregamorph: done

src/functionalTest/groovy/com/autonomousapps/jvm/BundleSpec.groovy
seregamorph: reverted

src/functionalTest/groovy/com/autonomousapps/jvm/BundleSpec.groovy
seregamorph: done

seregamorph:
I've discovered one more related problem. The project submodules are represented with both first and second coordinates key segment. Consider sample module with dependencies that have keys in `dependencyUsages.entries`:
* `"demo-gradle-multi-module:list|:list"`
* `"demo-gradle-multi-module:list:sublist|:list:sublist"`

and the filter `--id=demo-`.

Old behavior: it will take first non-ordered, there is ambiguity. In my experiment it took
```
You asked about the dependency 'demo-gradle-multi-module:list'.
There is no advice regarding this dependency.
```

The new behavior:
```
* What went wrong:
Execution failed for task ':app:reason'.
> A failure occurred while executing com.autonomousapps.tasks.ReasonTask$ExplainDependencyAdviceAction
   > Coordinates 'demo-' matches more than 1 dependency [:list, :list:sublist]
```
Instead of ambiguity it throws failure. But the error message can be confusing: we were filtering by `demo-*`, got `:list*`.
But from my perspective it's not a new problem, it does not make it worse and I'd say it's succession behavior :)

WDYT?

seregamorph:
Do you still have concerns or just need more time for better validation of the change?

autonomousapps:
> Do you still have concerns or just need more time for better validation of the change?

Just more time. I've been underwater at work. Thanks for your patience.

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -253,6 +253,29 @@ abstract class ReasonTask @Inject constructor(
       }
 
     private fun wasFiltered(): Boolean = finalAdvice == null && unfilteredAdvice != null
+
+    internal companion object {
+      internal fun findFilteredDependencyKey(dependencies: Set<Map.Entry<String, Any>>, requestedId: String): String? {
```
autonomousapps, line 258: does this need to be internal instead of private?

src/test/kotlin/com/autonomousapps/tasks/ReasonTaskTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.autonomousapps.model.declaration.Bucket
+import com.autonomousapps.model.declaration.Variant
+import com.autonomousapps.model.intermediates.Usage
+import com.autonomousapps.tasks.ReasonTask.ExplainDependencyAdviceAction.Companion.findFilteredDependencyKey
+import org.gradle.api.InvalidUserDataException
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+
+class ReasonTaskTest {
```
autonomousapps, line 12: Thank you for this great unit test

autonomousapps (approved):

src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
seregamorph: because of accessing from the unit test

Issue #1111, opened 2024-01-24
Could not determine the dependencies of task ':app:synthesizeProjectViewDebugTest'.


giovanischiar:
**Build scan link**
https://scans.gradle.com/s/4qwwjyf5chvkk

**Plugin version**
latest snapshot "+"

**Gradle version**
8.2

**JDK version**
17

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
1.9.21 

**(Optional) Android Gradle Plugin (AGP) version**
8.2.1

**Describe the bug**
When I add the `id("com.autonomousapps.dependency-analysis") version "+"` to root build.gradle and run `./gradlew projectHealth` I got this error Could not determine the dependencies of task ':app:synthesizeProjectViewDebugTest'.

**To Reproduce**
1. Add `id("com.autonomousapps.dependency-analysis") version "+"` to root build.gradle
2. Go to terminal and run `./gradlew projectHealth `

**Expected behavior**
The default output for projectHealth

autonomousapps:
Thanks for the issue!

Hopefully the build scan will be enough, but do you have a minimal reproducer? Those repro steps don't look sufficient to me.

gabrielittner:
I've got a reproducer, this is a new project created with Android Studio Hedgehog with the only change being that I added this plugin [MyApplication.zip](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/files/14068667/MyApplication.zip)

ZacSweers:
Here's another repro, just uncomment the commented out enabling of DAGP in the root build.gradle.kts: https://github.com/ZacSweers/CatchUp

autonomousapps:
> I've got a reproducer, this is a new project created with Android Studio Hedgehog with the only change being that I added this plugin [MyApplication.zip](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/files/14068667/MyApplication.zip)

@gabrielittner I downloaded this zip and ran `buildHealth` on it and there were no errors. Is there something I'm missing?

gabrielittner:
I can still reproduce it when I download the zip and run buildHealth. However after deleting `~/.android/analytics.settings` it doesn't reproduce anymore and as soon as the file is recreated. Just the existence of the file seems to be enough, so you should be able to reproduce it by just creating it without content.

autonomousapps:
how is that file used? I'm not familiar with it. do you have an example with actual content? links to documentation?

[...]

PR #1132, opened 2024-02-14
fix: don't pass in android res (incl layouts) to XmlSourceExploderTask.

autonomousapps:
Looks like an AGP bug. A ticket will be filed later.

Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1111.


autonomousapps:
https://issuetracker.google.com/issues/325307775

PR #1137, opened 2024-02-22
Do not suggest to move dependencies between feature variants

jjohannes:
...as this may lead to confusing reports. I had the case where I got an advice like this:

```
    testFixturesImplementation project(':producer')

    extraImplementation project(':producer') (was testFixturesImplementation)
```

This suggests to first move a dependency away from 'testFixturesImplementation' to 'extraImplementation' and afterwards add it back to 'testFixturesImplementation'.

Instead, the advice should have been:

```
    extraImplementation project(':producer')
```

The issue, as it seems, is the 'else' branch in the 'computeAdvice()' function. This is, if I understand correctly, for moving things that are already on the right scope, but in the wrong "sub-variant" (for lack of a better term) which makes sense for Android build types or flavors (e.g. 'api' vs 'debugApi') or in the special handling of 'test' ('test' vs 'testImplementation').

However, that code now interpreted different feature variants in one project similarly, leading to this weird overall advice: As two feature variants are not connected, moving something away from one of the variants means that you have to add it back afterwards.

The fix is rather crude, but I have to admit that this has grown so complex with the different cases that need to be supported that I would rather keep the change minimal to not break anything else.


src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantInConsumerTestProject.groovy
```diff
@@ -0,0 +1,113 @@
...
+    }
+    builder.withSubproject('consumer') { s ->
+      s.sources = consumerSources
+      s.withBuildScript { bs ->
+        bs.plugins = javaLibrary + Plugin.javaTestFixtures
+        bs.java = Java.ofFeatures(Feature.ofName('extra'))
+        bs.dependencies = [
+          project('api', ':producer'),
+          project('testFixturesImplementation', ':producer'),
+          project('extraImplementation', ':consumer')
```
autonomousapps, line 42: This dependency is saying that the `extra` source set on this project (the consumer project) has a dependency on the main source set, right?

autonomousapps (approved):

src/functionalTest/groovy/com/autonomousapps/jvm/projects/FeatureVariantInConsumerTestProject.groovy
jjohannes: Yes. For `testFixtures` this is added by the test fixtures plugin.

Issue #908, opened 2023-05-29
file names can be too long, exceeding OS limits


mklaehn:
**Build scan link**
https://scans.gradle.com/s/kkyopn5ngqkcs

**Plugin version**
1.20.0

**Gradle version**
Gradle 8.1.1

**(Optional) Android Gradle Plugin (AGP) version**
N/A

**`reason` output for bugs relating to incorrect advice**
N/A

**Describe the bug**
The buildHealth task is failing since v1.20.0 due to the fact that the file name is to long. This occurs on a dependabot version upgrade PR (https://github.com/TweetWallFX/TweetwallFX/pull/1536).
Tested on MacOS and Github Action (see https://github.com/TweetWallFX/TweetwallFX/actions/runs/5111514929/jobs/9188554225?pr=1536)

**To Reproduce**
1. buildHealth task on main project in https://github.com/TweetWallFX/TweetwallFX

**Expected behavior**
task does not fail

**Additional context**
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tweetwallfx-cache:synthesizeDependenciesMain'.
> A failure occurred while executing com.autonomousapps.tasks.SynthesizeDependenciesTask$SynthesizeDependenciesWorkAction
   > /home/runner/work/TweetwallFX/TweetwallFX/cache/build/reports/dependency-analysis/main/dependencies/org.ehcache__ehcache__3.10.8__org.ehcache.modules__ehcache-107__org.ehcache.modules__ehcache-107-provider__org.ehcache.modules__ehcache-api__org.ehcache.modules__ehcache-core__org.ehcache.modules__ehcache-impl__org.ehcache.modules__ehcache-xml__org.ehcache.modules__ehcache-xml-provider.json (File name too long)

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':tweetwallfx-cache:synthesizeDependenciesMain'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:149)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:147)
[...]

PR #1138, opened 2024-02-22
Use a hash in file name for dependencies with capabilities

jjohannes:
If this does not cause any unexpected problems, it's a "simple" solution for https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/908.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/model/Coordinates.kt
+++ b/src/main/kotlin/com/autonomousapps/model/Coordinates.kt
@@ -2,8 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.autonomousapps.model
 
+import com.google.common.hash.Hashing
 import com.squareup.moshi.JsonClass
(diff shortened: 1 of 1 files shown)


autonomousapps (approved):
I guess we're confident that, if the current test suite passes, we're good to go? 😅 

What would you guess is the likelihood of a hash collision 🤔 

I think if there's one other thing I'd like from this, it's an additional output file somewhere that lists the mapping from "full" filename to shortened filename, as a debug assistance in the (probably very unlikely) case of collisions. WDYT? I could just merge this as-is and we could add that later if we think it's necessary.

jjohannes:
I remembered now that the plugin is already using `Guava`, which brings in a number of hash functions. I now use `Hashing.fingerprint2011()` which I think is suited for our purpose as (with my limited understanding) fingerprint hashes are optimized for unique identifiers (very very low chance of collision). But as I said, I don't have much experience. You could pick anothee function from Guava's `Hashing`.

I am good to merge this as it is now.

PR #1139, opened 2024-02-27
feat: allow to force app behavior for pure Java projects

Vampire:
Currently if only the `java` plugin is applied, but not one of a small set of known
application-like plugins, a project is not analyzed.

This PR allows to set a property on the extensions that can force application behavior
even if none of the known application-like plugins is applied.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/ApplicationSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/ApplicationSpec.groovy
@@ -53,4 +53,42 @@ final class ApplicationSpec extends AbstractJvmSpec {
     where:
     gradleVersion << gradleVersions()
   }
+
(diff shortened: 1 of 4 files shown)


autonomousapps:
I'm ok with this conceptually, but I think we can do better than the proposed name. Also, for consistency, I'd prefer it to be exposed as a function. I think there's only one other property that is currently exposed directly. Maybe...

```kotlin
dependencyAnalysis {
  app()
}
```
I'm not convinced that's the best either! So I'm open to pushback if you have other ideas.

I kind of think it should go into an inner DSL, but I don't know what that would look like right now, so I'm content with it as a top-level function inside the outermost DSL.

Vampire:
I don't have a better naming idea right now either, so I changed it to the one you proposed.

autonomousapps (approved):

PR #1141, opened 2024-02-29
fix: Do not dotty for path matching and remove prefix and suffix from binary class name

Vampire:
For path matching the path was dottied too, which feels wrong,
I guess this was just copy & paste and path matching cannot be configured right now anyway,
so just fixed it while changing that code.

The main problem I'm trying to fix with this PR is, that I tried to do `excludeAnnotations("""^\Qorg.spockframework.runtime.model.SpecMetadata\E$""")`
which did not match because the "L" prefix and the ";" suffix from the binary class name are still present.
This PR removes the "L" prefix and the ";" suffix when dottying class names.


Vampire:
Btw. is there a reason the filtering by path is commented out?
Would be handy for me right now to exclude some classes by name pattern, but only within a specific source set.

autonomousapps:
> Btw. is there a reason the filtering by path is commented out? Would be handy for me right now to exclude some classes by name pattern, but only within a specific source set.

I don't recall. I think this was implemented by someone else and they didn't have the bandwidth to implement that component, so left it commented out as a reminder. If you want to contribute it, I'm happy to take a look.

autonomousapps:
Thanks!

Issue #1038, opened 2023-11-17
Dependencies used in Android Manifest are not recognized


Vichy97:
**Plugin version**
1.26.0

**Describe the bug**
I have a theme used in an Android Manifest file that is provided by another module `fixDependencies` will strip this dependency as unused. 
`android:theme="@style/Theme.MyTheme"` In this example, `MyTheme` is provided by another module and is stripped. It's unclear if it is just not recognizing xml resources or if it isn't checking Android Manifest files specifically.  

**Expected behavior**
Modules that provide xml resources that are then used in `AndroidManifest.xml` will not be removed when running `fixDependencies`

autonomousapps:
@Vichy97 DAGP does check the manifest file for resources that might be provided by other modules. It's possible it's missing the theme statement. Could you provide a minimal reproducer?

Vichy97:
🤔 I'm unable to repro this in a minimal project for some reason. It must be something specific about register. I'll try to repro there and provide a commit SHA.

SimonMarquis:
@autonomousapps We are hitting the same issue when referencing a theme with a period character in it (`.`) between two Android library modules.

Here is a PR demonstrating the repro of the issue: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1146

Should I create a new issue stating the actual issue, or is this one good enough?

Also, I tried to check where this issue could be solved but I'm a bit stuck. I've added some details on the draft PR.

PR #1146, opened 2024-03-13
test: Android theme containing a period

SimonMarquis:
This is a reproducer for #1038.

The consumer generates `AttrRef(type = "style", id = "AppTheme.Foo")`.
The producer generates `AttrRef(type = "attr", id = "AppTheme_Foo")`.

https://github.com/autonomousapps/dependency-analysis-gradle-plugin/actions/runs/8265204574/job/22610349535?pr=1146#step:8:393

```
ResSpec > detects theme set in manifest (#gradleVersion AGP #agpVersion) > detects theme set in manifest (Gradle 8.6 AGP 8.4.0-alpha01) FAILED
    org.spockframework.runtime.ConditionFailedWithExceptionError at ResSpec.groovy:131
        Caused by: com.google.common.truth.AssertionErrorWithFacts at ResSpec.groovy:131
```


autonomousapps:
Thank you for this! It was a big help in narrowing down to the root cause. I'm replacing it with my own PR, so will close.

PR #1158, opened 2024-04-08
Output dominator tree results in JSON format including size and total size of deps

mmollaverdi:


src/main/kotlin/com/autonomousapps/tasks/ComputeDominatorTreeTask.kt
```diff
@@ -153,8 +166,19 @@ abstract class ComputeDominatorTreeTask : DefaultTask() {
         nodeWriter = nodeWriter,
       )
 
+      val dataWriter = DominanceTreeDataWriter(
```
mmollaverdi, line 169: Should this only be done if a specific option is set to avoid any perf penalties for consumers which don't care about a JSON output?

graph-support/src/main/kotlin/com/autonomousapps/graph/DominanceTreeWriter.kt
```diff
@@ -63,6 +63,12 @@ public class DominanceTreeWriter<N : Any>(
 
     /** String representation of [node] that will be printed to console. */
     public fun toString(node: N): String
+
+    /** Total size of the node */
+    public fun getTreeSize(node: N): Long?
+
+    /** The size of the node itself */
+    public fun getSize(node: N): Long?
```
autonomousapps, line 71: Why do these two functions return nullable values?

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
```diff
@@ -189,6 +189,12 @@ inline fun <reified T> File.bufferWriteJson(set: T, indent: String = noJsonInden
   }
 }
 
+inline fun <reified T> File.bufferWriteJson(jsonAdapter: JsonAdapter<T>, data: T, indent: String = noJsonIndent) {
```
autonomousapps, line 192: I'm not sure how I feel about passing in the JsonAdapter here, as it is inconsistent with the other functions in this file. Is it necessary?

src/main/kotlin/com/autonomousapps/tasks/ComputeDominatorTreeTask.kt
autonomousapps: No, this is fine.

autonomousapps (commented):
Couple of small questions. I am ok with this conceptually.

graph-support/src/main/kotlin/com/autonomousapps/graph/DominanceTreeWriter.kt
mmollaverdi: `treeSize` is not calculated in lead nodes currently, and is hence nullable. But I guess I can default it to `size`.
`size` itself is also nullable though because it gets a value out of a map: `files[nodes]?.` ([link](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1158/files#diff-9b0b98a87dd28fc6a5a168c6b4c53c15d98d3034e88791b0703683098d38cb39R78)) Should I just default to 0 there?

src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
mmollaverdi: I've just pushed a second commit to make this consistent. The new method is for any generic type with one type parameter. The signature is a bit strange (with the type param only being used in the function body), but it's not too bad.

autonomousapps (approved):
I might tweak this later, but this is good enough for the moment. Thanks!

PR #1160, opened 2024-04-08
feat: allow to exclude ABI by source path

Vampire:
Following up the comment in #1141:
> I don't recall. I think this was implemented by someone else and they didn't have the bandwidth to implement that component, so left it commented out as a reminder. If you want to contribute it, I'm happy to take a look.

I had a quick look at the history.
#158 introduce these changes with the comment that for it to work the source file location is missing up to now and the two todos at https://github.com/autonomousapps/dependency-analysis-gradle-plugin/commit/e4e1d2a5e1221686ad2e9e139ea025b94d0e49ec#diff-44d4684fe90105b8c2f6bfcb9ed566be03e60c3f2464db27b22fd8b0c6f5c207R159-R162 and https://github.com/autonomousapps/dependency-analysis-gradle-plugin/commit/e4e1d2a5e1221686ad2e9e139ea025b94d0e49ec#diff-2bacf6be9cf95dbf682271812bc7338662b3ccaf7952b6880d65d0f89c6fe2e0R74-R75 remained.

The second of these todo which is about missing source location you seem to have fixed in https://github.com/autonomousapps/dependency-analysis-gradle-plugin/commit/7d006caf6fd3aff1ec70cfbd4beab2b050e7599f#diff-2bacf6be9cf95dbf682271812bc7338662b3ccaf7952b6880d65d0f89c6fe2e0L117-L118, but just incompletely, as it only puts the source file name in there as read from the class file, but not the source path as initially intended and necessary for this feature.

This is a first shot at an implementation that realizes the feature. Please have a look. Of course tests are missing, but I'd like to get feedback first, especially as the logic only provides the correct sourcefiles with an heuristic on a best-effort basis. If you totally dislike it and don't have a better idea, it might not be worth putting more effort into it.


src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
```diff
@@ -136,6 +142,21 @@ internal fun getBinaryAPI(
...
+            sourceFilePackageReversed
+              .asSequence()
+              .zip(clazzPackageReversed.asSequence())
+              .takeWhile { (sourceFilePart, clazzPart) -> sourceFilePart == clazzPart }
+              .count()
+          }
+          ?.key
+          ?.invariantSeparatorsPath
+          ?: sourceFileName
+
```
autonomousapps, line 159: Can you explain why we're doing this splitting and reversing?

autonomousapps (commented):
Sorry for the long delay; it's been a busy year. I'm open to this change. With some tests, I'd be happy to see it merged.

autonomousapps:
@Vampire I want to clean up old PRs on this repo, and this hasn't had activity in 6 months. (I recognize part of that was my doing for delaying my initial response.) If there isn't activity on it soon (even just a ping), then I'm going to close it. Thanks for understanding.

src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
Vampire: Ah, sorry, missed to answer.
It is two years ago that I wrote that code, but I think that is the best-effort heuristics I mentioned.
The source files that the task has are the fully-qualified file names.
The source files that are stored in the class file are relative paths from source root.
To not assume anything about the source root of consumer projects I search for the source file with the longest match starting from behind.
So I get the entries where the file name matches (or starts with in case the class file has no debug information and thus has no source file name and it was derived from the class name where the file extension is not known).
And then I search for the longest trailing path match.

autonomousapps (commented):
I think the direction on this looks fine. With some test coverage, I'd be happy to merge this feature.

Vampire:
Running tests became a bit cumbersome for contributors, because it fails to sign graph-support due to missing signatory.
To run the tests I changed `t.enabled = false` in `BaseConventionPlugin` for the `Sign` tasks.
You might consider to change this so that signing is not necessary for running functional tests.

I added a test now that tests the new exclusion similar to the others.

autonomousapps:
> Running tests became a bit cumbersome for contributors, because it fails to sign graph-support due to missing signatory.

Thanks for letting me know. I've had issues around disabling signing in a way that is CC-compliant. I'll try again.

Edit: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1751

src/functionalTest/groovy/com/autonomousapps/jvm/projects/AbiExclusionsProject.groovy
```diff
@@ -35,6 +35,9 @@ final class AbiExclusionsProject extends AbstractProject {
               exclusions {
                 excludeClasses("com\\\\.example\\\\.Main")
                 ignoreGeneratedCode()
+                excludePaths(
+                  'com/example/Other.java'
```
autonomousapps, line 39: Ah, it wasn't clear to me that the path needed to include the file extension. And in retrospect, it _was_ clear to me that `/` was required. (Would it be `\` on Windows?)

1. Is it possible to normalize the path separators so that they're `.`? Or do you think using slashes is better? Alternatively, should the example here use `invariantSeparatorsPath` or something?
2. Should we require file extensions as well? Would it be cleaner to _not_ require the extension, from the end-user perspective?

autonomousapps (commented):
Thanks for the test. It clarified something for me. Left some questions.

I'll work on improving the signing situation for external contributors.

autonomousapps:
I've merged a PR that skips signing when running functional tests.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/AbiExclusionsProject.groovy
Vampire: > Would it be `\` on Windows

Shouldn't be, no.
Where I deal with actual files, I use `invariantSeparatorsPath` and the `com.autonomousapps.internal.asm.tree.ClassNode#name` should also always be slash-separated or other code paths would be wrong already that split it on slash alone.

> Is it possible to normalize the path separators so that they're `.`?

They are normalized to slash currently by using `invariantSeparatorsPath` on file objects and by already being with slash in other parts like `ClassNode#name`. I don't see much value in replacing the slashes again to dots then. And especially in the consumer API (i.e. the `excludePaths` call) I would not use dots, because I as user would never expect that I have to use dots when I have to configure a path. Especially as the user configures a regex for the paths to be excluded where then the dots would need an escaping, Actually in the test it would be more correct to do `'com/example/Other\\\\.java'` just like for the `excludeClasses`.

> Alternatively, should the example here use `invariantSeparatorsPath` or something?

`invariantSeparatorsPath` is on `File`.
What you configure is a regex to match the path.
And it will always be checked against a slashed path, as in regex you would otherwise need up to four back-slashes depending on String literal type just to specify one file separator. Slash is just way more usable.
So I'm not sure where you think `invariantSeparatorsPath` could make sense.

> Should we require file extensions as well? Would it be cleaner to not require the extension, from the end-user perspective?

From end-user perspective I think it makes more sense to have the file extension there, because you configure a source file path to be excluded, not a class name and there I as user would consider the extension as part of. And if I don't want to specify it, I can still do `\.[^.]*` in the regex.

autonomousapps: Thanks for the detailed reply. My main concerns was about the end-user "API", but you're right, the name of this DSL method is `excludedPaths`, not `excludedClasses` or something, so I'm happy with the current implementation.

autonomousapps (approved):

PR #1165, opened 2024-04-11
fix: Fix Windows file separator incompatibility

### LINKED ISSUE #1164: synthesizeDependenciesMain is failing with exception when having a file collection with an absolute file on windows as dependency
**Build scan link**
https://scans.gradle.com/s/2tj64rxgo5xsm

**Plugin version**
1.31.0

**Gradle version**
8.7

**JDK version**
11

**Describe the bug**
If you have
- a file collection
- with an absolute file
- on Windows

as a dependency, then the `graph-compile-nodes.json` has the absolute path as `identifier`.
Following that, the `synthesizeDependenciesMain` task fails with `FileNotFoundException`
because it tries to create the file `build\reports\dependency-analysis\main\dependencies\C__\full\path\to\bar\baz.json` where the intermediate `C__\full\path\to\bar` directories are not present which makes it fail.

**To Reproduce**
```kts
plugins {
    application
    id("com.autonomousapps.dependency-analysis") version "1.31.0"
}

dependencies {
    compileOnly(files(file("bar/baz")))
}
```
Be on Windows and run
```bash
./gradlew synthesizeDependenciesMain
```

**Expected behavior**
No exception, everything working.

**Additional context**
Generally, it is of course not a good idea to use `files` or `fileTree` as dependencies, but sometimes you do not have a real choice.
The reproducer is a highly condensed MCVE after trying to figure out what exactly makes it break for 2 days.
In the real project the `com.klinec.gradle.javacard` plugin is applied which adds such a dependency automatically and you cannot prevent it to do so.

Vampire:
Fixes #1164

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
@@ -24,6 +24,7 @@ import org.gradle.api.internal.artifacts.DefaultProjectComponentIdentifier
 import org.gradle.api.provider.Provider
 import org.gradle.internal.component.local.model.OpaqueComponentArtifactIdentifier
 import org.gradle.internal.component.local.model.OpaqueComponentIdentifier
+import java.io.File
 import java.io.Serializable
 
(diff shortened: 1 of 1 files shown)


autonomousapps:
Thanks!

PR #1166, opened 2024-04-13
Explicit memory settings

seregamorph:
We build the project in a container with 10G memory and in most cases it hangs in a random place (occasionally it passes successfully).
The main assumption here is default memory settings resolved from environment. After some experiments I've found that this simple change stabilizes the build (reproducible).


autonomousapps (approved):
Thanks!

Issue #1172, opened 2024-04-19
File-level private vals of a type from an external dependency result in that dependency requiring to be an api dependency


travisMiehm:
**Plugin version**
1.31.0

**Gradle version**
8.7

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
Kotlin 1.9.23

**(Optional) `reason` output for bugs relating to incorrect advice**
```
Source: main
------------
* Exposes 1 class: some.externa.dependency.ExternalEnum (implies api).
```

**Describe the bug**
When declaring file-level `private val` variables which are typed with some type from some dependency, that dependency is required to be an `api()` dependency, even though that file-level variable is only used in non-public contexts, such as a constructor argument to an `internal class`
```
private val FOO_ENUMS = ExternalEnum.entries.filter { it.name.startsWith("Foo") }

internal class MyClass(private val foos: Set<ExternalEnum> = FOO_ENUMS)
```

**To Reproduce**
In on project, declare some type
```
enum class ExternalEnum {
  FOO_ONE
  FOO_TWO
  BAR
}
```
In a second project, declare a class and a file-level `private val` variable
```
private val FOO_ENUMS = ExternalEnum.entries.filter { it.name.startsWith("Foo") }

internal class MyClass(private val foos: Set<ExternalEnum> = FOO_ENUMS)
```

Now run
`./gradlew :second-project:reason --id :first-project`

**Expected behavior**
The output of `:reason` should be that `ExternalEnum` is a used class, implying `implementation`

**Additional context**
An existing work around is to move the file-level `private val` variables to the companion object of the class, eg
```
internal class MyClass(private val foos: Set<ExternalEnum> = FOO_ENUMS) {
  companion object {
    private val FOO_ENUMS = ExternalEnum.entries.filter { it.name.startsWith("Foo") }
  }
}
```
This class will result in the command giving the expected output.

autonomousapps:
Thanks for the report.

autonomousapps:
Interestingly enough, this only happens with the exact setup you have. If your internal class instead looks like
```internal class MyClass(private val foos: List<ExternalEnum>)```
then DAGP doesn't advise changing the dependency declaration. I think there's some kind of synthetic bridge class that is actually public in the bytecode. Investigating...

autonomousapps:
I created another repro in my project and decompiled it to Java. Here's what it looks like. You can clearly see the public methods that "expose" the private property.
```java
[...]

PR #1181, opened 2024-05-01
test: test validating that Issue-1172 is a non-issue.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1172

This test passes 🤔


autonomousapps:
Might be able to make use of https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isSynthetic-- to filter out the synthetic access properties for purposes of ABI analysis.

PR #1187, opened 2024-05-13
Update test kit to use minSdk, targetSdk, compileSdk, and an equals after applicationId and namespace

j-roskopf:
I was using testkit to create an android project locally, and I didn't see anyway to overwrite the default values for the `android` block.


autonomousapps (commented):
Thanks for the contribution. In which version of AGP were these property names updated?

j-roskopf:
I'm not entirely sure the exact version! I created a new sample project and used AGP 7.0 and saw the properties were still deprecated, so I'm assuming pre 7.0 they were deprecated

autonomousapps:
Thanks!

j-roskopf:
@autonomousapps do you have an inkling to when this would be released?

autonomousapps:
> @autonomousapps do you have an inkling to when this would be released?

Can't make any promises, so let's go with Soon ™️

Issue #844, opened 2023-01-12
Use Generated Project Accessors


eygraber:
**Is your feature request related to a problem? Please describe.**

Usages of Project accessors (e.g. `implementation(projects.foo)`) are detected by the plugin, but when it comes to fixing issues with them, or printing advice, they don't work so well.

**Describe the solution you'd like**

Looking for a match in the project accessors if they are enabled. [I've filed an issue with Gradle](https://github.com/gradle/gradle/issues/23515) to support this without violating project isolation.

```
subprojects.forEach { subproject ->
  val projectAccessor = CaseFormat.LOWER_HYPHEN.to(CaseFormat.LOWER_CAMEL, subproject.path.replace(':', '.'))
  map.put(subproject.path, "projects$projectAccessor")
}
```

**Describe alternatives you've considered**

The alternative would be for users to do this in their own projects.

**Additional context**
Similar to #794

eygraber:
I mean that usages of `projects.foo` get replaced with `project(":foo")`

autonomousapps:
I wonder if you can use https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/794#issuecomment-1349005040 as a workaround?

eygraber:
One issue with the workaround that I posted  (which is based on [#794](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/794#issuecomment-1349005040)) is that the value in `map` gets wrapped in `project()`.

So if I have a project `foo`, and access it with `projects.foo`, the plugin uses `project(projects.foo)`, which doesn't compile because `project` doesn't take a `ProjectDependency` (which is the type of `projects.foo`).

dees91:
> One issue with the workaround that I posted (which is based on [#794](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/794#issuecomment-1349005040)) is that the value in `map` gets wrapped in `project()`.
> 
> So if I have a project `foo`, and access it with `projects.foo`, the plugin uses `project(projects.foo)`, which doesn't compile because `project` doesn't take a `ProjectDependency` (which is the type of `projects.foo`).

I have a similar problem, in my multi-module project we use version catalog and declare dependencies as:
```kotlin
implementation(projects.common.a)
implementation(projects.feature.foo)
```
I came to a similar point and tried to use map:
```kotlin
subprojects.forEach {
[...]

PR #1189, opened 2024-05-18
Add support for Typesafe Project Accessors

ivanalvarado:
#### Description
- Adds support for typesafe project accessors through extension configuration
- Addresses: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/844

#### Usage
```groovy
dependencyAnalysis {
   projectProperties {
     // (Optional) Specify whether to print advice in typesafe-project-accessors format
     // Example: project(":foo:bar:baz") vs. projects.foo.bar.baz
     useTypesafeProjectAccessors(true)
   }
 }
```

#### Example
If this property is set to `true`, the console report will output:

```terminal
  implementation projects.account.impl (was api)
  implementation projects.account.public (was api)	
```
instead of

```terminal
  implementation project(':account:impl') (was api)	
  implementation project(':account:public') (was api)
```


src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
```diff
@@ -57,6 +58,11 @@ open class DependencyAnalysisExtension @Inject constructor(project: Project) : A
     action.execute(issueHandler)
   }
 
+  /** Customize project properties. See [ProjectHandler] for more information. */
+  fun projectProperties(action: Action<ProjectHandler>) {
+    action.execute(projectHandler)
+  }
```
autonomousapps, line 64: Not sure about this method name or using this as the access point.

src/main/kotlin/com/autonomousapps/DependencyAnalysisSubExtension.kt
```diff
@@ -42,6 +43,10 @@ open class DependencyAnalysisSubExtension(project: Project) : AbstractExtension(
     action.execute(abiHandler)
   }
 
+  fun projectProperties(action: Action<ProjectHandler>) {
+    action.execute(projectHandler)
+  }
```
autonomousapps, line 48: Same comment as earlier, unsure about this.

src/main/kotlin/com/autonomousapps/extension/ProjectHandler.kt
```diff
@@ -0,0 +1,27 @@
...
+ * dependencyAnalysis {
+ *   projectProperties {
+ *     // (Optional) Specify whether to print advice in typesafe-project-accessors format
+ *     // Example: project(":foo:bar:baz") vs. projects.foo.bar.baz
+ *     useTypesafeProjectAccessors(true)
+ *   }
+ * }
+ * ```
+ */
+open class ProjectHandler @Inject constructor(objects: ObjectFactory) {
```
autonomousapps, line 20: Should be `abstract`, not `open`.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -10,6 +10,8 @@ internal class AdvicePrinter(
   private val dslKind: DslKind,
   /** Customize how dependencies are printed. */
   private val dependencyMap: ((String) -> String?)? = null,
+  private val useTypesafeProjectAccessors: Boolean
+
```
autonomousapps, line 14: Use a trailing comma, and remove the empty line.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -20,8 +22,8 @@ internal class AdvicePrinter(
 
   fun gav(coordinates: Coordinates): String {
     val quotedDep = coordinates.mapped()
-    return when (coordinates) {
```
autonomousapps, line 23: I think we can still use the `return` form.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -46,6 +48,26 @@ internal class AdvicePrinter(
         }  }}"
       }
     }
+    return dependency
+  }
+
+  private fun getProjectFormat(quotedDep: String): String {
+    return if (useTypesafeProjectAccessors) {
+      if (dslKind == DslKind.KOTLIN) {
+        "projects${quotedDep.replace(':', '.').replace("\"", "").kebabToCamelCase()}"
```
autonomousapps, line 57: Is the prefix always `projects`?

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -46,6 +48,26 @@ internal class AdvicePrinter(
...
+      } else {
+        "projects${quotedDep.replace(':', '.').replace("'", "").kebabToCamelCase()}"
+      }
+    } else {
+      if (dslKind == DslKind.KOTLIN) "project(${quotedDep})" else "project(${quotedDep})"
+    }
+  }
+
+  private fun String.kebabToCamelCase(): String {
+    val pattern = "-[a-z]".toRegex()
```
autonomousapps, line 67: Does this regex work in all possible cases? Does gradle only permit ASCII characters or is a wider character range available?

src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilder.kt
```diff
@@ -11,11 +11,12 @@ internal class ProjectHealthConsoleReportBuilder(
   dslKind: DslKind,
   /** Customize how dependencies are printed. */
   dependencyMap: ((String) -> String?)? = null,
+  useTypesafeProjectAccessors: Boolean
```
autonomousapps, line 14: nit: missing trailing comma

src/main/kotlin/com/autonomousapps/tasks/GenerateBuildHealthTask.kt
```diff
@@ -72,7 +75,8 @@ abstract class GenerateBuildHealthTask : DefaultTask() {
           val report = ProjectHealthConsoleReportBuilder(
             projectAdvice = projectAdvice,
             dslKind = dslKind.get(),
-            dependencyMap = dependencyMap.get().toLambda()
+            dependencyMap = dependencyMap.get().toLambda(),
+            useTypesafeProjectAccessors = useTypesafeProjectAccessors.get()
```
autonomousapps, line 79: nit: missing trailing comma (I can see it was already missing it, but let's improve it 😅 )

src/main/kotlin/com/autonomousapps/tasks/RewriteTask.kt
```diff
@@ -66,7 +69,8 @@ abstract class RewriteTask : DefaultTask() {
       advice = projectAdvice.dependencyAdvice.filtered(isUpgrade),
       advicePrinter = AdvicePrinter(
         dslKind = dslKind,
-        dependencyMap = map.toLambda()
+        dependencyMap = map.toLambda(),
+        useTypesafeProjectAccessors = useTypesafeProjectAccessors.get()
```
autonomousapps, line 73: nit: missing trailing comma

src/test/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilderTest.kt
```diff
@@ -21,7 +22,11 @@ internal class ProjectHealthConsoleReportBuilderTest {
     )
     val projectAdvice = ProjectAdvice("dummy", dependencyAdvice, emptySet())
 
-    val consoleText = ProjectHealthConsoleReportBuilder(projectAdvice, DslKind.KOTLIN).text
+    val consoleText = ProjectHealthConsoleReportBuilder(
+      projectAdvice = projectAdvice,
+      dslKind = DslKind.KOTLIN,
+      useTypesafeProjectAccessors = false
```
autonomousapps, line 28: nit: missing trailing comma

autonomousapps (changes requested):
Thank you for this contribution!

I feel like this value should simply default to true if the project is using typesafe project accessors. I don't know why we'd want it configurable. Is there a way to just detect that and enable this output automatically in that case?

src/main/kotlin/com/autonomousapps/DependencyAnalysisExtension.kt
autonomousapps: I would be ok with exposing this new property in the top-level DSL object (`dependencyAnalysis`).

If you file (or discover) an issue for auto-detecting this, please link it to this PR.

autonomousapps: Gradle has confirmed to me directly that there is no API to detect this. Nevertheless I think it would be a better user experience if it were automatic. How do you feel about parsing the settings script with regex, looking for that specific line that enables the feature?

I also wonder if we can inspect the buildscript classloader for a specific jar or class file, where these accessors would live?

ivanalvarado: Sounds good. Yeah, my worry was that Gradle would remove the need for the `enableFeaturePreview` syntax once typesafe project accessors were promoted to stable.

I'll go ahead and move the property to top level.

alllex:
Ideally, the error should provide the suggestion compatible with the corresponding build script DSL, i.e. Groovy vs Kotlin.

For Kotlin DSL it should be:
```
  implementation(projects.account.impl) (was api)
  implementation(projects.account.public) (was api)	
```

alllex:
Given that it's not possible to detect whether type-safe accessors are enabled and that it might not be easy to detect the correct DSL, I suggest adding those as formatting options in the plugin configuration.

This will already be immediately useful for projects using the plugin because the errors will provide a better UX.

On that note, even if the type-safe accessors become stable and are ON by default, the users might still prefer the stringy project coordinates. This means they would still prefer to see the errors displayed in the format matching theirs. Thus, the plugin would still require configurability in this aspect.

autonomousapps:
> Given that it's not possible to detect whether type-safe accessors are enabled and that it might not be easy to detect the correct DSL, I suggest adding those as formatting options in the plugin configuration.

The plugin currently does detect whether the project is using Groovy or Kotlin based on the file extension (.gradle or .gradle.kts), and formats the output accordingly. All the tests here use Groovy DSL for historical reasons.

eygraber:
Is there anything that still needs to be done here?

autonomousapps:
At least it needs to be rebased and the conflicts fixed.

autonomousapps:
Here's the [error](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/actions/runs/13470762648/job/37643994800?pr=1189#step:6:197):
```
> Task :compileTestKotlin FAILED
e: file:///home/runner/work/dependency-analysis-gradle-plugin/dependency-analysis-gradle-plugin/src/test/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriterTest.kt:60:93 No value passed for parameter 'useTypesafeProjectAccessors'
```
I haven't looked at the latest code yet, but I would be suspicious of something that was suddenly a required parameter. There should be a default value, I think?

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
ivanalvarado: From Gradle's documentation, snakecase is also supported: https://docs.gradle.org/current/userguide/declaring_dependencies_basics.html#sec:type-safe-project-accessors

> Project accessors are based on project paths. For instance, the path `:commons:utils:some:lib` becomes `projects.commons.utils.some.lib`, while kebab-case (`some-lib`) and snake-case (`some_lib`) are converted to camel case: `projects.someLib`.

I've also seen examples of Gradle projects with `[0-9]` in the module name like this: https://github.com/cdsap/ExperimentKopy/blob/master/layer_1/module_1_111/build.gradle.kts

I extended support for these two cases and added tests here: [93ca03c](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1189/commits/93ca03c65bea2320ad9f29a918c1230325af1ae2)

src/main/kotlin/com/autonomousapps/DependencyAnalysisSubExtension.kt
ivanalvarado: The `useTypesafeProjectAccessors` property has been moved to a top-level configuration.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -48,6 +49,25 @@ internal class AdvicePrinter(
...
+      if (dslKind == DslKind.KOTLIN) "project(${quotedDep})" else "project(${quotedDep})"
+    }
+  }
+
+  private fun String.kebabOrSnakeToCamelCase(): String {
+    val pattern = "[-_][a-z0-9]".toRegex()
+    return replace(pattern) {
+      it.value.last().uppercaseChar().toString()
+    }
+  }
```
autonomousapps, line 69: Is there a way to iterate over all of the typesafe accessors, similarly to how we can iterate over all version catalog aliases?

ivanalvarado: After some investigation, I haven't been able to find any Gradle APIs that provide the _generated_ accessors. Is the motivation that we won't be making assumptions of the generated accessor format?

I assume you're hinting at building a map of typesafe accessors similar to the [Version Catalogs](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/extension/DependenciesHandler.kt#L152-L164)? I'm thinking, what we can do instead is leverage `subprojects`and mimic how Hinge generates accessors with their plugin: https://github.com/Hinge/project-accessors/blob/main/project-accessors/src/main/kotlin/co/hinge/gradle/projectaccessors/ProjectAccessorsGenerator.kt

ivanalvarado: Asked in ASG: https://androidstudygroup.slack.com/archives/C0SFT151T/p1740637883514059

autonomousapps: > I assume you're hinting at building a map of typesafe accessors similar to the [Version Catalogs](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/extension/DependenciesHandler.kt#L152-L164)?

Yes, that's exactly right. I think it would be more maintainable to use that same infrastructure, since the project accessors feel like the same kind of thing as the version catalog accessors.

ivanalvarado: Following up here, I was unable to find a public API that provides these generated accessors for the project. I spoke to Nicklas Ansman, the author of https://github.com/Hinge/project-accessors, and he manually transforms the project paths to accessor notation: https://github.com/Hinge/project-accessors/blob/main/project-accessors/src/main/kotlin/co/hinge/gradle/projectaccessors/ProjectAccessorsGenerator.kt

I've made a feature request to Gradle about this: https://github.com/gradle/gradle/issues/32602

In the meantime, what do you say we proceed with our current approach for this first iteration? Then, we make the improvement as soon as Gradle provides support for this feature.

src/main/kotlin/com/autonomousapps/AbstractExtension.kt
```diff
@@ -27,6 +27,8 @@ abstract class AbstractExtension @Inject constructor(
   // One instance of this per project
   internal val issueHandler: IssueHandler = objects.newInstance(dslService)
 
+  open var useTypesafeProjectAccessors: Boolean = false
```
autonomousapps, line 30: This should be a `Property<Boolean>`, and for consistency should be `internal`. This plugin exposes properties to users via functions, not directly.

src/test/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilderTest.kt
```diff
@@ -40,6 +42,60 @@ internal class ProjectHealthConsoleReportBuilderTest {
...
+      Advice.ofChange(Coordinates.of(":layer_1:module_0-50"), "implementation", "api")
+    )
+    val projectAdvice = ProjectAdvice("dummy", dependencyAdvice, emptySet())
+    val consoleText = ProjectHealthConsoleReportBuilder(
+      projectAdvice = projectAdvice,
+      postscript = postscript,
+      dslKind = DslKind.GROOVY,
+      useTypesafeProjectAccessors = true,
+    ).text
+    assertThat(consoleText.decolorize()).isEqualTo(
```
autonomousapps, line 59: nit: please insert a single blank line here. Assertions should be separated from the rest of a test.

src/test/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilderTest.kt
```diff
@@ -40,6 +42,60 @@ internal class ProjectHealthConsoleReportBuilderTest {
...
+      Advice.ofChange(Coordinates.of(":layer_1:module_0-50"), "implementation", "api")
+    )
+    val projectAdvice = ProjectAdvice("dummy", dependencyAdvice, emptySet())
+    val consoleText = ProjectHealthConsoleReportBuilder(
+      projectAdvice = projectAdvice,
+      postscript = postscript,
+      dslKind = DslKind.KOTLIN,
+      useTypesafeProjectAccessors = true,
+    ).text
+    assertThat(consoleText.decolorize()).isEqualTo(
```
autonomousapps, line 86: whitespace

autonomousapps (changes requested):
I'm ok with this approach, but there are a couple of small things that should be fixed before we can merge it. Thank you!

src/main/kotlin/com/autonomousapps/DependencyAnalysisSubExtension.kt
```diff
@@ -36,6 +36,11 @@ abstract class DependencyAnalysisSubExtension(
 
   private val path = project.path
 
+  @Suppress("UNUSED_PARAMETER")
+  fun useTypesafeProjectAccessors(enable: Boolean) {
+    throw OperationNotSupportedException("Typesafe Project Accessors configuration must be declared in the root project only")
+  }
+
```
ivanalvarado, line 43: @autonomousapps I feel like this property is a project-wide setting because their enabled in one place, `settings.gradle`, and are immediately available for the entire project. Similar to `structure`, if the user tries to configure this property at the subproject level, we fail letting them know to declare in root gradle script.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -996,6 +996,7 @@ internal class ProjectPlugin(private val project: Project) {
     val project = this
     val theProjectPath = path
     val paths = NoVariantOutputPaths(this)
+    val dependencyAnalysisExtension = project.rootProject.extensions.getByType(DependencyAnalysisExtension::class.java)
```
ivanalvarado, line 999: @autonomousapps Let me know your thoughts on this. The motivation for referencing `DependencyAnalysisExtension` here is, if we're forcing user to only be able to configure this property at root-level we need to be able to fetch that value for subproject tasks.

src/test/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilderTest.kt
ivanalvarado: [074b4e3](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1189/commits/074b4e31251302ae6eb84414bd7520fe6a2741af)

src/main/kotlin/com/autonomousapps/AbstractExtension.kt
ivanalvarado: [5700dfb](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1189/commits/5700dfb67b290ed20739ba1debf19d82dae4783d)

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
ivanalvarado: Actually, I need to revisit this as there's a failing test: https://scans.gradle.com/s/tuu4lkk5fvw6q/tests/task/:functionalTest/details/com.autonomousapps.jvm.SubprojectSpec/plugin%20doesn't%20have%20to%20be%20applied%20to%20root%20project%20(Gradle%208.12.1)?focused-exception-line=0-11&top-execution=1

I'm wondering if there's a way for the sub extension to inherit from the root extension?

autonomousapps:
@ivanalvarado sorry for the delay. I decided to grab your code and push it in a separate PR at https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1414. That is now green and I'll be merging it.

Issue #947, opened 2023-08-29
Plugin silently gives wrong advices when different versions of a dependency are used in different source sets of the same project


jjohannes:
**Build scan link**

https://scans.gradle.com/s/lyx5rjpyqfnhu/console-log

**Plugin version**

`1.21.0`

**Gradle version**

`7.6.1`

**`reason` output for bugs relating to incorrect advice**

https://scans.gradle.com/s/lyx5rjpyqfnhu/console-log?page=1#L304

**Describe the bug**

If there are different versions of a dependencies for different source sets (in the example `junit-jupiter-api:5.9.0` vs  `junit-jupiter-api:5.9.1`) only one of the Jars is analyzed. In the example `5.9.0` is used. Then this does not match in cases where `5.9.1` is in the dependency graph and the plugin assumes "an empty Jar" (for 5.9.1). As result, the plugin gives the wrong advice to remove the dependencies to `junit-jupiter-api` in such cases. In the example, it advices to remove `testImplementation` although `src/test/java` is using it.

https://scans.gradle.com/s/lyx5rjpyqfnhu/dependencies?dependencies=junit-jupiter-api&expandAll&projectFilter=WyI6YXBwIl0

**To Reproduce**
I am not yet sure if this happens every time in such a situation or if there are some special circumstances causing this. I'll provide a reproducer later. 

**Expected behavior**

I don't think this necessarily needs to be supported in all cases. Having consistent versioning in one build is what is desired in most cases. It's just easily missed. IMO, it would also be fine to fail in some form if the plugin can detect such a case it cannot handle.

Just the wrong advice you get right now is confusing.

autonomousapps:
I suspect this might be fixed in 1.24.0. I found a bug relating to version skew with external dependencies, in a different context than this, and fixed it. Can you test?

jjohannes:
I can confirm that the problem is gone with `1.24.0` in the project setup where I faced this

jjohannes:
Sorry I was mistaken. Problem still exists. New scan with 1.24.0:

https://scans.gradle.com/s/dvnrv3xf5xehy/console-log?page=1#L303

autonomousapps:
@jjohannes here's a reproducer, I think: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1191

jjohannes:
Thank's for fixing this!

PR #1191, opened 2024-05-23
test: reproduces issue 947

autonomousapps:
https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/947

This test currently fails, producing this erroneous advice:
```
> Task :buildHealth
Advice for :proj
  extraApi 'com.squareup.okio:okio:3.9.0'
```

DAGP is suggesting removing the dep on the "extra" source-set's dependency. The main source set depends on v3.8.0. This would lead to a compilation error.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/CustomSourceSetSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/CustomSourceSetSpec.groovy
@@ -191,9 +191,25 @@ final class CustomSourceSetSpec extends AbstractJvmSpec {
     )
   }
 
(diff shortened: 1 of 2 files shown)


jjohannes:
There was a `.withSourceSet("extra")` missing in the example setup. I am not sure if this reproduces (only) https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/947. There seems to be another issue with `api` vs `implementation` and multiple source sets.

Right now the test gives no advice, although it should give two. But this also happens if I use `com.squareup.okio:okio:3.8.0` (same version) in both dependencies.

MGaetan89:
@autonomousapps #947 seems to have been fixed with #1230. Is this PR still needed?

Issue #1091, opened 2023-12-30
Support Isolated Projects


ZacSweers:
**Is your feature request related to a problem? Please describe.**
Currently, project isolation doesn't appear to be supported (or at least it fails the build when project isolation is enabled). This is with `1.28.0` and Gradle 8.5.

```
org.gradle.api.InvalidUserCodeException: Cannot access project ':' from project ':app'
	...
	at com.autonomousapps.DependencyAnalysisExtensionKt.getExtensionOrNull(DependencyAnalysisExtension.kt:92)
	at com.autonomousapps.DependencyAnalysisPlugin.checkPluginWasAppliedToRoot(DependencyAnalysisPlugin.kt:76)
	at com.autonomousapps.DependencyAnalysisPlugin.apply(DependencyAnalysisPlugin.kt:27)
	at com.autonomousapps.DependencyAnalysisPlugin.apply(DependencyAnalysisPlugin.kt:22)
```

**Describe the solution you'd like**

Project isolation support

**Describe alternatives you've considered**

**Additional context**

jjohannes:
This already works for me if:
- The project does not use the `testFixtures(...)` notation - see https://github.com/gradle/gradle/issues/27872#issuecomment-2151775051 and #1198
- I turn off auto apply: `dependency.analysis.autoapply=false`
- ...and apply the plugin individually to each subproject

The improvement I would like to see next is to not having to use the `dependency.analysis.autoapply` property, but instead having a dedicated "root project" plugin – e.g. `com.autonomousapps.dependency-analysis-root` – only for the root project that never does the autoapply.

As a nice-to-have, I would also like to see the plugin working without applying anything to the root. Then you won't have the root tasks (buildHealth) but can still check each project individually (projectHealth) which is sufficient for certain setups (in my experience).

matejdro:
I've tried above and the main problem seems to be that with `autoapply=false`, bundles do not seem to be propagating down to subprojects and I cannot declare them on the subprojects (I get `Dependency bundles must be declared in the root project only` error).

Which means that with `autoapply=false` bundles are effectively not working.

autonomousapps:
Regarding this issue in general, I was under the (perhaps mistaken) impression that DAGP already supports IP. A Gradle eng confirmed it several months ago -- but of course it's a pre-incubating feature and maybe the contract has already shifted to be stricter in some way. **As such, I will close this issue.**

[...]

PR #1198, opened 2024-06-06
Avoid cross-project access of the 'group' property

jjohannes:
We need to do special handling for "capabilities without group" already. So this works along these lines and makes the plugin (more) compatible with project isolation (#1091).


jjohannes:
This change helps with using `testFixtures(...)`, but after creating the PR I realized that even Gradle core cannot handle it: https://github.com/gradle/gradle/issues/27872#issuecomment-2151775051

Maybe this change is not needed in the end.

PR #1200, opened 2024-06-12
Sort output of findDeclarations

seregamorph:
## Reason behind this change

The result of `findDeclarations` task is reproducible if inputs are not changed.
But while debugging there is needed to compare task output (JSON pretty print reformat first) between different executions and it's not convenient to compare non-ordered elements.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
@@ -114,6 +114,8 @@ abstract class FindDeclarationsTask : DefaultTask() {
             )
           }
         }
+        .sortedWith(compareBy<Declaration> { it.configurationName }
+          .thenComparing { it -> it.identifier })
         .toSet()
     }
   }


src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
```diff
@@ -114,6 +114,8 @@ abstract class FindDeclarationsTask : DefaultTask() {
             )
           }
         }
+        .sortedWith(compareBy<Declaration> { it.configurationName }
+          .thenComparing { it -> it.identifier })
```
seregamorph, line 118: `it ->` part cannot be omitted here as `thenComparing` is an overloaded method

src/main/kotlin/com/autonomousapps/tasks/FindDeclarationsTask.kt
```diff
@@ -114,6 +114,8 @@ abstract class FindDeclarationsTask : DefaultTask() {
             )
           }
         }
+        .sortedWith(compareBy<Declaration> { it.configurationName }
```
seregamorph, line 117: complementary with line 99 in this source file:
```
// We sort the map so that the task input property is consistent in different environments
return DeclarationContainer(mapping.toSortedMap())
```

PR #1201, opened 2024-06-13
Sort output of graphView

seregamorph:
Similar to https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1200

## Reason behind this change

The result of `graphViewX` tasks should be reproducible if inputs are not changed.
While debugging there is needed to compare task output (JSON pretty print reformat first) between different executions and it's not convenient to compare non-ordered elements.


seregamorph:
@autonomousapps please delete this PR via GitHub support (I don't have permissions)
https://support.github.com/request?q=delete+pull+request

autonomousapps:
Support request filed.

PR #1205, opened 2024-06-19
JSON pretty print option

seregamorph:
## Change
Add System property `"com.autonomousapps.pretty-json"` to do pretty print in JSON task outputs.

## Motivation
Tasks produce (mostly) JSON outputs which are friendly for text comparison between executions - either with changes in set of project dependencies or in the logic of the dependency-analysis plugin itself. But by default they are single-line JSONs, hence we need an options to pretty-print (a.k.a. indent) it.

Note: the task outputs are reproducible (once this is merged https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1202 )

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/moshi.kt
@@ -117,7 +117,7 @@ inline fun <reified K, reified V> File.bufferWriteJsonMap(
   indent: String = noJsonIndent
 ) {
   JsonWriter.of(sink().buffer()).use { writer ->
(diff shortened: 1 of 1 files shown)


autonomousapps (changes requested):
I apologize for not responding to this. I don't think we want this, and I'll be closing the PR I'll note there are various ways to pretty-print the json once it's already been written to disk. In the IDE, you can `cmd-opt-l`, or you could use [lesspipe](https://lesspipe.org/) for CLI viewing. `jq` also works. Actually printing the json to disk like this just wastes space.

PR #1225, opened 2024-07-16
Test against AGP 8.7.0-alpha06

MGaetan89:
This PR updates the 8.6.0 branch to rc01, and adds the 8.7.0 branch of AGP to the test matrix.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
@@ -21,7 +21,7 @@ abstract class AbstractAndroidSpec extends AbstractFunctionalSpec {
   protected static final AGP_8_4 = AgpVersion.version('8.4.2')
   protected static final AGP_8_5 = AgpVersion.version('8.5.2')
   protected static final AGP_8_6 = AgpVersion.version('8.6.0-rc01')
-  protected static final AGP_8_7 = AgpVersion.version('8.7.0-alpha04')
+  protected static final AGP_8_7 = AgpVersion.version('8.7.0-alpha05')
 
   protected static final AGP_LATEST = AGP_8_7
 
--- a/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
@@ -20,8 +20,8 @@ abstract class AbstractAndroidSpec extends AbstractFunctionalSpec {
   protected static final AGP_8_3 = AgpVersion.version('8.3.2')
   protected static final AGP_8_4 = AgpVersion.version('8.4.2')
   protected static final AGP_8_5 = AgpVersion.version('8.5.2')
-  protected static final AGP_8_6 = AgpVersion.version('8.6.0-beta02')
-  protected static final AGP_8_7 = AgpVersion.version('8.7.0-alpha02')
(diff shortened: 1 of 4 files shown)


MGaetan89:
It looks like there are failing tests with AGP 8.7.0. I'll have a closer look at it this weekend.

autonomousapps:
> do you think it would make sense to extract the test matrix out of the project's code, into the GitHub Actions workflow file?

I used to do this, ages and ages ago. I have the current setup because I wanted the full matrix encapsulated in the build system so I could easily run all the tests locally without having to invoke a bash script. This was before there were so many tests I risked waiting till the heat death of the universe for the suite to complete (I never run the full suite locally anymore).

tl;dr I'm open to a change, but it'll be a higher barrier to merge.

src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
```diff
@@ -41,6 +42,7 @@ abstract class AbstractAndroidSpec extends AbstractFunctionalSpec {
     AGP_8_4,
     AGP_8_5,
     AGP_8_6,
+    AGP_8_7,
```
autonomousapps, line 45: The problem is this is too many versions. CI can't handle this load. We should test against the min version (8.0), the current stable version (8.5.1), and maybe also the latest alpha.

autonomousapps (changes requested):

src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
MGaetan89: Thanks for the tip!
I've updated the PR to test only against AGP 8.0 (min supported stable version), 8.5 (max supported stable version), 8.7 (latest alpha version).

MGaetan89: It's still failing. I think it was fine running the tests for every AGP versions, but there is probably some breaking change in AGP 8.7

MGaetan89: It looks like AGP 8.7 requires Gradle 8.9. So I've updated the project to use Gradle 8.9 (previously 8.7) and made the CI run tests for every AGP version again.
Let me know if you still prefer to keep only the min stable version, max stable version and the latest alpha version.

MGaetan89: It seems that everything is working fine now.
- Is it ok to upgrade to Gradle 8.9, or do you prefer a separate PR?
- Should I keep the tested AGP versions list as it is now? Or only the three versions you suggested?

autonomousapps: I think updating to Gradle 8.9 would be better in a separate PR.

For this PR, you could just update the AGP/Gradle compatibility matrix, and then it won't try to run the test against an incompatible combination.

MGaetan89: I've created the PR to update Gradle here: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1236

I'll update this PR soon too.

MGaetan89: It should be good now. Let me know if I missed something.

autonomousapps (approved):
LGTM

PR #1229, opened 2024-07-25
Fix `test.yml` workflow reporting

MGaetan89:
This PR fixes a small issue introduced in https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1218

Currently, the `Maybe add Build Scan URLs as PR comment` step may fail with the following error:
Screenshot 2024-07-25 at 13 43 09

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,13 +91,14 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               body: `❌ ${{ github.workflow }} workflow failed:
-            
+
                 * Root buildHealth         : ${{ steps.gradle-buildhealth.outputs.build-scan-url }}
-                * TestKit buildHealth      : ${{ steps.gradle-testkit-buildhealth.outputs.build-scan-url }}`
+                * TestKit buildHealth      : ${{ steps.gradle-testkit-buildhealth.outputs.build-scan-url }}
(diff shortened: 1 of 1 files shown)


autonomousapps:
I actually think it fails because I updated the workflows and no longer have identifiers. The core gradle action involved now has support for the thing I was doing manually, and we just need to update the workflows to usethat.

MGaetan89:
Unless I misunderstood what you meant, each step still has an identifier. For example, for the `Execute buildHealth for main project` step:
https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/79e83f9195450090a9ada6d75b5bdb034a1cd861/.github/workflows/pr.yml#L57

Do you have any pointers regarding what should be used from the Gradle action instead?

autonomousapps (approved):

autonomousapps (approved):

MGaetan89:
@autonomousapps I've updated #1238 to see if this fixed the issue, and now there is [a new error](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/actions/runs/10495706780/job/29074715109?pr=1238):
```
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/actions/github-script/v7/dist/index.js:9537:21
Error: Unhandled error: HttpError: Resource not accessible by integration
```

I don't know if it was a temporary issue, or if there is actually something else to fix in that step.

PR #1230, opened 2024-07-25
fix: when building 'dependencyUsages', merge usages of the same dependency at different versions.

autonomousapps:
This resolves an issue when two versions of a dependency are in the various dependency graphs
(e.g., main and test). Without this, we were getting erroneous advice to remove used dependencies,
and then correct advice to re-add used-transitive dependencies. Additionally, the 'reason' results
were confusing and incorrect for a similar reason. As the simplest solution, require full GAV when
requesting the 'reason for' some advice in this circumstance.

There is still one minor issue in the reason output, which is it'll print the shortest path to all
versions of a given dep, even when a specific version is requested.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/model/intermediates/Usage.kt
+++ b/src/main/kotlin/com/autonomousapps/model/intermediates/Usage.kt
@@ -4,6 +4,7 @@ package com.autonomousapps.model.intermediates
 
 import com.autonomousapps.model.Advice
 import com.autonomousapps.model.Coordinates
+import com.autonomousapps.model.ModuleCoordinates
 import com.autonomousapps.model.declaration.Bucket
 import com.autonomousapps.model.declaration.Variant
 import com.squareup.moshi.JsonClass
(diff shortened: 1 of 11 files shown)


autonomousapps:
@jjohannes I think this resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/947.

src/main/kotlin/com/autonomousapps/model/intermediates/Usage.kt
```diff
@@ -90,6 +91,29 @@ internal class UsageBuilder(
...
+          put(trace.coordinates, usages)
+          remove(other)
+        }
+
+        // We're done
+        return
+      }
+    }
+
+    // TODO: could this now be simplified to a put? The above handles our complex merging requirements.
```
autonomousapps, line 116: TODO

src/main/kotlin/com/autonomousapps/model/intermediates/Usage.kt
```diff
@@ -90,8 +91,30 @@ internal class UsageBuilder(
...
+
+        // If the new coordinates are "greater than" the other coordinates, add the new and remove the old
+        if (thisVersion > otherVersion) {
+          put(trace.coordinates, usages)
+          remove(other)
+        }
+
+        // We're done
+        return
+      }
```
autonomousapps, line 111: This change here is the fix.

PR #1238, opened 2024-08-14
Remove code specific to Gradle < 8.0

MGaetan89:
According to the following link, AGP 8.0 (which is the minimum supported version by this plugin) requires at least Gradle 8.0. https://developer.android.com/build/releases/gradle-plugin#updating-gradle

This PR removes all the code specific to Gradle versions below 8.0.


src/main/kotlin/com/autonomousapps/internal/android/AndroidGradlePluginFactory.kt
MGaetan89, line 1: This change may be too extreme. Let me know if you'd rather keep this class for flexibility in the future.

MGaetan89:
There's a failing functional JVM test:
```
AbiExclusionsSpec > can exclude custom source set from ABI analysis (#gradleVersion) > can exclude custom source set from ABI analysis (Gradle 8.0) FAILED
    org.spockframework.runtime.ConditionFailedWithExceptionError at AbiExclusionsSpec.groovy:37
        Caused by: com.google.common.truth.AssertionErrorWithFacts at AbiExclusionsSpec.groovy:37
```

autonomousapps (changes requested):
This plugin isn't just for Android, but also JVM projects. It's ok to make changes for AGP 8+, but the rest seems too aggressive.

MGaetan89:
> This plugin isn't just for Android, but also JVM projects. It's ok to make changes for AGP 8+, but the rest seems too aggressive.

Oops, my bad. I overlooked that. So for Android projects we can use Gradle 8.0+, and for JVM project it's Gradle 7.4+, correct?

autonomousapps:
> So for Android projects we can use Gradle 8.0+, and for JVM project it's Gradle 7.4+, correct?

Yup.

MGaetan89:
I won't have time to update this properly, so I'm closing this.
I'll open a new PR if/when I can submit something targeting Android only.

PR #1242, opened 2024-08-19
feat: add 'com.autonomousapps.build-health' settings plugin.

autonomousapps:
feat: add 'com.autonomousapps.build-health' settings plugin.
feat: remove support for auto-apply flag.


testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Plugins.kt
```diff
@@ -6,7 +6,7 @@ import com.autonomousapps.kit.render.Element
 import com.autonomousapps.kit.render.Scribe
 
 public class Plugins(
-  private val plugins: MutableList<Plugin> = mutableListOf(),
+  public val plugins: MutableList<Plugin> = mutableListOf(),
```
autonomousapps, line 9: Kinda gross, can we do better?

PR #1254, opened 2024-08-27
fix: consider dependency unused candidate if only capability is InferredCapability.

autonomousapps:
Needs a test.


src/functionalTest/groovy/com/autonomousapps/jvm/projects/ClasspathConfusionProject.groovy
```diff
@@ -81,7 +79,10 @@ final class ClasspathConfusionProject extends AbstractProject {
     return actualProjectAdvice(gradleProject)
   }
 
-  private final Set<Advice> consumerAdvice = []
+  private final Set<Advice> consumerAdvice = [
+    Advice.ofRemove(projectCoordinates(':producer'), 'implementation')
+  ]
```
autonomousapps, line 84: Oo, turns out I already kind of had a test that showcased the bug, but didn't know it 😅

Issue #1134, opened 2024-02-19
`primary` feature gives advices to add already added dependencies


BraisGabin:
**Plugin version**
1.30.0 (but it was present before too)

**Gradle version**
8.6

**JDK version**
17

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
1.9.22

**(Optional) Android Gradle Plugin (AGP) version**
8.2.2

**Describe the bug**
I have a project like this: `consumer -> unused -> used` ~and `consumer` only uses `:used`~ (this is not important, if consumer uses both the problem is still there). And a bundle like this:

```
bundle('facade') {
  includeDependency(':unused')
  includeDependency(':used')
}
```

I expect not advices. And that works! But when I change the bundle to this:
```
bundle('facade') {
  primary(':unused')
  includeDependency(':used')
}
```

I also expect not advices but I get an advice telling me that `:consumer` should depend on `:unused`. But `:consumer` already depends on `:unused` so the advice shouldn't exist

**To Reproduce**
1. Go to `src/functionalTest/groovy/com/autonomousapps/jvm/projects/BundleProject2.groovy` and edit the line 31 from `includeDependency` to `primary`
2. Run the test

**Expected behavior**
No advices

autonomousapps:
Thanks for the report.

PR #1260, opened 2024-09-13
fix: when setting bundle primary, also 'include' dependency.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1134


src/functionalTest/groovy/com/autonomousapps/jvm/projects/BundleProject2.groovy
```diff
@@ -28,7 +28,9 @@ final class BundleProject2 extends AbstractProject {
           dependencyAnalysis {
             structure {
               bundle('facade') {
-                includeDependency(':unused')
+                // TODO cleanup
+//                includeDependency(':unused')
+                primary(':unused')
```
autonomousapps, line 33: cleanup

PR #1262, opened 2024-09-13
fix: consistently replace '.' with '_' in Android res.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1038
Replaces https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1146


src/main/kotlin/com/autonomousapps/internal/parse/explodeAndroidResSource.kt
```diff
@@ -150,7 +150,7 @@ internal class AndroidManifestParser(
     val file = first
     val parseResult = second
     val applicationName = parseResult.applicationName
-    val theme = AndroidResSource.AttrRef.style(parseResult.theme)
+    val theme = AndroidResSource.AttrRef.style(parseResult.theme.replace('.',  '_')) // TODO: centralize this replace() function
```
autonomousapps, line 153: - [ ] TODO

PR #1271, opened 2024-09-27
fix: improve separation of "annotation" vs "not-annotation" in bytecode analysis.

autonomousapps:
This resolves an issue where an annotation library was inaccurately showing up in the
"add to implementation" advice, while also preserving the correct behavior around
Kotlin typealiases.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/asm.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/asm.kt
@@ -412,11 +412,20 @@ private class MethodAnalyzer(
 private class AnnotationAnalyzer(
   private val logger: Logger,
   private val classes: MutableSet<ClassRef>,
-  private val level: Int = 0
+  private val level: Int = 0,
+  private val arrayName: String? = null
 ) : AnnotationVisitor(ASM_VERSION) {
+
+  private var arraySize = 0
+  private var isTypeAlias = false
+  private val arrayElements = mutableSetOf<ClassRef>()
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/autonomousapps/internal/asm.kt
```diff
@@ -430,18 +439,19 @@ private class AnnotationAnalyzer(
...
+      if (valueString == "alias") {
+        isTypeAlias = true
       }
     }
 
-    log("${indent()}- AnnotationAnalyzer#visit: name=$name, value=(${value?.javaClass?.simpleName}, ${getValue(value)})")
     if (value is String) {
       METHOD_DESCRIPTOR_REGEX.findAll(value).forEach { result ->
-        addClass(result.value, ClassRef.Kind.NOT_ANNOTATION)
+        addClass(result.value, ClassRef.Kind.ANNOTATION)
```
autonomousapps, line 454: This was a bug. But fixing it broke TypeAliasSpec, because that was relying on analyzing the `@Metadata` annotation for detecting typealias usage.

autonomousapps:
[Transient false positive failure](https://scans.gradle.com/s/wivqiol2wzjcu/tests/task/:functionalTest/details/com.autonomousapps.android.NoOpSpec/can%20filter%20unused%20dependencies%20(Gradle%208.9%20AGP%208.7.0-alpha09)?top-execution=1), ignoring.

PR #1277, opened 2024-10-02
fix: include project description json at both external and internal coordinates.

autonomousapps:
@jjohannes fyi


jjohannes:
Thanks for the heads up. I guess if all tests pass, this change should not cause any problems. I vaguely remember that we already were at a point where the information was duplicated. Before the now commented-out code was added. And that that caused some issue. But  other changes now probably compensate for that. Or my memory is wrong. I don't remember details.

If I should run into anythings suspicious, I'll let you know.

PR #1285, opened 2024-10-15
feat: DSL option that opts-in source sets to requiring explicit dependencies.

autonomousapps:


src/functionalTest/kotlin/com/autonomousapps/kit/gradle/dependencies/DependencyProvider.kt
```diff
@@ -9,6 +9,14 @@ class DependencyProvider(
 
   private val pluginUnderTestVersion = AbstractGradleProject.PLUGIN_UNDER_TEST_VERSION
 
+  /*
+   * Common configurations. TODO(tsr): add later.
+   */
```
autonomousapps, line 14: follow-up

Issue #1097, opened 2024-01-10
Enable `:reason` task to get reason for specific gradle capabilities (e.g. `testFixtures`)


Nava2:
**Describe the solution you'd like**
Some form of CLI parameters that enables specifying which gradle capability that is cared about. Such that, only that capability is shown. 

For example, 
```
./gradlew :project:reason --id ':other-project' --capability 'com.example:other-project-test-fixtures'
```

**Describe alternatives you've considered**
Add notation to siginify _which_ capability an output belongs to if there are multiple usages. For example, showing that content comes from `testFixtures` vs. `default`.

Alternately, including all capabilities in the output of just `--id ':other-project'` would work!

**Additional context**

autonomousapps:
I don't entirely follow this request. Could you share a sample of the expected output from such a feature?

Nava2:
This may be user error, but if you have a build that leverages `java-test-fixtures` ([Docs](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures)), it will add a dependency on a project with a capability rather than the normal project. 

If you want to look at the "reason" output, i.e. figure out why that dependency is used in the project, there's no way to request it today. Calling `--id ':other-project'` will either return "unused" if it's not used, or if both `testFixtures(project(":other-project"))` and `project(":other-project")` are used, it'll give the result for `project(":other-project")` leaving no way currently to request to know why the `testFixtures` capability is requested.

Nava2:
Here's a project: https://github.com/Nava2/dependency-analysis-test-fixtures-project-example

```bash
$ ./gradlew :lib:reason --id ":other-project"

> Configure project :
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `1.9.20` that might work differently than in the requested version `1.9.22`.

> Task :lib:reason

----------------------------------------
You asked about the dependency ':other-project'.
There is no advice regarding this dependency.
----------------------------------------

Shortest path from :lib to dependency-analysis-test-fixtures-project-example:other-project for compileClasspath:
:lib
\--- dependency-analysis-test-fixtures-project-example:other-project

Shortest path from :lib to dependency-analysis-test-fixtures-project-example:other-project for runtimeClasspath:
:lib
\--- dependency-analysis-test-fixtures-project-example:other-project

[...]

PR #1291, opened 2024-10-21
feat: enable reason task to get reason for specific gradle capabilities (e.g., test-fixtures).

autonomousapps:
- [x] TODO: update two ReasonSpecs to include full output.

Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1097

Proposed user experience:
```
./gradlew project:reason --id some:dependency --capability test-fixtures
```
As a special case, it accepts either `test-fixtures` or `testFixtures` for the test fixtures capability. Also, it does not require the user to specify the _full_ capability as Gradle encodes it (the full dependency identifier with the suffix `"-<some capability>"`), which seems like it would be cumbersome. We essentially try to match a suffix here (although it's a bit more sophisticated than that).


src/functionalTest/groovy/com/autonomousapps/jvm/ReasonSpec.groovy
```diff
@@ -77,6 +78,26 @@ final class JvmReasonSpec extends AbstractFunctionalSpec {
...
+      '--id', ':producer',
+      '--capability', 'test-fixtures',
+    )
+
+    then:
+    assertThat(result.output).contains("You asked about the dependency 'com.squareup.okio:okio-jvm:3.0.0'.")
+
+    where:
+    gradleVersion << gradleVersions()
+  }
```
autonomousapps, line 99: Iterate on this test.

PR #1299, opened 2024-11-01
Update Kotlin DSL rewriting functionality to use KotlinEditor

wsutina:
## 📝 Description
Updating the `RewritingTask` to use [kotlin-editor](https://github.com/cashapp/kotlin-editor), which contains an updated grammar for more complex Kotlin Gradle build scripts. The grammar used for parsing Groovy build scripts remains unchanged.

Note: this depends on https://github.com/cashapp/kotlin-editor/pull/17

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,7 +15,7 @@ kotlin = "1.9.25"
 
 # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
 kotlineditor-core = "0.10"
-kotlineditor-relocated = "0.10.0-SNAPSHOT"
+kotlineditor-relocated = "0.10.0"
 
 kotlinx-metadata = "0.9.0"
 moshi = "1.15.1"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,4 +1,5 @@
 [versions]
+antlr-base = "4.10.1"
 antlr-shadowed = "4.10.1.6"
 asm-relocated = "9.7.1.0"
 caffeine = "3.1.8"
@@ -11,7 +12,11 @@ guava = "33.1.0-jre"
 java = "11"
 junit = "5.10.2"
 kotlin = "1.9.25"
-kotlineditor = "0.10" # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
+
+# Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
+kotlineditor = "0.10"
+kotlineditor-shaded = "0.10.0-SNAPSHOT"
+
 kotlinx-metadata = "0.9.0"
 moshi = "1.15.1"
 moshix = "0.25.1"
@@ -27,6 +32,9 @@ agp-common = "31.2.0"
 [libraries]
 agp = { module = "com.android.tools.build:gradle", version.ref = "agp" }
 android-tools-common = { module = "com.android.tools:common", version.ref = "agp-common" }
+
+antlr-core = { module = "org.antlr:antlr4", version.ref = "antlr-base" }
+antlr-runtime = { module = "org.antlr:antlr4-runtime", version.ref = "antlr-base" }
 
 caffeine = { module = "com.github.ben-manes.caffeine:caffeine", version.ref = "caffeine" }
 commons-io = { module = "commons-io:commons-io", version.ref = "commons-io" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,6 +11,7 @@ guava = "33.1.0-jre"
 java = "11"
 junit = "5.10.2"
 kotlin = "1.9.25"
+kotlineditor = "0.10" # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
 kotlinx-metadata = "0.9.0"
 moshi = "1.15.1"
 moshix = "0.25.1"
@@ -38,6 +39,8 @@ javax-inject = "javax.inject:javax.inject:1"
 
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
 kotlin-dokka = { module = "org.jetbrains.dokka:kotlin-as-java-plugin", version.ref = "kotlin" }
+kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlineditor" }
+kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlineditor" }
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
 kotlin-stdlib-core = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,7 +14,7 @@ junit = "5.10.2"
 kotlin = "1.9.25"
 
 # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
-kotlineditor = "0.10"
+kotlineditor-core = "0.10"
 kotlineditor-shaded = "0.10.0-SNAPSHOT"
 
 kotlinx-metadata = "0.9.0"
@@ -47,8 +47,8 @@ javax-inject = "javax.inject:javax.inject:1"
 
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
 kotlin-dokka = { module = "org.jetbrains.dokka:kotlin-as-java-plugin", version.ref = "kotlin" }
-kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlineditor" }
-kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlineditor" }
+kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlineditor-core" }
+kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlineditor-core" }
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
 kotlin-stdlib-core = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,7 +15,7 @@ kotlin = "1.9.25"
 
 # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
 kotlineditor-core = "0.10"
-kotlineditor-shaded = "0.10.0-SNAPSHOT"
+kotlineditor-relocated = "0.10.0-SNAPSHOT"
 
 kotlinx-metadata = "0.9.0"
 moshi = "1.15.1"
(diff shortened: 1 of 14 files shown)


gradle/libs.versions.toml
```diff
@@ -11,6 +11,7 @@ guava = "33.1.0-jre"
 java = "11"
 junit = "5.10.2"
 kotlin = "1.9.25"
+kotlineditor = "0.6" # Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
```
autonomousapps, line 14: Note I published two new versions yesterday, so this is now up to 0.8.

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,114 @@
...
+import java.nio.file.Path
+
+/**
+ * Rewrites a Kotlin build script with advice.
+ *
+ * Logic:
+ * - Ignores buildscript dependencies
+ * - When enter each declaration, removes or updated it if based on the advice. See `handleDeclaration` function
+ * - At the end of the script, add any new dependencies
+ */
```
autonomousapps, line 26: Love the docs!

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Returns build script with advice applied.
+   *
+   * Throws [KotlinParseException] if the script has some idiosyncrasy that impairs parsing.
+   *
+   */
+  @Throws(KotlinParseException::class)
+  override fun rewritten(): String {
+    errorListener.getErrorMessages().ifNotEmpty {
+      throw KotlinParseException.withErrors(it)
```
autonomousapps, line 54: Do we need a custom exception just for Kotlin? (thinking aloud)

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  override fun exitNamedBlock(ctx: KotlinParser.NamedBlockContext) {
+    dependencyExtractor.onExitBlock()
+
+    // TODO(swipawiwat): all new dependencies to the end of the block if not in buildscript block
+  }
+
+  override fun exitScript(ctx: KotlinParser.ScriptContext?) {
```
autonomousapps, line 74: minor nits: (1) we can add imports for all contexts so we need need to qualify them all with `KotlinParser.` and (2) they're all non-null, but written in Java, so we have to manually remove the `?`.

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,114 @@
...
+    return rewriter.text
+  }
+
+  override fun enterNamedBlock(ctx: KotlinParser.NamedBlockContext) {
+    dependencyExtractor.onEnterBlock()
+
+    if (ctx.isDependencies) {
+      handleDependencies(ctx)
+    }
+  }
```
autonomousapps, line 66: Does this handle checking if it's in a buildscript block?

autonomousapps (commented):
This is looking really good! Thanks! <3

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: 👍 I think it comes from auto complete. I will fix them :)

wsutina: Fixed!

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: Yup! Just implemented it today

src/test/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriterTest.kt
```diff
@@ -0,0 +1,490 @@
...
+          testImplementation("pan-galactic:gargle-blaster:2.0-SNAPSHOT") {
+              because("life's too short not to")
+          }
+        }
+        
+        println("hello, world!")
+      """.trimIndent().trimmedLines()
+    )
+  }
+
```
wsutina, line 487: **Todos**

- [ ] Add test case for complicated build script e.g.

```
dependencies {
        if (true) {
          testImplementation("heart:of-gold:1.+")
        }

        devImplementation(group = "io.netty", name = "netty-transport-native-unix-common", classifier = "osx-aarch_64")
      }
```

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: Good point. This can just be `BuildScriptParseException` like . I updated it

gradle/libs.versions.toml
wsutina: - [ ] Bump this after [this](https://github.com/cashapp/kotlin-editor/pull/17) is merged

src/test/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriterTest.kt
```diff
@@ -485,6 +485,45 @@ internal class KotlinBuildScriptDependenciesRewriterTest {
     )
   }
 
+  @Test fun `can parse complex dependencies`() {
```
wsutina, line 488: dependencies in `if` statement and non-dependency declarations can be parsed but will not be rewritten

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,176 @@
...
+      inBuildscriptBlock = true
+    }
+
+    if (ctx.isDependencies) {
+      handleDependencies(ctx)
+    }
+  }
+
+  override fun exitNamedBlock(ctx: NamedBlockContext) {
+    dependencyExtractor.onExitBlock()
```
autonomousapps, line 85: I think the (implied) contract of this function is that it should be at the bottom of the `exitNamedBlock` function.

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,176 @@
...
+    if (hasDependenciesBlock) return
+
+    advice.filterToOrderedSet { it.isAnyAdd() }.ifNotEmpty { addAdvice ->
+      rewriter.insertBefore(
+        ctx.stop,
+        addAdvice.joinToString(prefix = "\ndependencies {\n", postfix = "\n}\n", separator = "\n") { a ->
+          printer.toDeclaration(a)
+        }
+      )
+    }
```
autonomousapps, line 114: This block looks basically identical to the block above, with the exception that this one uses `filterToOrderedSet` and the above one is just `filterToSet`. I think We should extract this to a function and use the ordered set function.

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
```diff
@@ -0,0 +1,176 @@
...
+
+      findAdvice(declaration)?.let { a ->
+        if (a.isAnyRemove()) {
+          rewriter.delete(context.start, context.stop)
+          rewriter.deleteWhitespaceToLeft(context.start)
+          rewriter.deleteNewlineToRight(context.stop)
+        } else if (a.isAnyChange()) {
+          val configuration = context.primaryExpression()
+          rewriter.replace(configuration.start, configuration.stop, a.toConfiguration)
+        }
```
autonomousapps, line 135: I think this is being too clever. The `DependencyDeclaration` data class has all the information we need, and we shouldn't try to re-parse the context object to get the configuration context out of it. We should just rewrite the whole line using `printer.toDeclaration(a)`. That ensures there's just one method responsible for printing declarations.

I think the above ☝️ is true, but I will note one caveat that came to mind as I wrote it. There are cases where there are really complex dependency declarations, and what I said assumes that DAGP is correctly modeling a declaration in its totality. Your approach is more conservative by _just_ rewriting the configuration name. 

All that said, I prefer keeping the code simple and using `printer.toDeclaration(a)` to rewrite the whole line. `fixDependencies` is an optional task and is allowed to make assumptions that `buildHealth` can't.

autonomousapps (changes requested):
Looks great! Thank you <3 Left a few comments.

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: > . We should just rewrite the whole line using printer.toDeclaration(a). That ensures there's just one method responsible for printing declarations.

Good point. I will update this

wsutina: 👍 I'm happy to update this to `printer.toDeclaration(a)` for simplicity. I was just following the existing pattern e.g. https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/d36ba8a9cf5cf6d03ba51863f32512ff38002ced/src/main/kotlin/com/autonomousapps/internal/parse/GroovyBuildScriptDependenciesRewriter.kt#L149-L150

### Note on the test set up
After making the [change](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/7c7c9154d8885437092bd881e34637860ae1fc87/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt), the test "update dependencies with dependencyMap" test failed e.g. `compileOnly(project(:marvin))` (without `""`) instead of `compileOnly(project(":marvin"))`

It's because the `AdvicePrinter` doesn't print the quote if `dependencyMap` is presented for the specific advice.
https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/7c7c9154d8885437092bd881e34637860ae1fc87/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt#L58-L66

To get the test passing, I need to avoid providing `dependencyMap` for the dependency we want to remain the same. See [change](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1299/commits/c358cd6521cc6afbfa677514d6b275be42d16e10#diff-85ff32ed71048450273d2fd46bcc842a7ef269ee87773608f362a8a55cc1bb96L152)  

### Question
Can we assume that the given `dependencyMap` from users should always include quotes?

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: Done in https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1299/commits/d36ba8a9cf5cf6d03ba51863f32512ff38002ced#diff-b2a94f740ad8ff6516bf944965d212e085314087d0edc9cc852710ad1ef35c21R94-R95

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
wsutina: Done in https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1299/commits

gradle/libs.versions.toml
wsutina: Done 52425f7dba261e6b47928ef0bf5b974e50bcbfd8

src/main/kotlin/com/autonomousapps/internal/parse/KotlinBuildScriptDependenciesRewriter.kt
autonomousapps: > Can we assume that the given dependencyMap from users should always include quotes?

Yes. Good catch and fix!

autonomousapps:
The `:buildHealth` failure is interesting, and I'll have to think about how to resolve it. The existing antlr support for Groovy deliberately shades antlr to work around common problems with people bringing in different versions of that tool and breaking this plugin. However, now we have KotlinEditor _also_ bringing in antlr. Maybe the most straightforward answer is to shade KotlinEditor as well. Could maybe do it with the extant `shadowed/antlr` project. 🤔

PR #1309, opened 2024-11-14
refactor: make a lot of data models internal that don't need to be public.

autonomousapps:


src/main/kotlin/com/autonomousapps/model/ProjectAdvice.kt
```diff
@@ -16,8 +16,9 @@ data class ProjectAdvice(
   val shouldFail: Boolean = false
 ) : Comparable<ProjectAdvice> {
 
-  fun isEmpty(): Boolean = dependencyAdvice.isEmpty() && pluginAdvice.isEmpty()
+  fun isEmpty(): Boolean = dependencyAdvice.isEmpty() && pluginAdvice.isEmpty() && warning.isEmpty()
   fun isNotEmpty(): Boolean = !isEmpty()
 
+  // TODO(tsr): this compareTo function violates the Comparable contract by only considering one field.
```
autonomousapps, line 22: TODO

PR #1315, opened 2024-11-19
test: delete test fixtures after test run.

autonomousapps:


autonomousapps:
Lots of memory used: https://scans.gradle.com/s/4ugjmzdpscj3c/performance/build#memory-g-1-old-gen

PR #1326, opened 2024-11-27
feat: improve output of ProjectGraphTask and include topological sort.

autonomousapps:
In a follow-up, I will add a task at the root level for doing this globally in one pass.


src/main/kotlin/com/autonomousapps/internal/graph/GraphWriter.kt
```diff
@@ -38,4 +44,29 @@ internal object GraphWriter {
...
+   */
+  fun topological(graph: Graph<Coordinates>): String {
+    return Topological(graph, graph.root())
+      .order
+      .joinToString(separator = "\n") {
+        // ":foo:bar 0"
+        // ":other:baz 1"
+        it.maybeProjectCoordinates(buildPath).gav() + " ${graph.inDegree(it)}"
+      }
+  }
```
autonomousapps, line 63: @staktrace this is the main area of interest for us currently.

I can think of very expensive algorithms that check if there's a path from one node to another, and using that to create the batches, but I feel like there might also be some common(ish) graph algo that can also cut up a graph into mutually semi-independent chunks. I'll page through the lit, but if you can think of anything I'm all ears.

autonomousapps: Omg, I just realized that what I want is basically just the dominance tree. It already contains all the information necessary to go in order in a maximally concurrent way. At least, I think. Will test later!

staktrace: Yeah I think using the dominator tree of the dependency graph makes sense here.

graph-support/src/main/kotlin/com/autonomousapps/graph/Graphs.kt
autonomousapps, line 1: Need to publish a new version of graph-support before I can merge.

src/functionalTest/groovy/com/autonomousapps/jvm/WorkPlanSpec.groovy
```diff
@@ -0,0 +1,26 @@
...
+
+import com.autonomousapps.jvm.projects.WorkPlanProject
+import spock.lang.PendingFeature
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class WorkPlanSpec extends AbstractJvmSpec {
+
+  @PendingFeature
```
autonomousapps, line 11: Fix test or delete feature and test.

src/main/kotlin/com/autonomousapps/internal/graph/GraphWriter.kt
```diff
@@ -38,4 +45,56 @@ internal object GraphWriter {
...
+      top.forEach { node ->
+        appendLine(node.maybeProjectCoordinates(buildPath).gav())
+        // append(" ")
+        // appendLine("${paths.distanceTo(node)}")
+      }
+    }
+  }
+
+  // TODO replace with Hu's algorithm (for scheduling concurrent work), and push relevant bits into Graphs.kt.
+  fun workPlan(graph: Graph<Coordinates>): String {
```
autonomousapps, line 69: Follow-up.

PR #1333, opened 2024-12-06
Allow adding custom content to android blocks

evant:
This is so you can test android dsl extensions.
https://developer.android.com/reference/tools/gradle-api/8.7/com/android/build/api/variant/AndroidComponentsExtension#registerExtension(com.android.build.api.variant.DslExtension,kotlin.Function1)


autonomousapps (approved):
LGTM!

PR #1339, opened 2024-12-14
chore: adjust shadow plugin coordinates

patrick-dedication:
The plugin was relocated see [here](https://github.com/GradleUp/shadow/releases/tag/8.3.0).
Version 8.3.0 is very similar to 8.1.1, the gradle requirement was bumped to 8.3 which shouldn't be an issue since you are using 8.11 as far as I can tell.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -28,7 +28,7 @@ pluginManagement {
     }
   }
   plugins {
-    id("com.github.johnrengelman.shadow") version "8.1.1"
+    id("com.gradleup.shadow") version "8.3.0"
     id("com.gradle.develocity") version "3.18.2"
     id("com.gradle.plugin-publish") version "1.1.0"
   }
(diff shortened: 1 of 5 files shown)


settings.gradle.kts
```diff
@@ -28,15 +28,15 @@ pluginManagement {
...
+    id("com.gradleup.shadow") version "8.3.0"
     id("com.gradle.develocity") version "3.18.2"
     id("com.gradle.plugin-publish") version "1.1.0"
   }
 }
 
 plugins {
   id("com.gradle.develocity")
-  id("org.gradle.toolchains.foojay-resolver-convention") version("0.8.0")
+  id("org.gradle.toolchains.foojay-resolver-convention") version ("0.8.0")
```
autonomousapps, line 39: nit: this either shouldn't have a space or should remove the `()`

autonomousapps (approved):
Thanks!

settings.gradle.kts
patrick-dedication: Ah, I like it, you pay attention to the details :)
Autoformatting changed it, so I thought I'd add it to the commit.
Reverted the change since it had nothing to do with this PR.
I'll create a new one to remove the (), so it is more consistent with the rest.

Issue #1330, opened 2024-12-03
Configuration for duplicate class detection


eygraber:
The output for `buildHealth` is spammed with warnings for duplicate classes. This is output for each module:

```
  Warnings
  Some of your classpaths have duplicate classes, which means the compile and runtime behavior can be sensitive to the classpath order.
  
  Source set: devDebug
  \--- compile classpath
       \--- org/jetbrains/annotations/NotNull is provided by multiple dependencies: [com.android.tools.layoutlib:layoutlib:14.0.11, org.jetbrains:annotations:23.0.0]
  \--- runtime classpath
       \--- org/jetbrains/annotations/NotNull is provided by multiple dependencies: [com.android.tools.layoutlib:layoutlib:14.0.11, org.jetbrains:annotations:23.0.0]
  
  Source set: devRelease
  \--- compile classpath
       \--- org/jetbrains/annotations/NotNull is provided by multiple dependencies: [com.android.tools.layoutlib:layoutlib:14.0.11, org.jetbrains:annotations:23.0.0]
  \--- runtime classpath
       \--- org/jetbrains/annotations/NotNull is provided by multiple dependencies: [com.android.tools.layoutlib:layoutlib:14.0.11, org.jetbrains:annotations:23.0.0]

```

which makes it really hard to find other errors. In the example above I filed issues with the relevant parties, and now I don't want to see the warnings while I wait for it to get resolved.

autonomousapps:
Thanks for the issue! I am aware this is a gap in the issues handler configuration. This gap will be closed soon.

PR #1340, opened 2024-12-14
feat: can set severity and filter duplicate class warnings.

autonomousapps:
TODO
- [x] Add tests


src/main/kotlin/com/autonomousapps/tasks/FilterAdviceTask.kt
```diff
@@ -162,7 +167,11 @@ abstract class FilterAdviceTask @Inject constructor(
         .filterNot {
           moduleStructureBehavior is Ignore || it.shouldIgnore(moduleStructureBehavior)
         }
-        .toSet()
+        .toSortedSet()
```
autonomousapps, line 170: This should have always been sorted.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1330.

PR #1341, opened 2024-12-15
chore: align gradle toolchain resolver syntax with project convention

patrick-dedication:
As discussed [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1339#discussion_r1885497306)


autonomousapps (approved):
Thanks!

PR #1343, opened 2024-12-15
chore: simplify KotlinCompile configuration

patrick-dedication:
Reason: the kotlin-gradle-plugin respects JvmToolchains, see [here](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support).

In case you wanted to be explicit about it, feel free to just close this PR :)

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -1,6 +1,5 @@
 // Copyright (c) 2024. Tony Robalik.
 // SPDX-License-Identifier: Apache-2.0
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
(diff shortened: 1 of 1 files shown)


autonomousapps (commented):
It didn't used to always respect it!

patrick-dedication:
So merge or close?

patrick-dedication:
Do you an automated test?
Or once with `javap -verbose -classpath build-logic/convention/build/libs/convention.jar com.autonomousapps.convention.ConventionPlugin | grep -b20 -a20 "major version"`?

A thought that came to me a few days ago:
This configuration configures the bytecode version for the plugins.
As far as I understand it the convention plugins runs in the gradle jvm.
I think using toolchains it is possible to compile them to a different version than gradle is run with.
So is it even wise to use toolchains at all in this case?

PR #1344, opened 2024-12-16
chore: fix typo in readme

patrick-dedication:


autonomousapps (changes requested):
Thanks! Would you mind simplifying this PR so it only addresses the typo, instead of also adjusting line lengths in unrelated sections?

patrick-dedication:
No I would not mind, I do mind though that IntelliJ had the impression it has to autoformat the file.
I'm not yet used to looking at the diff in GitHub when I created a PR, will be there soon, took me a while to do that in GitLab too.

autonomousapps (approved):

Issue #1346, opened 2024-12-24
DAGP ignores output of Android artifacts transform tasks


dkostyrev:
**Build scan link**
[Build scan](https://scans.gradle.com/s/ib3z5ovffa2is)

**Plugin version**
`2.6.1`

**Gradle version**
`8.10.2`

**JDK version**
`17`

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
`2.1.0`

**(Optional) Android Gradle Plugin (AGP) version**
`8.7.3`

**Describe the bug**
The plugin ignores output of tasks that transform Android variant artifacts. This behaviour leads to an empty ABI report generated (`abi-dump.txt`) and incorrect `implementation` / `api` advices (dependency should be `implementation`).

**To Reproduce**
[Minimal reproducer](https://github.com/dkostyrev/dagp-reproducer)
1.  `./gradlew :module:abiAnalysisRelease`
2. `cat module/build/reports/dependency-analysis/releaseMain/abi-dump.txt`
3. The report is empty

**Expected behavior**
The plugin should take into account output of tasks that transform Android variants artifacts. 

**Additional context**
The plugin makes wrong assumption [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/tasks/AndroidClassesTask.kt#L23) and not taking into account JAR file [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/main/kotlin/com/autonomousapps/tasks/AndroidClassesTask.kt#L35) (which could be not empty with such transform task).

autonomousapps:
Thanks for the issue and the PR! I'll take a look.

PR #1347, opened 2024-12-26
fix: fixes issue 1346

dkostyrev:
Fixes the issue #1346.

`ClassReferenceParser` and `getBinaryAPI` extension functions now handle `jar` files as an input alongside with class files. There is a bit of duplication with regard to to reading jar files, maybe it would be better to hide two sources of class files inside `AndroidClassesTask` by unpacking jars into intermediate directory. I've also added a test that fails without aforementioned fixes, it is a bit verbose because of the actual implementation of the transform task.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
@@ -64,6 +64,10 @@ internal fun Iterable<File>.filterToClassFiles(): List<File> {
   return filter { it.extension == "class" && !it.name.endsWith("module-info.class") }
 }
 
+internal fun Iterable<File>.filterToJarFiles(): List<File> {
+  return filter { it.extension == "jar" }
+}
+
 /** Filters a [FileCollection] to contain only class files. */
 internal fun FileCollection.filterToClassFiles(): FileCollection {
   return filter {
--- a/src/main/kotlin/com/autonomousapps/tasks/AndroidClassesTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/AndroidClassesTask.kt
@@ -3,6 +3,7 @@
 package com.autonomousapps.tasks
 
 import com.autonomousapps.internal.utils.filterToClassFiles
+import com.autonomousapps.internal.utils.filterToJarFiles
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.Directory
 import org.gradle.api.file.RegularFile
@@ -20,7 +21,7 @@ import java.io.File
  */
 abstract class AndroidClassesTask : DefaultTask() {
 
-  /** Will be empty for this task. */
+  /** May be empty. */
   @get:PathSensitive(PathSensitivity.RELATIVE)
   @get:InputFiles
   abstract val jars: ListProperty<RegularFile>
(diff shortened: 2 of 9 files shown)


src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
```diff
@@ -64,6 +64,10 @@ internal fun Iterable<File>.filterToClassFiles(): List<File> {
   return filter { it.extension == "class" && !it.name.endsWith("module-info.class") }
 }
 
+internal fun Iterable<File>.filterToJarFiles(): List<File> {
+  return filter { it.extension == "jar" && !it.name.endsWith("R.jar") }
```
autonomousapps, line 68: why are we excluding `R.jar`?

src/main/kotlin/com/autonomousapps/tasks/AndroidClassesTask.kt
```diff
@@ -20,7 +21,7 @@ import java.io.File
  */
 abstract class AndroidClassesTask : DefaultTask() {
 
-  /** Will be empty for this task. */
+  /** May be empty. */
```
autonomousapps, line 24: Thanks for updating the comment!

autonomousapps (commented):
This looks very reasonable. Thanks for the contribution! Left you a question.

Please note I think you'll have to rebase and fix some (hopefully minor) conflicts after I merge https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1349.

src/main/kotlin/com/autonomousapps/internal/utils/collections.kt
dkostyrev: Initially I've excluded `R.jar` because of the transitive dependencies false positives. For example, if the project uses `androidx.appcompat:appcompat:1.1.0` as a dependency, the plugin would report following transitive dependencies, although none of them were really used:

```
  implementation 'androidx.activity:activity:1.0.0'
  implementation 'androidx.appcompat:appcompat-resources:1.1.0'
  implementation 'androidx.core:core:1.1.0'
  implementation 'androidx.customview:customview:1.0.0'
  implementation 'androidx.drawerlayout:drawerlayout:1.0.0'
  implementation 'androidx.fragment:fragment:1.1.0'
  implementation 'androidx.loader:loader:1.0.0'
  implementation 'androidx.vectordrawable:vectordrawable-animated:1.1.0'
  implementation 'androidx.vectordrawable:vectordrawable:1.1.0'
  implementation 'androidx.viewpager:viewpager:1.0.0'
```

According to `usage-dependencies.json` this advice was given because of `androidx.**.R` classes in the `R.jar`:

```
Uses 11 classes, 5 of which are shown: androidx.core.R, androidx.core.R$attr, androidx.core.R$color, androidx.core.R$dimen, androidx.core.R$drawable
```

I've dug a bit deeper and find out that the plugin will add inner's class parent name to the `ExplodedClass.nonAnnotationClasses` and vice versa (inner class names will be added to the parent class `nonAnnotationClasses`), I've added a fix similar to the exclusion of the class own name in `fixup` functions.

dkostyrev:
@autonomousapps Hi, could you take a look at this PR please?

autonomousapps:
I apologize. I don't remember why this fell off my radar.

PR #1349, opened 2025-01-09
feat: don't recommend removing implementation dependency on necessary superclass.

autonomousapps:


graph-support/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
   id("com.autonomousapps.dependency-analysis")
 }
 
-version = "0.4"
+version = "0.5-SNAPSHOT"
```
autonomousapps, line 9: publish

graph-support/src/main/kotlin/com/autonomousapps/graph/Graphs.kt
```diff
@@ -5,6 +5,7 @@ package com.autonomousapps.graph
 import com.google.common.graph.Graph
 import com.google.common.graph.Graphs as GuavaGraphs
 
+// TODO: publish new version of graph-support
```
autonomousapps, line 8: publish

src/main/kotlin/com/autonomousapps/model/internal/Capability.kt
```diff
@@ -102,140 +104,132 @@ internal data class AnnotationProcessorCapability(
...
+@TypeLabel("binaryClass")
 @JsonClass(generateAdapter = false)
-internal data class ClassCapability(
-  val classes: Set<String>,
+internal data class BinaryClassCapability(
+  val binaryClasses: Set<BinaryClass>,
 ) : Capability() {
 
+  // TODO need to dotty everything, this is driving me nuts
+  val classes by unsafeLazy { binaryClasses.mapToOrderedSet { it.className.replace('/', '.') } }
```
autonomousapps, line 114: todo

PR #1350, opened 2025-01-10
Update to AGP 8.8.0

MGaetan89:
This updates the project to:
- AGP 8.8.0 (latest feature release).
- AGP 8.7.3 (latest bugfix release).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -19,7 +19,6 @@ kotlinMetadata = "2.0.21"
 kotlineditor-core = "0.18"
 kotlineditor-relocated = "0.18.0"
 
-kotlinx-metadata = "0.9.0"
 moshi = "1.15.1"
 moshix = "0.25.1"
 okhttp = "4.12.0"
@@ -43,7 +42,7 @@ commons-io = { module = "commons-io:commons-io", version.ref = "commons-io" }
 errorProne = { module = "com.google.errorprone:error_prone_annotations", version.ref = "error-prone" }
 gradle-publish-plugin = { module = "com.gradle.plugin-publish:com.gradle.plugin-publish.gradle.plugin", version.ref = "gradle-publish-plugin" }
 grammar = { module = "com.autonomousapps:gradle-script-grammar", version.ref = "grammar" }
-guava = { module ="com.google.guava:guava", version.ref = "guava" }
+guava = { module = "com.google.guava:guava", version.ref = "guava" }
 
 javax-inject = "javax.inject:javax.inject:1"
 
--- a/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
@@ -21,23 +21,22 @@ abstract class AbstractAndroidSpec extends AbstractFunctionalSpec {
   protected static final AGP_8_4 = AgpVersion.version('8.4.2')
   protected static final AGP_8_5 = AgpVersion.version('8.5.2')
   protected static final AGP_8_6 = AgpVersion.version('8.6.1')
-  protected static final AGP_8_7 = AgpVersion.version('8.7.0')
-  protected static final AGP_8_8 = AgpVersion.version('8.8.0-alpha04')
+  protected static final AGP_8_7 = AgpVersion.version('8.7.3')
+  protected static final AGP_8_8 = AgpVersion.version('8.8.0')
 
   protected static final AGP_LATEST = AGP_8_8
 
   /**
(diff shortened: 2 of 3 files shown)


MGaetan89:
It looks like there are failing tests with AGP 8.9.0-alpha08 and data binding.
I'll look into it.

MGaetan89:
It looks like data binding uses Kotlin 2.1.0 in AGP 8.9.0:
```
e: file:///home/runner/work/dependency-analysis-gradle-plugin/dependency-analysis-gradle-plugin/build/tmp/functionalTest/work/.gradle-test-kit/caches/8.11.1/transforms/52eff739c692a29ed08427835199b376/transformed/databinding-ktx-8.9.0-alpha08-api.jar!/META-INF/databindingKtx_release.kotlin_moduleModule was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.1.0, expected version is 1.9.0. 
e: file:///home/runner/work/dependency-analysis-gradle-plugin/dependency-analysis-gradle-plugin/build/tmp/functionalTest/work/.gradle-test-kit/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.1.0/85f8b81009cda5890e54ba67d64b5e599c645020/kotlin-stdlib-2.1.0.jar!/META-INF/kotlin-stdlib-jdk7.kotlin_moduleModule was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.1.0, expected version is 1.9.0.
e: file:///home/runner/work/dependency-analysis-gradle-plugin/dependency-analysis-gradle-plugin/build/tmp/functionalTest/work/.gradle-test-kit/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.1.0/85f8b81009cda5890e54ba67d64b5e599c645020/kotlin-stdlib-2.1.0.jar!/META-INF/kotlin-stdlib-jdk8.kotlin_moduleModule was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.1.0, expected version is 1.9.0.
e: file:///home/runner/work/dependency-analysis-gradle-plugin/dependency-analysis-gradle-plugin/build/tmp/functionalTest/work/.gradle-test-kit/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.1.0/85f8b81009cda5890e54ba67d64b5e599c645020/kotlin-stdlib-2.1.0.jar!/META-INF/kotlin-stdlib.kotlin_moduleModule was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.1.0, expected version is 1.9.0.
```

MGaetan89:
@autonomousapps I've updated Kotlin to 2.1.0 to see if that works.
Let me know if AGP 8.9/Kotlin 2.1.0 should be added in a separate PR, so AGP 8.8.0 can be integrated first.

autonomousapps:
Could we instead bump to Kotlin 2.0.x? And also continue to target the Kotlin 1.9 api? I have important users (i.e. at my job) that still use Kotlin 1.9 and I will not break them.

I'm open to discussing strategies regarding how to support latest AGP while also maintaining support for projects using an older version of Kotlin.

MGaetan89:
> Could we instead bump to Kotlin 2.0.x? And also continue to target the Kotlin 1.9 api? I have important users (i.e. at my job) that still use Kotlin 1.9 and I will not break them.

I'll give it a try.

> I'm open to discussing strategies regarding how to support latest AGP while also maintaining support for projects using an older version of Kotlin.

The latest stable AGP (8.8.0) still works with Kotlin 1.9.x.
The update seems to only be necessary for AGP 8.9 alpha. So I could submit a PR to update to AGP 8.8.0 stable, and a second for AGP 8.9 alpha and Kotlin 2.x.
Since AGP 8.9 is only in alpha, Kotlin 2.x might not be necessary in the final version, or we might think of a solution until then.

autonomousapps:
> The update seems to only be necessary for AGP 8.9 alpha. So I could submit a PR to update to AGP 8.8.0 stable, and a second for AGP 8.9 alpha and Kotlin 2.x.

It sounds reasonable to split into two PRs so we can land the stable changes and then consider the alpha changes separately. Thanks!

MGaetan89:
All good for the 8.8 branch.
I'll open an other PR for the 8.9 one soon.

gradle/libs.versions.toml
```diff
@@ -19,7 +19,6 @@ kotlinMetadata = "2.0.21"
 kotlineditor-core = "0.18"
 kotlineditor-relocated = "0.18.0"
 
-kotlinx-metadata = "0.9.0"
```
autonomousapps, line 22: thanks for catching this 👍

autonomousapps (approved):

PR #1355, opened 2025-01-11
Add toolchain plugin everywhere and unify plugin style

TWiStErRob:
> Configure project :testkit:gradle-testkit-plugin
Using a toolchain installed via auto-provisioning, but having no toolchain repositories configured. This behavior is deprecated. Consider defining toolchain download repositories, otherwise the build might fail in clean environments; see https://docs.gradle.org/8.11-rc-3/userguide/toolchains.html#sub:download_repositoriesed artifacts.

It's needed in all projects where there's a settings.gradle AND toolchains are used, which happens to be all 3.

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -7,6 +7,13 @@ pluginManagement {
     gradlePluginPortal()
     mavenCentral()
   }
+  plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+  }
+}
+
+plugins {
+  id("org.gradle.toolchains.foojay-resolver-convention")
 }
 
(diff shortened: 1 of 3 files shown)


autonomousapps (commented):
Latest version is 1.0.0.

I'm going through some old PRs (sorry, I was very busy for a while). Unfortunately I don't think I can merge this right now given there's currently an [outage related to foojay](https://github.com/foojayio/discoapi/issues/124). I'll come back to this later.

autonomousapps:
> Sounds fun, want me to resolve conflicts?
> 
> Note: because it is working at the moment I have a feeling Gradle might be applying the plugin anyway, but haven't dug too deep into it ever.

Yes, if you resolve conflicts I'd be happy to take another look!

PR #1362, opened 2025-01-28
fix: compileOnly dependencies are not visible to the test compile classpath.

autonomousapps:


src/main/kotlin/com/autonomousapps/model/declaration/internal/Bucket.kt
```diff
@@ -37,9 +43,30 @@ internal enum class Bucket(val value: String) {
...
+     * Note that the `compileOnly` configuration _is not_ visible to the `testImplementation` configuration.
+     *
+     * @see <a href="https://docs.gradle.org/current/userguide/java_plugin.html#resolvable_configurations.">Java configurations</a>
+     */
+    fun isVisibleToTestSource(usages: Set<Usage>, declarations: Set<Declaration>): Boolean {
+      return usages.reallyAll { usage ->
+        VISIBLE_TO_TEST_SOURCE.any { it == usage.bucket }
+          && declarations.any { API.matches(it) || IMPL.matches(it) }
+      }
+    }
```
autonomousapps, line 70: The fix.

autonomousapps, line 70: The fix.

src/main/kotlin/com/autonomousapps/transform/StandardTransform.kt
```diff
@@ -51,17 +51,15 @@ internal class StandardTransform(
...
 
     /*
      * Main usages.
      */
 
     val singleVariant = mainUsages.size == 1
-    val isMainVisibleDownstream = mainUsages.reallyAll { usage ->
-      Bucket.VISIBLE_TO_TEST_SOURCE.any { it == usage.bucket }
-    }
+    val isMainVisibleDownstream = Bucket.isVisibleToTestSource(mainUsages, mainDeclarations)
```
autonomousapps, line 62: The fix.

PR #1363, opened 2025-01-29
fix: do not filter out advice to add to testImplementation if there is conflicting advice to downgrade from implementation.

autonomousapps:


src/main/kotlin/com/autonomousapps/tasks/ReasonTask.kt
```diff
@@ -356,7 +356,15 @@ abstract class ReasonTask @Inject constructor(
...
-    private fun wasFiltered(): Boolean = finalAdvice == null && unfilteredAdvice != null
+    // TODO cleanup after tests pass (fingers crossed)
+    // TODO fix: these empty checks are not sufficient. Need to check _identity_ if the advice
+    private fun wasFiltered(): Boolean {
+      return unfilteredAdvice.any { unfiltered ->
+        unfiltered !in finalAdvice
+      }
+
+      // return finalAdvice.isEmpty() && unfilteredAdvice.isNotEmpty()
+    }
```
autonomousapps, line 367: TODO: cleanup before merging.

autonomousapps:
Spurious test failure. Passes locally.

PR #1367, opened 2025-01-29
build: fix deprecation warning by mangling attributes instead.

autonomousapps:


jjohannes:
My initial instinct would have been to change the `USAGE` attribute to something unknown. Because, in my understanding, that would render the variant completely unusable. But the current change looks more "correct" in a way. So if this works, I think it is good.

autonomousapps:
> My initial instinct would have been to change the `USAGE` attribute to something unknown. Because, in my understanding, that would render the variant completely unusable. But the current change looks more "correct" in a way. So if this works, I think it is good.

I feel like you're probably right. `Bundling` has some compatibility rules defined and I don't feel like looking them up -- it feels like kind of like a "secondary attribute" compared to `Usage`.

autonomousapps:
Ah, updating the usage got me [this error](https://scans.gradle.com/s/bjqjip7bubri6/failure#1):
```
Could not resolve all dependencies for configuration ':runtimeClasspath'.
> Could not resolve com.autonomousapps:kotlin-editor-relocated:0.18.0.
  Required by:
      root project :
  > Consumable configurations with identical capabilities within a project (other than the default configuration) must have unique attributes, but configuration ':kotlin-editor-relocated:apiElements' and [configuration ':kotlin-editor-relocated:runtimeElements'] contain identical attribute sets. Consider adding an additional attribute to one of the configurations to disambiguate them.  Run the 'outgoingVariants' task for more details. For more information, please refer to https://docs.gradle.org/8.12.1/userguide/upgrading_version_7.html#unique_attribute_sets in the Gradle documentation.
```
@tresat I think this would be a useful first-class feature for Gradle, to support "hiding" the non-shadowed variants.

tresat:
> Consumable configurations with identical capabilities within a project (other than the default configuration) must have unique attributes, but configuration ':kotlin-editor-relocated:apiElements' and [configuration ':kotlin-editor-relocated:runtimeElements'] contain identical attribute sets. Consider adding an additional attribute to one of the configurations to disambiguate them.

I forgot about this requirement.  But, this error message is pretty good, it tells you exactly where the problem is and suggests how to solve it!  (I think I added this one. 😄)

I think this would be a useful feature as well.  I'll make us an issue for it.

autonomousapps:
> But, this error message is pretty good

💯

Issue #1358, opened 2025-01-13
2.7.0 dramatically increases heap and CPU usage, probably due to `isForMissingSuperclass` check


eygraber:
I run `buildHealth` as part of my PR checks on GitHub Actions. The task usually takes around 10 minutes, but occasionally it will run for 20+ minutes and then the build suddenly fails. My hypothesis is that the OOM killer stops the process.

My Gradle and Kotlin jvm args are:

```
org.gradle.jvmargs=-Xmx8g -Xms8g -XX:ReservedCodeCacheSize=256m -XX:+UseParallelGC -XX:SoftRefLRUPolicyMSPerMB=2 -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError
kotlin.daemon.jvm.options=-Xmx8g -Xms8g -XX:ReservedCodeCacheSize=320m -XX:+UseParallelGC -XX:SoftRefLRUPolicyMSPerMB=2 -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError
```

I've found that I need to lower the `Xmx` value for Gradle in order for the task to not get killed. `buildHealth` is the only task running on that runner, so I'm guessing there's a memory leak or something that is causing memory usage to spike. Haven't had the chance to investigate locally yet.

This is happening on 2.7.0 but it has also happened on previous versions.

StefMa:
I can confirm that 2.7.0 explodes on GitHub Actions for us as well.
We have a big project and using a large runner with 8 cores and 32 Gigs of RAM.
Our average CI time is 15 minutes (including tests, linting and stuff).
However, with 2.7.0 we went up to 40 minutes on average.
Sometimes even to 1h15m, we have had two runs with 2hours...

eygraber:
I ran the gradle profiler benchmark on my project using 2.4.2 (which is the last version I remember running smoothly) and 2.8.0.

2.4.2 - 10 iterations, mean time 29s
2.8.0 - 10 iterations, mean time 47s

I'm going to rerun the profile now with scans to get more information.

eygraber:
The zip file contains the output of profiling the build using 2.4.2 and 2.8.0 with JFR and build scans.

2.4.2 build scan - https://scans.gradle.com/s/wkx3n6v4wf44u/
2.8.0 build scan - https://scans.gradle.com/s/pmcop52xrhixq

[...]

PR #1369, opened 2025-02-07
fix: use less heap.

autonomousapps:


autonomousapps:
> Would this help with https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1358
> 
> It seems like it might actually be more of a CPU issue (as opposed to memory like the title says), although the memory usage has increased dramatically as well.

I'll know soon if this helps with my work project and if so may publish a 2.8.1.

PR #1379, opened 2025-02-20
perf: improve performance of isForMissingSuperclass (again).

autonomousapps:


eygraber:
Nice this got me ~10% faster builds on my projects, thanks!

PR #1389, opened 2025-03-04
feature: new task for producing a version catalog file containing all dependencies

mmollaverdi:


src/functionalTest/groovy/com/autonomousapps/android/AllDependenciesSpec.groovy
```diff
@@ -0,0 +1,30 @@
...
+import static com.autonomousapps.kit.truth.BuildResultSubject.buildResults
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertAbout
+import static com.google.common.truth.Truth.assertThat
+
+final class AllDependenciesSpec extends AbstractAndroidSpec {
+
+  def "can generate a version catalog file with all dependencies (#gradleVersion AGP #agpVersion)"() {
+    given:
+    def project = new DuplicateDependencyVersionsProject(agpVersion as String)
```
mmollaverdi, line 17: Using this existing project fixture `DuplicateDependencyVersionsProject` because it had what I needed, happy to rename it to be more generic or create a new one for this test.

src/main/kotlin/com/autonomousapps/tasks/ComputeAllDependenciesTask.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
+
+abstract class ComputeAllDependenciesTask : DefaultTask() {
```
autonomousapps, line 18: I think this could be annotated with `@CacheableTask`.

src/main/kotlin/com/autonomousapps/tasks/ComputeAllDependenciesTask.kt
```diff
@@ -0,0 +1,71 @@
...
+
+  @TaskAction
+  fun action() {
+    val outputFile = output.getAndDelete()
+
+    val libs: List<String> = resolvedDependenciesReports
+      .dependencyCoordinates()
+      .map { "${it.toVersionCatalogAlias()} = { module = \"${it.identifier}\", version = \"${it.resolvedVersion}\" }" }
+      .toSet() // remove duplicates
+      .sorted()
```
autonomousapps, line 40: Could we use `toSortedSet()` and type this as a `Set<String>` instead?

src/main/kotlin/com/autonomousapps/tasks/ComputeAllDependenciesTask.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    val tomlContent = buildString {
+      appendLine("[libraries]")
+      libs.forEach { appendLine(it) }
+    }
+
+    outputFile.writeText(tomlContent)
+
+    logger.quiet("Generated version catalog for all project's dependencies at ${outputFile.absolutePath} " +
+      "containing ${libs.size} entries")
```
autonomousapps, line 50: I think this would be a bit more readable if the file path were on its own line. So maybe this:
```
logger.quiet("Generated version catalog for all dependencies, containing ${libs.size} entries:\n${outputFile.absolutePath} ")
```
Note that I also include a single space at the end of the output, because this helps with clickability in various terminals.

src/main/kotlin/com/autonomousapps/tasks/ComputeAllDependenciesTask.kt
```diff
@@ -0,0 +1,71 @@
...
+
+  private fun ModuleCoordinates.toVersionCatalogAlias(): String {
+    return "${this.identifier}-${this.resolvedVersion}"
+      .split(':', '.')
+      // replace reserved keywords with safe alternatives
+      .joinToString(separator = "-") { tomlReservedKeywordMappings.getOrDefault(it, it) }
+      .lowercase()
+  }
+
+  companion object {
```
autonomousapps, line 61: let's make this `private`.

autonomousapps (changes requested):
I think the general direction of this is great, but have left a couple of comments that mostly relate to visibility (I don't like leaking functions or classes onto buildscript classpaths).

src/functionalTest/groovy/com/autonomousapps/android/projects/DuplicateDependencyVersionsProject.groovy
```diff
@@ -70,6 +71,42 @@ final class DuplicateDependencyVersionsProject extends AbstractAndroidProject {
     return resolvedDependenciesReport(gradleProject, projectPath)
   }
 
+  String actualAllDependencies() {
+    return gradleProject.singleArtifact(':', OutputPathsKt.getAllLibsVersionsTomlPath()).asPath.text
+  }
+
+  String expectedAllDependencies = '''\
+    [libraries]
+    androidx-activity-activity-1-0-0 = { module = "androidx.activity:activity", version = "1.0.0" }
```
MGaetan89, line 80: It looks like the generated name is not valid:
Screenshot 2025-03-12 at 09 31 47

PR #1400, opened 2025-03-17
feature: computeResolvedDependencies to also generate a version catalo file

mmollaverdi:


src/main/kotlin/com/autonomousapps/model/Coordinates.kt
```diff
@@ -190,6 +190,25 @@ data class ModuleCoordinates(
...
 ) : Coordinates(identifier, gradleVariantIdentification) {
   override fun gav(): String = "$identifier:$resolvedVersion"
+
+  internal fun toVersionCatalogAlias(): String {
+    return "${identifier}-${resolvedVersion}"
+      .split(':', '.')
+      // replace reserved keywords with safe alternatives
+      .joinToString(separator = "-") { tomlReservedKeywordMappings.getOrDefault(it, it) }
+      .lowercase()
+  }
```
autonomousapps, line 200: Can you add some unit tests in `CoordinatesTest` for this functionality?

Hmm... do we need to include the `resolvedVersion` in the ailas for the case where there are duplicates (except by version)? Would be nice if we had the version suffix only when there were duplicates and not always?

autonomousapps (changes requested):
There's one [failed test](https://scans.gradle.com/s/x7ven2qy4g7oa/tests/task/:functionalTest/details/com.autonomousapps.android.AllDependenciesSpec/can%20generate%20a%20version%20catalog%20file%20with%20all%20dependencies%20(Gradle%208.0.2%20AGP%208.0.2)?top-execution=1). It looks like it's failing because you changed a line from `toSet()` to `toSortedSet()` and so now the actual output isn't matching the expected output. That was a good change though, so let's just fix the assertion.

autonomousapps (approved):

PR #1401, opened 2025-03-18
fix: issue 1360. Include check for java/main in ArtifactsReportTask

### LINKED ISSUE #1360: Not all classes included as part of project ABI when upgrading to Kotlin 2.0.21
**Build scan link**
Proprietary project :(

**Plugin version**
2.6.1

**Gradle version**
8.10.2

**JDK version**
17 (Corretto-17.0.10.7.1)

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
2.0.21

8.7.2

**(Optional) `reason` output for bugs relating to incorrect advice**

```
  implementation(project(":diagnostics:common"))    <------ incorrect

  api("androidx.datastore:datastore-core:1.0.0")
  api("androidx.datastore:datastore-preferences-core:1.0.0")
  api(globalLibs.javaxInject)
  api(globalLibs.kotlinCoroutine)
  implementation("androidx.datastore:datastore:1.0.0")
  implementation("com.bugsnag:bugsnag-android-core:6.7.0")
```

**Describe the bug**
When upgrading from Kotlin `2.0.0` to Kotlin `2.0.21` the dependency analysis plugin begins to incorrectly mark some of our modules as unused in other modules, even though they are explicitly used. After digging into it a bit, I discovered that for some module dependencies in `/reports/dependency-analysis/<variant>/<dependency>.json` the `files` property is different. 

* Using kotlin 2.0.0, files is `/Users/ward.bonnefond/Development/android/shared/diagnostics/common/build/classes`
* Using kotlin 2.0.21 files is`/Users/ward.bonnefond/Development/android/shared/diagnostics/common/build/classes/java/main`

Image

This no longer include the classes from `/build/classes/kotlin` as part of the module's ABI, so it gets marked as unused in most other modules that use it.

Image

I don't see this with every module though, so I don't know what has changed with Kotlin 2.0.21 here.

**To Reproduce**
1. ...

**Expected behavior**

**Additional context**
I will try to get a reproducer project together, but it might take me a bit of time.

wbonnefond:
This PR fixes #1360. 

After updating to Kotlin 2.0.21 in our large Android project, the artifact path is sometimes incorrect. This PR applies that same workaround to find the grandparent folder that was previously added for `kotlin/main`.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
@@ -74,7 +74,7 @@ abstract class ArtifactsReportTask : DefaultTask() {
       .mapNotNull {
         try {
           // https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/948#issuecomment-1711177139
-          val file = if (it.file.path.endsWith("kotlin/main")) {
+          val file = if (it.file.path.endsWith("kotlin/main") || it.file.path.endsWith("java/main")) {
             it.file.parentFile!!.parentFile!!
           } else {
             it.file


autonomousapps (approved):
This looks reasonable. Thanks!

PR #1404, opened 2025-03-25
fix: detect usages of SAM interfaces.

autonomousapps:
* fix: binaryClassAccesses uses sorted keys.
    * Previously only had sorted values. This should improve cacheability.
* fix: detect usages of SAM interfaces.
    * Instead of doing simply string matching, use regex to find all valid Java class descriptors.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/SamInterfacesProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/SamInterfacesProject.groovy
@@ -1,0 +1,104 @@
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Source
+import com.autonomousapps.model.ProjectAdvice
+
+import static com.autonomousapps.AdviceHelper.actualProjectAdvice
+import static com.autonomousapps.AdviceHelper.emptyProjectAdviceFor
+import static com.autonomousapps.kit.gradle.Dependency.project
+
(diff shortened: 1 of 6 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/SamInterfacesProject.groovy
```diff
@@ -0,0 +1,104 @@
...
+    Source.kotlin(
+      """\
+      package com.example.consumer
+            
+      import com.example.consumer.java.SamInterfaceConsumer
+            
+      class Consumer {
+        fun usesJavaProducer() {
+          val consumer = SamInterfaceConsumer()
+          consumer.setWhatever(com.example.producer.java.SamInterface { "magic!" })
```
autonomousapps, line 57: This was part of the issue that I saw in the wild. If this had been imported, we previously caught it with the import-checking backstop. But now we also detect it in the bytecode, which is better.

src/main/kotlin/com/autonomousapps/internal/asm.kt
```diff
@@ -587,14 +588,11 @@ private class FieldAnalyzer(
...
-    // Only add class types (not primitives)
-    if (it.startsWith("L")) {
-      add(ClassRef(it.substring(1, it.length - 1), kind))
-    }
-  }
+  classRef ?: return
+
+  JAVA_FQCN_REGEX_ASM.findAll(classRef)
+    .map { it.value }
+    .forEach { add(ClassRef(it.substring(1, it.length - 1), kind)) }
```
autonomousapps, line 595: Slightly worried this will have other impacts, but if the tests pass, should be good.

PR #1405, opened 2025-03-27
fix: use projectPath, not identityPath, for included build coordinates.

autonomousapps:
This has been a longstanding issue and will probably have multiple knock-on effects. The immediate reason I saw this was that SuperGraph was trying to find nodes built with one version of a graph using nodes from another version of a graph, and failing. This change should sync the two, which ought to also mean fewer outputs in the dependencies directory (no duplication for included build dependencies). I think lots of other code was working around this problem too without realizing it.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
@@ -43,7 +43,7 @@ private fun ResolvedDependencyResult.compositeRequest(): IncludedBuildCoordinate
     gradleVariantIdentification = gradleVariantIdentification
   )
   val resolved = ProjectCoordinates(
-    identifier = (selected.id as ProjectComponentIdentifier).identityPath(),
+    identifier = (selected.id as ProjectComponentIdentifier).projectPath(),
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -53,11 +53,16 @@ private fun ResolvedDependencyResult.compositeRequest(): IncludedBuildCoordinate
...
-  return (this as? DefaultProjectComponentIdentifier)?.identityPath?.toString()
+private fun ProjectComponentIdentifier.projectPath(): String {
+  return (this as? DefaultProjectComponentIdentifier)?.projectPath?.toString()
     ?: error("${toCoordinates(GradleVariantIdentification.EMPTY)} is not a DefaultProjectComponentIdentifier")
 }
 
+// private fun ProjectComponentIdentifier.identityPath(): String {
+//   return (this as? DefaultProjectComponentIdentifier)?.identityPath?.toString()
+//     ?: error("${toCoordinates(GradleVariantIdentification.EMPTY)} is not a DefaultProjectComponentIdentifier")
+// }
```
autonomousapps, line 64: - [ ] Cleanup.

PR #1408, opened 2025-04-02
feat: add temporary flag to test opting out of superclass analysis.

autonomousapps:
This is a hack that will be reverted or improved very quickly, and should not be relied upon.


eygraber:
A very unscientific, non-profiled comparison showed that `buildHealth` with the opt-out set to true ran ~10% faster and use 50% less RAM, if that helps.

PR #1424, opened 2025-04-19
Fix issues in testkit readme

ZacSweers:
This fixes a few issues in the README

- Invoke the base class constructor
- Add a list as the `sources` property is a List, not assignable to a single value
- Move the source used to _above_ the `gradleProject` init, as otherwise it isn't initialized yet and will be (a sneaky) null at init

### DIFF
--- a/testkit/README.asciidoc
+++ b/testkit/README.asciidoc
@@ -91,25 +91,11 @@ class Test {
   }
 }
 
-class MyFixture : AbstractGradleProject {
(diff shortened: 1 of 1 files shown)


autonomousapps:
Thank you! I just pushed a [commit](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/commit/3917938db67bbdadd599865b692f203098ec5f3a) with a change I like a bit better.

PR #1445, opened 2025-05-12
Fix ProjectHealthSpec on Windows environments

SimonMarquis:
Build Scan: https://scans.gradle.com/s/csbapb2cecxwo/tests/overview

Gstrings are surprising... The last line would not be correctly handled...

Besides this issue, multiple solutions are available for the file path:
- remove the manual string path concatenation with proper file resolving
- replace `ProjectHealthTask`'s `buildFilePath` input with `invariantSeparatorsPath`

I chose the first to avoid changing production code.


autonomousapps (approved):
Thanks! I'll trust you that this works on Windows. Maybe I should update my github actions to run against Windows as well as Linux.

PR #1449, opened 2025-05-13
feat: replace custom BuildScan comments with official solution

SimonMarquis:
https://github.com/gradle/actions/blob/main/docs/setup-gradle.md


autonomousapps (approved):
Thank you! I've been meaning to do this.

PR #1450, opened 2025-05-13
feat: detect and report all Android resource references in Manifest files

SimonMarquis:


src/test/kotlin/com/autonomousapps/internal/ManifestParserTest.kt
```diff
@@ -150,26 +152,60 @@ class ManifestParserTest {
...
+      AttrRef(type = "string", id = "deeplink_path"),
+      AttrRef(type = "string", id = "androidx_startup"),
+      AttrRef(type = "string", id = "google_maps_api_key"),
+      AttrRef(type = "integer", id = "google_play_services_version"),
+      AttrRef(type = "bool", id = "google_analytics_enabled"),
+    )
   }
 
-  private fun parse(manifest: String, dslNamespace: String = ""): ManifestParser.ParseResult {
+  private fun parse(@Language("XML") manifest: String, dslNamespace: String = ""): ManifestParser.ParseResult {
```
SimonMarquis, line 208: Adding this for syntax highlighting in the IDE

src/main/kotlin/com/autonomousapps/internal/ManifestParser.kt
```diff
@@ -18,8 +19,8 @@ internal class ManifestParser(
     val packageName: String,
     /** The value of the `android:name` attribute. May be empty. */
     val applicationName: String,
-    /** The values of the `android:theme` attributes of application and activities. May be empty. */
-    val themes: Set<String>,
```
SimonMarquis, line 22: I merge the `themes` inside a generic `attrRefs`.
Also, I parsed it directly as `AndroidResSource.AttrRef` to make tests more expressive (instead of in `explodeAndroidResSource.kt`).

src/main/kotlin/com/autonomousapps/model/internal/Source.kt
```diff
@@ -162,7 +161,7 @@ internal data class AndroidResSource(
        *
        * @see <a href="https://developer.android.com/guide/topics/resources/providing-resources#ResourcesFromXml">Accessing resources from XML</a>
        */
-      private val TYPE_REGEX = Regex("""@(\w+:)?(?<type>\w+)/(\w+)""")
+      private val TYPE_REGEX = Regex("""@(\w+:)?(?<type>\w+)/([\w.]+)""")
```
SimonMarquis, line 164: `.` are actually used for styles/themes, otherwise it breaks `detects theme set in manifest` and `detects theme set in manifest on Activity` tests.

SimonMarquis: I also added a unit test for this.

src/test/kotlin/com/autonomousapps/internal/ManifestParserTest.kt
```diff
@@ -150,26 +152,60 @@ class ManifestParserTest {
     assertThat(manifest.applicationName).isEqualTo("mutual.aid.explode")
   }
 
-  @Test fun `parse themes`() {
+  @Test fun `parse resource references`() {
```
SimonMarquis, line 155: In this test, I added common references that we are using in our own app. This is definitely not an exhaustive list of all places where references are allowed.

autonomousapps (approved):
I don't know why it took me so long to get to this - sorry about that. This looks good, but there are some conflicts. Do you mind resolving? Happy to merge once that is done.

PR #1466, opened 2025-06-13
Support typesafe project accessors in IssueHandler

timoloewe:
With support for type-safe project accessors, it would be nice to allow them in the plugin's configuration DSL.

This MR adds support for using type-safe project accessors as inputs in the `IssueHandler` DSL as an alternative to plain `projectPath` Strings.

Here's an example snippet how this would look like from one of our projects:
```kotlin
dependencyAnalysis {
    issues {
        project(projects.core.ui.impl) {
            onIncorrectConfiguration {
                exclude(projects.core.ui.api)
            }
        }
    }
}
```

As of today, we are already using project accessors in the plugin configuration similar to this, but it currently requires us to always convert them to a String ourselves using `.path`, e.g. `project(projects.core.ui.impl.path)`. This change would eliminate this inconvenience.

If you expect tests to be added for this, please point me to the place where they should go, and I will look into it.

One open question I had was whether `ProjectDependency` should also be allowed as an input to the `Bundle` APIs (like `primary()` and `includeDependency()`? I'm unsure if project dependencies are actually a thing there and supported by the plugin. If yes, I'm happy to make similar changes there.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/extension/Issue.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/Issue.kt
@@ -6,6 +6,7 @@ package com.autonomousapps.extension
 
 import org.gradle.api.InvalidUserDataException
 import org.gradle.api.artifacts.MinimalExternalModuleDependency
+import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderConvertible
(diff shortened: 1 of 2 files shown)


autonomousapps (approved):
Looks good, thanks!

I wouldn't object to some tests, but they could come as a follow-up if you also want to implement the `BundleHandler` changes you mentioned.

When writing specs for this project, I typically take something similar-ish, copy-paste into a new spec/project pair, and adjust as necessary for the new use-case. You could for example look at [SeverityOverlaySpec](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/src/functionalTest/groovy/com/autonomousapps/jvm/SeverityOverlaySpec.groovy) and make a new Spec class for this feature.

PR #1472, opened 2025-06-26
Updated guava version up to 33.4.8-jre

9tilov:
Update guava version up to 33.4.8-jre


autonomousapps:
Thanks for the PR. May I ask if you have a specific use-case for this version of Guava?

9tilov:
Yeah, I have build fail in our project because of guava versions conflict. I can explicitly define guava version but maybe it would be better just to update it here.

Issue #1323, opened 2024-11-26
Incorrect advice for testFixtures configuration in android project


zepurplez:
**Plugin version**
2.5.0

**Gradle version**
8.11

**JDK version**
17

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
1.9.24

**(Optional) Android Gradle Plugin (AGP) version**
8.5.2

**(Optional) `reason` output for bugs relating to incorrect advice**
```
------------------------------------------------------------
You asked about the dependency ':kt-lib'.
You have been advised to remove this dependency from 'testFixturesImplementation'.
------------------------------------------------------------

Shortest path from :lib to :kt-lib for debugCompileClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for debugRuntimeClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for debugUnitTestCompileClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for debugUnitTestRuntimeClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for debugAndroidTestCompileClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for debugAndroidTestRuntimeClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for releaseCompileClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for releaseRuntimeClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for releaseUnitTestCompileClasspath:
:lib
\--- :kt-lib

Shortest path from :lib to :kt-lib for releaseUnitTestRuntimeClasspath:
:lib
\--- :kt-lib

Source: debug, main
-------------------
* Uses 1 class: com.example.utils.SomeUtils (implies implementation).

Source: release, main
---------------------
* Uses 1 class: com.example.utils.SomeUtils (implies implementation).

Source: debug, test
-------------------
(no usages)

Source: release, test
---------------------
(no usages)

Source: debug, android_test
---------------------------
(no usages)
```

**Describe the bug**
So i have two modules, the first one is just pure kotlin with one util class
```kotlin
object SomeUtils {
    fun blabla() = println("Hey")
}
```

the second is android module with enabled testFixtures option and `android.experimental.enableTestFixturesKotlinSupport=true`

[...]

PR #1475, opened 2025-06-27
Fix: recognize test fixtures not only in libraries, but in app modules, too

Mexator:
See https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1323#issuecomment-3012027325


autonomousapps (approved):
LGTM! Thanks.

PR #1478, opened 2025-06-28
chore(graph-support): prepare next development version.

SimonMarquis:


autonomousapps:
The release process for this module deliberate doesn't include this step. Thanks though.

Issue #1476, opened 2025-06-27
Cannot run functional tests locally after new publication plugin was introduced


Mexator:
The test task fails with the following error

```
❯ ./gradlew :functionalTest --tests "com.autonomousapps.android.TestFixturesSpec.should not falsely report duplicated dependencies with main source set(#gradleVersion AGP #agpVersion)" > log

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':graph-support:signMavenPublication'.
> Cannot perform signing task ':graph-support:signMavenPublication' because it has no configured signatory
```

[log.txt](https://github.com/user-attachments/files/20945890/log.txt)

## Workaround

The workaround is to add the following lines to `graph-support/build.gradle.kts`
```
signing {
  setRequired { false }
}
```

But this is potentially dangerous.

autonomousapps:
Is this happening locally or on CI? I have [this configuration](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt#L79-L94) for this project to disable signing in some cases.

SimonMarquis:
https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/c54835289194c03735a7d31938eecf9a3cae8ec9/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt#L84-L87

As you said, this will keep the auto-signing enabled when both conditions are true:
- ✅ "Not on CI" -> we are using our own local envs
- ❌ "Not a snapshot" -> the main codebase is configured with this:
  https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/c54835289194c03735a7d31938eecf9a3cae8ec9/gradle.properties#L12
  but the issue is in the `:graph-support` module, where the version is overridden to a non-snapshot:
  https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/c54835289194c03735a7d31938eecf9a3cae8ec9/graph-support/build.gradle.kts#L7
  This module seems to have its own release cycle, but was never "reset" to a snapshot version.
  I've prepared a draft PR to update to the next release cycle if this is the solution you prefer to implement: 
  - https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1478

SimonMarquis:
[...]

PR #1480, opened 2025-07-01
test: disable signing when running :functionalTest.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1476.

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
@@ -76,20 +76,34 @@ internal class BaseConventionPlugin(private val project: Project) {
       .orElse("false")
       .map { it.toBoolean() }
 
+    val taskGraph = gradle.taskGraph
+    val isFunctionalTest: Provider<Boolean> = providers
+      .provider { taskGraph.hasTask(":functionalTest") }
+
     tasks.withType(Sign::class.java).configureEach { t ->
       with(t) {
         inputs.property("version", publishedVersion)
         inputs.property("is-ci", isCi)
+        inputs.property("is-functional-test", isFunctionalTest)
 
         // Don't sign snapshots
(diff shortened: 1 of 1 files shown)


build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
```diff
@@ -76,20 +76,34 @@ internal class BaseConventionPlugin(private val project: Project) {
...
         }
       }
     }
+
+    tasks.withType(DokkaTask::class.java) { t ->
+      t.inputs.property("is-functional-test", isFunctionalTest)
+
+      // Don't sign when running functional tests
+      t.onlyIf("not running functional tests") { !isFunctionalTest.get() }
+    }
```
autonomousapps, line 107: Dokka (v1 at least) is not CC-compatible. This doesn't resolve that though, because the task is still in the graph.

build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
```diff
@@ -76,20 +76,34 @@ internal class BaseConventionPlugin(private val project: Project) {
...
         inputs.property("version", publishedVersion)
         inputs.property("is-ci", isCi)
+        inputs.property("is-functional-test", isFunctionalTest)
 
         // Don't sign snapshots
         onlyIf("Not a snapshot") { !isSnapshot.get() }
         // We currently don't support publishing from CI
         onlyIf("release environment") { !isCi.get() }
+        // Don't sign when running functional tests
+        onlyIf("not running functional tests") { !isFunctionalTest.get() }
```
autonomousapps, line 94: Ideally this should be a little different: it should still sign if we're testing as part of the same flow as publishing to maven central. But the releasing notes already say to use `-x :functionalTest`, so I think it's not an issue in practice. Can return to this later if necessary.

Issue #490, opened 2021-09-30
String index out of range: -1 in task analyzeClassUsageDebug


blundell:
**Build scan link**
Sorry, can't share :)

**Plugin version**
0.78.0

**Gradle version**
7.1.1

**(Optional) Android Gradle Plugin (AGP) version**
7.0.0

**Describe the bug**
Exception
`Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1`

**To Reproduce**
1. run `projectHealth` on a specific module
(note we have several hundred modules, and this only happens on 1 of them)

**Expected behavior**
Not crash.

**Additional context**

stacktrace:
```
FAILED TASK: :featureA:analyzeClassUsageDebug
  Description: org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':featureA:analyzeClassUsageDebug'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:188)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:186)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:174)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
[...]

PR #1484, opened 2025-07-03
Add AGP 8.12 alpha

MGaetan89:
This PR adds AGP 8.12 alpha.

Additional changes:
- Remove the `ArbitraryFileSpec` spec.
- Update Gradle 7.6 to the latest fix (7.6.5 → 7.6.6).
- Update Gradle 8.14 to the latest fix (8.14.2 → 8.14.3).


src/functionalTest/groovy/com/autonomousapps/android/AbstractAndroidSpec.groovy
```diff
@@ -24,21 +24,23 @@ abstract class AbstractAndroidSpec extends AbstractFunctionalSpec {
   protected static final AGP_8_9 = AgpVersion.version('8.9.3')
   protected static final AGP_8_10 = AgpVersion.version('8.10.1')
   protected static final AGP_8_11 = AgpVersion.version('8.11.0')
+  protected static final AGP_8_12 = AgpVersion.version('8.12.0-alpha07')
```
MGaetan89: Android Studio Narwhal Feature Drop Canary 8 was released: https://androidstudio.googleblog.com/2025/07/android-studio-narwhal-feature-drop.html
But the corresponding AGP artefact is not yet available: https://maven.google.com/web/index.html#com.android.tools.build:gradle

MGaetan89:
It seems that it is no longer allowed to modify the properties of the `MergeResources` task: https://cs.android.com/android-studio/platform/tools/base/+/19999bb75f7f897df22cd1aca224d253181b9009
What would you like to happen to this test?

autonomousapps:
I think we can just delete this from `ArbitraryFileProject.groovy`:
```groovy
bs.withGroovy("""
afterEvaluate {
  tasks.withType(com.android.build.gradle.tasks.MergeResources).configureEach {
    aaptEnv.set("FOO")
  }
}""")
```
but otherwise leave the test alone.

autonomousapps:
> since the FOO resource file (created a few lines above) is not valid.

You get the same error as [this](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/490#issuecomment-933182073)?

PR #1486, opened 2025-07-07
chore: add API-tracking tasks using metalava.

autonomousapps:


build-logic/convention/src/main/kotlin/com/autonomousapps/convention/tasks/metalava/CheckApiTask.kt
```diff
@@ -0,0 +1,199 @@
...
+    public val output: RegularFileProperty
+  }
+
+  public abstract class Action : WorkAction<Parameters> {
+
+    @get:Inject public abstract val execOps: ExecOperations
+
+    override fun execute() {
+//      val outputDir = parameters.outputDir.get()
+//      outputDir.asFile.deleteRecursively() // TODO(tsr): do this in other cases where an @OutputDirectory is used
```
autonomousapps, line 83: Follow-up.

src/main/kotlin/com/autonomousapps/internal/advice/DslKind.kt
```diff
@@ -4,6 +4,7 @@ package com.autonomousapps.internal.advice
 
 import java.io.File
 
+// TODO(tsr): this is exposed in tasks as an input. Move to non-internal package.
```
autonomousapps, line 7: Follow-up.

PR #1498, opened 2025-07-22
chore: cleanup build script.

autonomousapps:
Inspired by https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1354.

### DIFF
--- a/build-logic/convention/build.gradle.kts
+++ b/build-logic/convention/build.gradle.kts
@@ -18,27 +18,16 @@ tasks.withType<ValidatePlugins>().configureEach {
   enableStricterValidation = true
 }
 
-val javaTarget = libs.versions.javaTarget.get()
-// val jdkVersion = libs.versions.jdkVersion.get()
-//
-// java {
-//   toolchain {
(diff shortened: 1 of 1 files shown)


build-logic/convention/build.gradle.kts
```diff
@@ -18,27 +18,16 @@ tasks.withType<ValidatePlugins>().configureEach {
...
-
   withType<GroovyCompile>().configureEach {
-    options.release = releaseTarget
+    options.release = libs.versions.javaTarget.map(String::toInt)
   }
   withType<JavaCompile>().configureEach {
-    options.release = releaseTarget
+    options.release = libs.versions.javaTarget.map(String::toInt)
   }
   withType<KotlinCompile>().configureEach {
```
TWiStErRob, line 28: @autonomousapps You shouldn't configure KotlinCompile directly, because it doesn't affect Kapt, maybe Ksp tasks for example. Using the top level `kotlin { compilerOptions { ...` wire things correctly.

PR #1506, opened 2025-07-31
fix: properly handle annotations (#1210)

### LINKED ISSUE #1210: Annotations used to annotate and things (enum constants, class literals, ...) used for annotation member values should not count to ABI
**Plugin version**
1.32.0

**Gradle version**
8.8

**JDK version**
17

```plain
You asked about the dependency 'com.fasterxml.jackson.core:jackson-annotations:2.17.1'.
There is no advice regarding this dependency.

Shortest path from root project to com.fasterxml.jackson.core:jackson-annotations:2.17.1 for compileClasspath:
:
\--- com.fasterxml.jackson.core:jackson-annotations:2.17.1

There is no path from root project to com.fasterxml.jackson.core:jackson-annotations:2.17.1 for runtimeClasspath


There is no path from root project to com.fasterxml.jackson.core:jackson-annotations:2.17.1 for testCompileClasspath


There is no path from root project to com.fasterxml.jackson.core:jackson-annotations:2.17.1 for testRuntimeClasspath


Source: main
------------
* Exposes 1 class: com.fasterxml.jackson.annotation.JsonIgnore (implies api).

Source: test
------------
(no usages)
```

**Describe the bug**
Annotations used to annotate and things (enum constants, class literals, ...) used for annotation member values should not be part of the ABI / always be `compileOnly`, no matter what retention the annotation has.

Annotations are never necessary to be in the compile classpath or runtime classpath of any downstream projects just for being present in the class file / at runtime.
The annotation parser that processes annotations in the class file just skips annotations for which the class is not present.

If some code wants to retrieve the annotation from a piece of code, then that "some code" is responsible to also have the annotation added to the classpath.

**To Reproduce**
```plain
./build.gradle.kts
plugins {
    `java-library`
    id("com.autonomousapps.dependency-analysis") version "1.32.0"
}

repositories {
    mavenCentral()
}

dependencies {
    compileOnly("com.fasterxml.jackson.core:jackson-annotations:2.17.1")
}

./gradle/wrapper/gradle-wrapper.properties
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
distributionUrl=https\://services.gradle.org/distributions/gradle-8.8-bin.zip
networkTimeout=10000
validateDistributionUrl=true
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists

./settings.gradle.kts
rootProject.name = "foo"

./src/main/java/foo/Foo.java
package foo;

import com.fasterxml.jackson.annotation.JsonIgnore;

public class Foo {
    @JsonIgnore
    public Object foo() {
        return null;
    }
}
```

**Expected behavior**
Annotations that are only used to annotate stuff, and things only used for annotation member values like class literaly, enum constants, ... should always be considered and recommended `compileOnly`.

Currently in the given example the advice is to change to `api` because a public method is annotated.
If a private method or field is annotated, `implementation` is advised.

If an annotation is of course used in some other way, like as class literal outside an annotation member value, or as variable type, or as return type, or as parameter type and so on, then advising as `implementation` or `api` is perfectly correct of course.

**Additional context**
Another example would be, if you develop a library that works stand-alone but is also DI friendly, supporting Spring, CDI, and Guice.
The library developer adds annotations for all three frameworks to this class.
All those annotations are runtime retention.
But you do not want to enforce any specific or any at all on downstream projects.

Imho annotations that are only used to annotate things (and of course then also the types used for the annotation member values) should always be `compileOnly`.
In case you strongly disagree with this and without any chance to be convinced to change the logic, please at least provide some feature-switch to treat it like described here.

Vampire:
Fixes #1210

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/AnnotationsImplementationProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/AnnotationsImplementationProject.groovy
@@ -13,8 +13,10 @@ import static com.autonomousapps.kit.gradle.dependencies.Dependencies.junit
 final class AnnotationsImplementationProject extends AbstractProject {
 
   final GradleProject gradleProject
+  private final boolean visible
 
-  AnnotationsImplementationProject() {
+  AnnotationsImplementationProject(boolean visible = true) {
+    this.visible = visible
     this.gradleProject = build()
   }
 
@@ -26,6 +28,11 @@ final class AnnotationsImplementationProject extends AbstractProject {
           bs.plugins = javaLibrary
           bs.dependencies(
             project('testImplementation', ':testrunner'),
+            project('testCompileOnly', ':annotation-with-class-and-enum'),
+            project('testCompileOnly', ':annotation'),
+            project(visible ? 'testImplementation' : 'testCompileOnly', ':class'),
+            project('testCompileOnly', ':member-annotation'),
+            project('testCompileOnly', ':enum'),
             junit('testImplementation'),
           )
         }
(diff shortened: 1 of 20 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/AnnotationsImplementationProject.groovy
```diff
@@ -79,12 +122,154 @@ final class AnnotationsImplementationProject extends AbstractProject {
       .build(),
   ]
 
+  private annotationWithClassAndEnumSource() {
+    return [
+      Source.java(
+        """\
+          // Deliberate in the same package as the consumer, to avoid import-based detection heuristics
+          package com.example.test;
```
autonomousapps, line 130: Thanks for this comment 👍

src/main/kotlin/com/autonomousapps/model/internal/intermediates/Reason.kt
```diff
@@ -177,24 +177,12 @@ internal sealed class Reason(open val reason: String) {
...
    */
   @TypeLabel("annotation")
   @JsonClass(generateAdapter = false)
   data class Annotation(override val reason: String) : Reason(reason) {
     constructor(inAnnotationClasses: Set<String>) : this(
-      buildReason(inAnnotationClasses, "Uses (in an annotation)", Kind.Class)
-    )
-
-    // TODO: ugh.
-    override val configurationName: String = "implementation, sometimes"
```
autonomousapps, line 190: yay

autonomousapps (commented):
This looks good! Thanks for the contribution.

Could you add a line to the changelog that summarizes this, including a brief explanation as to why we're doing this? This is a timely change, since my next release will be 3.0.0, and this is a breaking change.

PR #1525, opened 2025-08-21
fix resource ref problem

magicalmcgrady:
My modification moved the container variable into the method that generates ExplodedRes, ensuring that the reference relationships of individual resources are not polluted by project-level reference relationships.


autonomousapps:
Thanks for the PR. Could you describe the issue you encountered that led to this?

magicalmcgrady:
image
image
For example, this project has dimens resources and drawable resources. When using the plugin to count resource references, you will find that only "text_edit_intelligent_color_width" should be counted as referenced, but in fact other drawable resources are also counted as referenced resources of the dimens file.

magicalmcgrady:
how to merge this pull request？ @autonomousapps

autonomousapps (approved):

PR #1530, opened 2025-08-22
fix: set apiVersion and languageVersion for better Gradle 8 support.

autonomousapps:


build-logic/convention/src/main/kotlin/com/autonomousapps/convention/internal/kotlin/configureKotlin.kt
```diff
@@ -21,8 +22,12 @@ internal fun Project.configureKotlin(versionCatalog: VersionCatalog) {
 
   tasks.withType(KotlinCompile::class.java).configureEach { t ->
     t.compilerOptions {
+      // Ensure compatibility with Gradle 8.x. See https://docs.gradle.org/9.0.0/userguide/compatibility.html.
+      apiVersion.set(KotlinVersion.KOTLIN_1_8)
```
3flex: If Gradle 8.11 is the minimum then 2.0 is OK as 8.11 embeds Kotlin 2.0.20. You could use 2.1 as metadata versions are forwards compatible +1 version.

```suggestion
      apiVersion.set(KotlinVersion.KOTLIN_2_0)
```

build-logic/convention/src/main/kotlin/com/autonomousapps/convention/internal/kotlin/configureKotlin.kt
```diff
@@ -21,8 +22,12 @@ internal fun Project.configureKotlin(versionCatalog: VersionCatalog) {
...
+      languageVersion.set(KotlinVersion.KOTLIN_1_8)
```
3flex: As above.

```suggestion
      languageVersion.set(KotlinVersion.KOTLIN_2_0)
```

Issue #1443, opened 2025-05-12
A generic type is not recognized as `api` when exposed as return type from a suspend function


timoloewe:
**Plugin version**

2.17.0

**Gradle version**

8.12

**JDK version**

21

**Kotlin and Kotlin Gradle Plugin (KGP) version**

2.1.20

**(Optional) `reason` output for bugs relating to incorrect advice**

**Describe the bug**

If a generic type is only exposed as a return type from a suspend function, the plugin does not recognize the type as being part of the public API and thus advises to change the dependency from `api` to `implementation`.

We have a network module that defines a generic `ApiResult<T>` type:
```kotlin
sealed class ApiResult<T> { ... }
```
In other modules, we depend on it and expose it as a return type on a public interface like this:
```kotlin
interface FooService {
    suspend fun getFoo(): ApiResult<Foo>
} 
```
The plugin then generates an incorrect advice to change the dependency from `api` to `implementation`:
```
Existing dependencies which should be modified to be as indicated:
  implementation(project(":network:api")) (was api)
```

If the type in question is not a generic type or if the function is not a suspend function, the advice is not given.
```kotlin
interface FooService {
    // Wrong advice is given:
    suspend fun getFoo(): ApiResult<Foo>

    // Not suspending, no advice given:
    fun getFoo(): ApiResult<Foo> 

    // Return type not generic, no advice given:
    suspend fun getFoo(): ApiResult.Error 
}
```

**To Reproduce**

Here is a reproducer project: [dagp-suspend-repro.zip](https://github.com/user-attachments/files/20164888/dagp-suspend-repro.zip)

Running `buildHealth` on this project generates the incorrect advice.

**Expected behavior**

The plugin should not advise changing the dependency to `implementation` since the type is exposed in the public API no matter if the function is suspending or not.

autonomousapps:
Thanks for the report and reproducer.

eygraber:
Dupe of https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1324?

eygraber:
@autonomousapps I was testing out Jules (Google's new agent thing) and threw it at this issue. 

[...]

Issue #1324, opened 2024-11-26
Return type not considered for public API


eygraber:
**Plugin version**
2.5.0

**Gradle version**
8.11.1

**JDK version**
22

**Kotlin and Kotlin Gradle Plugin (KGP) version**
2.0.21

**Android Gradle Plugin (AGP) version**
8.7.2

**Describe the bug**
I have 3 modules:

module A
```kotlin
interface MyResult
```

module B (`api(projects.A)`)
```kotlin
interface Foo {
  fun bar(): MyResult
}
```

module C (`implementation(projects.B)`, `implementation(projects.A)`)
```kotlin
class Baz(
  private val foo: Foo,
) {
  fun bing() {
    foo.bar()
  }
}
```
Running `buildHealth` fails with:

```
> Advice for :B
  Existing dependencies which should be modified to be as indicated:
    implementation(projects.A) (was api)
  
  Advice for :C
  Unused dependencies which should be removed:
    implementation(projects.A)
```

Running `:B:reason --id :A`:

```
------------------------------------------------------------
You asked about the dependency ':A'.
You have been advised to change this dependency to 'implementation' from 'api'.
------------------------------------------------------------

Source: devDebug, main
----------------------
* Imports 1 class: MyResult (implies implementation).
```

Running `:B:reason --id :A`:
```
------------------------------------------------------------
You asked about the dependency ':A'.
You have been advised to remove this dependency from 'implementation'.
------------------------------------------------------------

Source: devDebug, main
----------------------
(no usages)
```

If I follow the advice then compiling `C` fails:

```
e: Baz.kt:20:52 Cannot access class 'MyResult'. Check your module classpath for missing or conflicting dependencies.
```

eygraber:
[repro.zip](https://github.com/user-attachments/files/17931831/repro.zip)

Running `./gradlew buildHealth` on the repro above fails only when `foo` is a suspending function in `services/foo/public/src/main/kotlin/template/services/foo/Foo.kt`.

autonomousapps:
Thanks for the report. I note the description of the issue (in the text) doesn't include any `suspend` functions. Does the repro contain suspend functions? (I haven't downloaded it yet.) If so, could you update the description?

eygraber:
Updated. The repro project looks pretty different though, the description here was originally derived from that.

PR #1553, opened 2025-09-11
Consider suspend function return types with type parameters

eygraber:
Fixes https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1324
Fixes https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1443

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
@@ -15,8 +15,13 @@ import java.io.File
 import java.io.InputStream
 import java.io.PrintStream
 import java.util.jar.JarFile
+import kotlin.metadata.KmClassifier
+import kotlin.metadata.KmType
+import kotlin.metadata.isSuspend
 import kotlin.metadata.jvm.JvmFieldSignature
 import kotlin.metadata.jvm.JvmMethodSignature
+import kotlin.metadata.jvm.KotlinClassMetadata
+import kotlin.metadata.jvm.signature
 
 internal fun main(args: Array<String>) {
   val src = args[0]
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
```diff
@@ -213,3 +235,24 @@ internal fun <T : Appendable> List<ClassBinarySignature>.dump(to: T): T = to.app
     appendReproducibleNewLine("}\n")
   }
 }
+
+private fun extractClassesAsDescriptors(type: KmType): Set<String> {
+  val classes = mutableSetOf<String>()
+
+  when (val classifier = type.classifier) {
+    is KmClassifier.Class -> classes.add("L${classifier.name};")
+    is KmClassifier.TypeAlias -> classes.add("L${classifier.name};")
```
autonomousapps, line 244: These two cases are identical. Could we collapse the two branches into one?

src/main/kotlin/com/autonomousapps/internal/kotlin/PublicApiDump.kt
```diff
@@ -213,3 +235,24 @@ internal fun <T : Appendable> List<ClassBinarySignature>.dump(to: T): T = to.app
...
+
+    // Type parameters are not concrete types, so we don't add them directly.
+    // Their bounds might be relevant, but the requirement is to extract constituent types.
+    is KmClassifier.TypeParameter -> Unit
+  }
+
+  type.arguments.forEach { projection ->
+    projection.type?.let { argumentType ->
+      classes.addAll(extractClassesAsDescriptors(argumentType))
```
autonomousapps, line 253: Oo, recursion ➰

PR #1555, opened 2025-09-11
Consider imports when computing usage of typealiases

eygraber:
Fixes https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1556

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypealiasProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypealiasProject.groovy
@@ -5,6 +5,7 @@ package com.autonomousapps.jvm.projects
 import com.autonomousapps.AbstractProject
 import com.autonomousapps.kit.GradleProject
 import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.gradle.dependencies.Dependencies
 import com.autonomousapps.model.ProjectAdvice
 
 import static com.autonomousapps.AdviceHelper.actualProjectAdvice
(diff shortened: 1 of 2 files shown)


src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypealiasProject.groovy
```diff
@@ -78,6 +88,24 @@ final class TypealiasProject extends AbstractProject {
...
+      import kotlinx.coroutines.CancellationException
+      
+      private class ExternalAliasConsumer {
+        private fun throwCancellationException() {
+          throw CancellationException()
+        }
+      }
+      """
+    )
+      .withPath('com.example.external.alias.consumer', 'ExternalAliasConsumer')
```
autonomousapps, line 105: fyi this is no longer necessary, as the `kotlin()` (and related) static functions infer the package and classname. But it's also fine.

eygraber: Oh nice! If I remove it, should I remove it from the other sources in the file for consistency?

eygraber:
> I'm curious why the used-classes test isn't catching this. Do you have any thoughts on that?

Which spec is that in?

My initial thought was that it was related to typealiases and KMP, but it looks like the problem is related to a typealias that targets a stdlib type (e.g. `typealias MyAlias = java.util.concurrent.CancellationException`).

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypealiasProject.groovy
autonomousapps: Sure, if you want! Thanks.

PR #1557, opened 2025-09-14
wip. Instead of building capability string from group:name, just use name.

autonomousapps:
This avoids use of project.getGroup() from ProjectDerivedCapability.


src/main/kotlin/com/autonomousapps/internal/utils/gradleStrings.kt
```diff
@@ -297,8 +302,39 @@ internal fun ResolvedVariantResult?.toGradleVariantIdentification(): GradleVaria
...
+    emptyMap(),
   )
 }
 
-private fun Capability.toGA() = "$group:$name".intern()
+// TODO(tsr) cleanup
+private fun Capability.toGA(): String {
+  return name
+//  return "$group:$name".intern()
+}
```
autonomousapps: If this works, it makes my life a lot easier.

PR #1565, opened 2025-10-06
Fix: Test fixtures in a Spring Boot project are not allowed to have testFixturesApi dependencies

### LINKED ISSUE #1554: Test fixtures in a Spring Boot project are not allowed to have testFixturesApi dependencies
**Plugin version**

`3.0.2`

**Gradle version**
`9.0.0`

**JDK version**
`21`

**Describe the bug**

When having test fixtures in a project that is also a Spring Boot Application (uses `org.springframework.boot` plugin), all test fixtures dependencies are suggested to have `testFixturesImplementation` scope even if they are `testFixturesApi`.

The reason for this is likely this piece of code that turns APIs off in general (for all source sets) of a Spring boot project:

https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/061ce79319c2ff275e895d111b7a1363e398df0d/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt#L352-L361

**To Reproduce**

- Add `Plugins.springBoot` here: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/526803b0ea9dcfe9a73eee1b5a9e2ab20d1c1517/src/functionalTest/groovy/com/autonomousapps/jvm/projects/TestFixturesTestProject.groovy#L50
- See how `CustomSourceSetSpec.dependencies to test fixtures` fails.

**Expected behavior**

`testFixturesApi` , and any other `myCustomFeatureApi`, should be support even if the project is a Spring boot project.

Solution options:
1. Solve the special case in a way that it only turns off api for the "main" source set
2. Remove the special case altogether (?) - if there is `api` or `*Api`, use it. 
3. Introduce an option to turn special treatment of Spring boot on/off

**Additional context**

- If I understand the problem correctly, this is a special case not considered in the custom source set support
- I can have a look at doing a PR for this, once we know what it the wanted behavior.

jjohannes:
Resolves #1554 
Follow up to #852

### DIFF
--- a/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
+++ b/src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
@@ -592,38 +592,30 @@ internal class ProjectPlugin(private val project: Project) {
       return
     }
 
+    if (pluginManager.hasPlugin(SPRING_BOOT_PLUGIN)) {
+      logger.warn(
+        "(dependency analysis) You have both java-library and org.springframework.boot applied. You probably " +
+                "want java, not java-library."
+      )
+    }
+
     j.sourceSets.forEach { sourceSet ->
       try {
         val sourceKind = JvmSourceKind.of(sourceSet.name)
         val hasAbi = hasAbi(sourceSet)
-
-        // Regardless of the fact that this is a "java-library" project, the presence of Spring
-        // Boot indicates an app project.
-        val dependencyAnalyzer = if (pluginManager.hasPlugin(SPRING_BOOT_PLUGIN)) {
-          logger.warn(
-            "(dependency analysis) You have both java-library and org.springframework.boot applied. You probably " +
-              "want java, not java-library."
+        val dependencyAnalyzer = if (hasAbi) {
+          JavaWithAbiAnalyzer(
+            project = this,
+            sourceSet = sourceSet,
+            sourceKind = sourceKind,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -592,38 +592,30 @@ internal class ProjectPlugin(private val project: Project) {
       return
     }
 
+    if (pluginManager.hasPlugin(SPRING_BOOT_PLUGIN)) {
+      logger.warn(
+        "(dependency analysis) You have both java-library and org.springframework.boot applied. You probably " +
+                "want java, not java-library."
+      )
+    }
```
jjohannes, line 600: Moved up from below. Missed this in #852 and the warning would currently be emitted several times (once per source set) which makes no sense. There is no test for this warning. Do you want to keep it or should we remove it?

Turns out that this is unrelated to the fix. The use case most (all?) folks have is combining these three plugins: `java`, `springframework.boot`, `java-test-fixtures`. If you do that, you won't get the warning.

autonomousapps: If you think the warning is correct, let's leave it. Probably worth adding a test (which I can do later).

autonomousapps (approved):
Thank you!

Issue #1563, opened 2025-09-30
DAGP not compatible with IP


jvandort:
**Build scan link**
https://ge.gradle.org/s/ol5pb63unmnd2/failures?expanded-stacktrace=WyIwLTEtMiJd&focused-exception-line=0-1-2-0

**Plugin version**
3.0.1

**Gradle version**
9.x nightly

**JDK version**
17

**Describe the bug**
NPEs and build process hangs when IP is enabled. 

See: 
[deadlock.txt](https://github.com/user-attachments/files/22622145/deadlock.txt)

And: 
```
Caused by: java.lang.NullPointerException: Cannot read field "right" because "l" is null
at org.gradle.api.internal.collections.SortedSetElementSource.add(SortedSetElementSource.java:104)	
at org.gradle.api.internal.DefaultDomainObjectCollection.doAdd(DefaultDomainObjectCollection.java:278)	
at org.gradle.api.internal.DefaultNamedDomainObjectCollection.doAdd(DefaultNamedDomainObjectCollection.java:126)	
at org.gradle.api.internal.AbstractNamedDomainObjectContainer.create(AbstractNamedDomainObjectContainer.java:56)	
at org.gradle.api.internal.AbstractNamedDomainObjectContainer.maybeCreate(AbstractNamedDomainObjectContainer.java:66)	
at com.autonomousapps.services.GlobalDslService.project$dependency_analysis_gradle_plugin(GlobalDslService.kt:205)	
at com.autonomousapps.extension.IssueHandler.project(IssueHandler.kt:38)	
at com.autonomousapps.DependencyAnalysisSubExtension.issues(DependencyAnalysisSubExtension.kt:46)	
at Gradlebuild_jvm_library_gradle._init_$lambda$1(gradlebuild.jvm-library.gradle.kts:41)
```
(from above linked build scan)

**To Reproduce**
[...]

PR #1584, opened 2025-10-28
Improve IP safety of GlobalDslService

jvandort:
With IP enabled, GlobalDslService will be accessed concurrently at configuration-time by multiple projects. Previously, the serviced use NamedDomainObjectContainers as a generic container for DSL elements to be accessed from multiple projects. DOCs are not thread safe. Further, the provider API is not completely thread safe, though operations like map and flatMap should be, as they return new Provider instances. Finally, objects created using ObjectFactory.newInstance are likely not completely thread safe, as there is internal laziness for example when fetching injected service instances.

This commit reworks GlobalDslService by synchronizing access to shared DSL objects. The 'all' ProjectIssueHandler is unconditionally synchronized. The 'projects' container has been migrated to a ConcurrentHashMap, and access to any individual element of the container is synchronized using 'ConcurrentHashMap#compute'. This allows us to keep these objects internally mutable/not thread safe, as they themselves are instantiated by Objects.newInstance, and also contain non-thread-safe constructs like SourceSets, DomainObjectContainers, and Properties.

This implementation assumes the 'overlay' method performs no mutable operations, and only performs safe provider-based operations like map and flatMap, which produce new providers.

This commit passes ./gradlew check locally, but I have not run any additional IP tests. Nor have I added additional integration tests for IP, as IP is not yet stable. However, I believe this should fix https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1563

### DIFF
--- a/src/main/kotlin/com/autonomousapps/services/GlobalDslService.kt
+++ b/src/main/kotlin/com/autonomousapps/services/GlobalDslService.kt
@@ -16,6 +16,7 @@ import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
+import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 
 /**
@@ -24,7 +25,7 @@ import javax.inject.Inject
  * terms) configure the entire build, globally, without any subproject touching mutable properties of any other project.
  */
 public abstract class GlobalDslService @Inject constructor(
-  objects: ObjectFactory,
+  private val objects: ObjectFactory,
 ) : BuildService<BuildServiceParameters.None> {
 
   // Used for validity check. The plugin must be registered on the root project.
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/services/GlobalDslService.kt
```diff
@@ -245,18 +254,25 @@ public abstract class GlobalDslService @Inject constructor(
   }
 
   internal fun redundantPluginsIssueFor(projectPath: String): Provider<Behavior> {
-    return overlay(all.redundantPluginsIssue, projects.findByName(projectPath)?.redundantPluginsIssue)
+    return overlay(all.redundantPluginsIssue, getForProject(projectPath) { it.redundantPluginsIssue })
```
autonomousapps, line 257: I don't entirely understand this change. Why not more simply:
```kotlin
projects.get(projectPath)?.redundantPluginsIssue
```

autonomousapps (commented):
Thank you! Left a question.

Also, just to clarify in public (vs a private conversation), there's no harm in leaving the `SourceSetHandler` internal to `ProjectIssueHandler` as it is? That is, using named domain object containers?

src/main/kotlin/com/autonomousapps/services/GlobalDslService.kt
jvandort: Your suggested code is probably fine as-is, but I opted for the consistency route, piping _all_ accesses to `projects` through `getForProject`

`redundantPluginsIssue` is instantiated like:

```
  internal val redundantPluginsIssue = objects.newInstance(Issue::class.java)
```

so it exists after `ProjectIssueHandler` is constructed and the property access is thread-safe.

However, my implementation protects against future changes to `ProjectIssueHandler` where `redundantPluginsIssue` may be changed to something more dynamic like a Kotlin property, or a Gradle `@Nested` object (where Gradle manages the instantiation of nested `Issue` objects)

Issue #1588, opened 2025-10-30
ABI allow inclusions


ArtyomGabeev:
**Is your feature request related to a problem? Please describe.**
Currently ABI configuration supports only excludeClasses, excludeAnnotations.
When working with existing codebase, which does not follow package convention, it's hard to differentiate between public/internal api, bcz everything are essentially public, leading to dependency exposure if I follow plugin advices.

**Describe the solution you'd like**
Add to the abi { exclusions {} } possibility to configure:
- includeClasses('some.package')
- includeAnnotations('PublicApi')

In order to allow graceful migration and enforcements.

**Describe alternatives you've considered**
1. Major refactoring to use packages to differentiate api/internal package, will cause backward compatibility issues on the client and mess with git history. However, package separation is a common practice and used by java modules or ArchUnit.
2. As a workaround I came up with exclude everything except:
```
dependencyAnalysis {
    abi {
        exclusions {
            def includes = [
                'some.package.'
            ]
            def includesQuoted = includes.collect { Pattern.quote(it) }
            def excludeRegex = '^' + includesQuoted.collect { "(?!${it})" }.join('') + '.*$'

            ignoreGeneratedCode()
            excludeAnnotations('org.springframework.boot.autoconfigure.AutoConfiguration')
            excludeClasses(excludeRegex)
        }
    }
}
```

**Additional context**
If it makes sense I can contribute the change.
What I am not sure, how existing exclusion work with transitive class dependencies.
E.g. if we introduce inclusion configuration and include single class, I assume it will be challenging and non-explicit to add all dependent classes from class ABI.

autonomousapps:
Thanks for the issue. I think that, given we already support exclusions for ABI analysis, it would be symmetrical to also support inclusions. The way I think it would work is, if there are include rules, then anything not meeting those rules is effectively excluded. Is that also your understanding? I believe that is what you're saying, and just want to double-check.

ArtyomGabeev:
Sounds good, and additionally I would suggest to apply exclusion rules as well, as a secondary filter. E.g.
```
includeClasses('com.some.')
excludeAnnotations('.Generated')
```

Something like this:

included = (noIncludes || matchesAny(include))
excluded = matchesAny(exclude)
participates = included && !excluded

autonomousapps:
I agree. Is this something you're interested in contributing? I won't have time for the foreseeable future.

PR #1591, opened 2025-11-03
Support ABI inclusions

ArtyomGabeev:
Fixes https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1588

Top level DSL left to be exclusions for backward compatibility.

Example usage:
```
dependencyAnalysis {
  abi {
    exclusions {
      includeClasses(".*\\.api\\..*")
      excludeAnnotations('.Generated')
    }
  }
}
```

### DIFF
--- a/src/main/kotlin/com/autonomousapps/extension/AbiHandler.kt
+++ b/src/main/kotlin/com/autonomousapps/extension/AbiHandler.kt
@@ -18,6 +18,8 @@ import javax.inject.Inject
  *     exclusions {
  *       excludeSourceSets(/* source sets to exclude from ABI analysis */)
  *
+ *       includeClasses(".*\\api\\..*")
+ *       includeAnnotations(".*\\.PublicApi")
  *       ignoreSubPackage("internal")
  *       ignoreInternalPackages()
  *       ignoreGeneratedCode()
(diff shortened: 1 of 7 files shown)


autonomousapps:
> I did not found where wiki documentation is located, it should be updated as well.

Unfortunately, the wiki isn't tracked in version control. I don't know why Github does that. If you provide sample text, I can update the wiki manually.

src/main/kotlin/com/autonomousapps/extension/AbiHandler.kt
```diff
@@ -40,11 +42,29 @@ public abstract class AbiHandler @Inject constructor(objects: ObjectFactory) {
...
+   */
+  public fun includeClasses(@Language("RegExp") vararg classRegexes: String) {
+    classInclusions.addAll(*classRegexes)
+  }
+
+  /**
+   * Include any class into ABI analysis that is annotated with annotations that match [annotationRegexes], which means
+   * those classes will be considered as part of the module's ABI.
+   */
+  public fun includeAnnotations(@Language("RegExp") vararg annotationRegexes: String) {
```
autonomousapps, line 64: Is the inverse also true (as above), which is that once an inclusion is set, anything that doesn't match _isn't_ considered part of the ABI?

ArtyomGabeev: Yep, that correct.

autonomousapps: Could you update that kdoc then? The wording isn't clear on that point. The other method you added _is_ clear about that though.

ArtyomGabeev: Hey @autonomousapps ,

Sorry, It took longer than I would expect to update the kdoc. Should be more clear now.

PR #1602, opened 2025-11-18
feat: check binary compatibility between consumer and producer code.

autonomousapps:


src/main/kotlin/com/autonomousapps/model/internal/intermediates/Reason.kt
```diff
@@ -108,15 +108,16 @@ internal sealed class Reason(open val reason: String) {
           nonMatches.forEachIndexed { i, (access, incompatibleMembers) ->
             if (i > 0) appendLine()
 
+            // TODO(tsr): I don't like all this going back and forth between dotty and slashy representations.
```
autonomousapps: Can we fix this? 🤔

src/main/kotlin/com/autonomousapps/internal/asm.kt
```diff
@@ -765,6 +765,12 @@ internal class KotlinMetadataVisitor(
   }
 }
 
+// TODO(tsr): look at asmUtils' `ClassNode.isEffectivelyPublic()` method, which is substantially different.
+private fun isEffectivelyPublic(access: Int): Boolean {
+  val flags = AccessFlags(access)
+  return !flags.isSynthetic && (flags.isPublic || flags.isProtected)
+}
```
autonomousapps, line 770: Do this in a follow-up.

Issue #1637, opened 2026-02-06
Expose usage report as part of dependency analysis


Laimiux:
**Is your feature request related to a problem? Please describe.**

I would like to be able to get a report around dependency usages per Gradle module and a summary for the whole project. These reports would
- Provide a signal around module complexity
- Capture most frequently used APIs
- Enable a way to track migrations such as RxJava -> Coroutines

I'm including an example report that I generated using intermediary types already generated by this plugin
```json
{
    "module": ":instacart-flyers:public",
    "summary": {
        "dependencyTypes": 6,
        "dependencyTypeUsages": 10,
        "internalTypes": 4,
        "projectDependencies": 2,
        "libraryDependencies": 3
    },
    "internal": {
        "com.instacart.client.flyers.eligibility.ICFlyersBadgeFormula$DefaultImpls": 1,
        "com.instacart.client.flyers.eligibility.ICFlyersBadgeState$Companion": 1,
        "com.instacart.client.flyers.eligibility.ICFlyersBadgeFormula": 1,
        "com.instacart.client.flyers.eligibility.ICFlyersBadgeState": 1
    },
    "projectDependencies": {
        ":instacart-analytics:public": {
            "com.instacart.analytics.latency.ICPageKey": 3
        },
        ":instacart-global-home-tabs:public": {
            "com.instacart.client.ui.pages.ICTabFeature": 2
        }
    },
    "libraryDependencies": {
        "com.instacart.formula:formula": {
            "com.instacart.formula.IFormula": 2,
            "com.instacart.formula.IFormula$DefaultImpls": 1
        },
        "com.instacart.formula:formula-android": {
            "com.instacart.formula.android.RouteKey": 1
        },
        "io.reactivex.rxjava3:rxjava": {
            "io.reactivex.rxjava3.core.Observable": 1
        }
    }
}
```

**Describe the solution you'd like**

This plugin is already doing deep usage analysis to be able to determine if a dependency is used or not. The ideal solution would provide first-class APIs to get access to this usage data. There are a couple of options here:
- Generate a first-class usage report similar `advise.json`
- Provide an extension API that provides direct access to the data models. The user can generate any kind of report using the data.

**Regarding the implementation:**
- To generate report, we need access to `ProjectVariant` and `Dependencies` types to generate the r
- There already exists `ComputeUsagesTask` that has access to both. It generates `dependency-trace-report.json`. 
[...]

PR #1639, opened 2026-02-10
A new task to build type usage report (ComputeTypeUsageTask).

Laimiux:
### Summary 
A new task to build type usage report (#1637).

Introducing a new `ComputeTypeUsageTask` task. It generates `type-usage.json` reporting which types (classes) your code uses from each dependency. Designed for analyzing module dependency usage and complexity.
```json
{
  "projectPath": ":app",
  "summary": {
    "totalTypes": 245,
    "totalFiles": 67,
    "internalTypes": 12,
    "projectDependencies": 3,
    "libraryDependencies": 18
  },
  "internal": {
    "com.example.MyClass": 5
  },
  "projectDependencies": {
    ":core": {
      "com.example.core.Utils": 2
    }
  },
  "libraryDependencies": {
    "org.apache.commons:commons-collections4": {
      "org.apache.commons.collections4.bag.HashBag": 3
    }
  }
}
```

**Usage:**
```bash
./gradlew computeTypeUsageMain
```

**Output location:**
```
build/reports/dependency-analysis/main/intermediates/type-usage.json
```

**Configuration:**
```kotlin
dependencyAnalysis {
  typeUsage {
    excludePackages("kotlin.jvm.internal")
    excludeTypes("kotlin.Unit")
    excludeRegex(".*_Factory$", ".*Companion$")
  }
}
```

**Key design decisions:**
- Task runs **on-demand only** (not part of `buildHealth` or other critical paths)
- Uses existing intermediate files (`synthetic-project.json`, `exploded-jars.json.gz`)
- Tracks both annotation and non-annotation class usage
- Sensible defaults exclude common generated/internal types

### DIFF
--- a/api/api.txt
+++ b/api/api.txt
@@ -4,6 +4,7 @@ package com.autonomousapps {
   public abstract class AbstractExtension {
     ctor @javax.inject.Inject public AbstractExtension(org.gradle.api.model.ObjectFactory objects, org.gradle.api.invocation.Gradle gradle);
     method public final org.gradle.api.file.RegularFileProperty adviceOutput();
+    method public final org.gradle.api.file.RegularFileProperty typeUsageOutput();
     method public final void app();
     method public final void registerPostProcessingTask(org.gradle.api.tasks.TaskProvider<? extends com.autonomousapps.AbstractPostProcessingTask> task);
     field public static final String NAME = "dependencyAnalysis";
@@ -28,6 +29,7 @@ package com.autonomousapps {
     method public final void issues(org.gradle.api.Action<com.autonomousapps.extension.IssueHandler> action);
     method public final void reporting(org.gradle.api.Action<com.autonomousapps.extension.ReportingHandler> action);
     method public final void structure(org.gradle.api.Action<com.autonomousapps.extension.DependenciesHandler> action);
+    method public final void typeUsage(org.gradle.api.Action<com.autonomousapps.extension.TypeUsageHandler> action);
     method public final void usage(org.gradle.api.Action<com.autonomousapps.extension.UsageHandler> action);
     method @Deprecated public final void usages(org.gradle.api.Action<com.autonomousapps.extension.UsageHandler> action);
     method public final void useTypesafeProjectAccessors(boolean enable);
@@ -44,6 +46,7 @@ package com.autonomousapps {
     method public final void abi(org.gradle.api.Action<com.autonomousapps.extension.AbiHandler> action);
     method public final void issues(org.gradle.api.Action<com.autonomousapps.extension.ProjectIssueHandler> action);
     method public final void structure(org.gradle.api.Action<com.autonomousapps.extension.DependenciesHandler> action);
+    method public final void typeUsage(org.gradle.api.Action<com.autonomousapps.extension.TypeUsageHandler> action);
   }
 
   public final class Flags {
@@ -215,6 +218,13 @@ package com.autonomousapps.extension {
     method @InaccessibleFromKotlin @org.gradle.api.tasks.Input public org.gradle.api.provider.Property<java.lang.String> getPostscript();
     property @org.gradle.api.tasks.Input public abstract org.gradle.api.provider.Property<java.lang.Boolean> onlyOnFailure;
     property @org.gradle.api.tasks.Input public abstract org.gradle.api.provider.Property<java.lang.String> postscript;
+  }
+
+  public abstract class TypeUsageHandler {
+    ctor @javax.inject.Inject public TypeUsageHandler(org.gradle.api.model.ObjectFactory objects);
+    method public final void excludePackages(java.lang.String... packages);
+    method public final void excludeRegex(@org.intellij.lang.annotations.Language("RegExp") java.lang.String... patterns);
+    method public final void excludeTypes(java.lang.String... types);
   }
 
   public abstract class SourceSetsHandler implements org.gradle.api.Named {
@@ -517,6 +527,48 @@ package com.autonomousapps.model {
     property public String? buildPath;
     property public com.autonomousapps.model.GradleVariantIdentification gradleVariantIdentification;
     property public String identifier;
+  }
+
+  @com.squareup.moshi.JsonClass(generateAdapter=false) public final class ProjectTypeUsage implements java.lang.Comparable<com.autonomousapps.model.ProjectTypeUsage> {
+    ctor public ProjectTypeUsage(String projectPath, com.autonomousapps.model.TypeUsageSummary summary, java.util.Map<java.lang.String,java.lang.Integer> internal, java.util.Map<java.lang.String,? extends java.util.Map<java.lang.String,java.lang.Integer>> projectDependencies, java.util.Map<java.lang.String,? extends java.util.Map<java.lang.String,java.lang.Integer>> libraryDependencies);
+    method public int compareTo(com.autonomousapps.model.ProjectTypeUsage other);
+    method public String component1();
+    method public com.autonomousapps.model.TypeUsageSummary component2();
+    method public java.util.Map<java.lang.String,java.lang.Integer> component3();
+    method public java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> component4();
+    method public java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> component5();
+    method public com.autonomousapps.model.ProjectTypeUsage copy(String projectPath, com.autonomousapps.model.TypeUsageSummary summary, java.util.Map<java.lang.String,java.lang.Integer> internal, java.util.Map<java.lang.String,? extends java.util.Map<java.lang.String,java.lang.Integer>> projectDependencies, java.util.Map<java.lang.String,? extends java.util.Map<java.lang.String,java.lang.Integer>> libraryDependencies);
+    method public java.util.Map<java.lang.String,java.lang.Integer> getInternal();
+    method public java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> getLibraryDependencies();
+    method public java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> getProjectDependencies();
+    method public String getProjectPath();
+    method public com.autonomousapps.model.TypeUsageSummary getSummary();
+    method public boolean isEmpty();
+    property public final java.util.Map<java.lang.String,java.lang.Integer> internal;
+    property public final java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> libraryDependencies;
+    property public final java.util.Map<java.lang.String,java.util.Map<java.lang.String,java.lang.Integer>> projectDependencies;
+    property public final String projectPath;
+    property public final com.autonomousapps.model.TypeUsageSummary summary;
+  }
+
+  @com.squareup.moshi.JsonClass(generateAdapter=false) public final class TypeUsageSummary {
+    ctor public TypeUsageSummary(int totalTypes, int totalFiles, int internalTypes, int projectDependencies, int libraryDependencies);
+    method public int component1();
+    method public int component2();
+    method public int component3();
+    method public int component4();
+    method public int component5();
+    method public com.autonomousapps.model.TypeUsageSummary copy(int totalTypes, int totalFiles, int internalTypes, int projectDependencies, int libraryDependencies);
+    method public int getInternalTypes();
+    method public int getLibraryDependencies();
+    method public int getProjectDependencies();
+    method public int getTotalFiles();
+    method public int getTotalTypes();
+    property public final int internalTypes;
+    property public final int libraryDependencies;
+    property public final int projectDependencies;
+    property public final int totalFiles;
+    property public final int totalTypes;
   }
 
   @com.squareup.moshi.JsonClass(generateAdapter=false) public final class Warning implements java.lang.Comparable<com.autonomousapps.model.Warning> {
--- a/api/api.txt
+++ b/api/api.txt
@@ -4,9 +4,9 @@ package com.autonomousapps {
   public abstract class AbstractExtension {
     ctor @javax.inject.Inject public AbstractExtension(org.gradle.api.model.ObjectFactory objects, org.gradle.api.invocation.Gradle gradle);
     method public final org.gradle.api.file.RegularFileProperty adviceOutput();
-    method public final org.gradle.api.file.RegularFileProperty typeUsageOutput();
     method public final void app();
     method public final void registerPostProcessingTask(org.gradle.api.tasks.TaskProvider<? extends com.autonomousapps.AbstractPostProcessingTask> task);
+    method public final org.gradle.api.provider.MapProperty<java.lang.String,org.gradle.api.file.RegularFile> typeUsageOutputs();
     field public static final String NAME = "dependencyAnalysis";
   }
 
@@ -220,16 +220,16 @@ package com.autonomousapps.extension {
     property @org.gradle.api.tasks.Input public abstract org.gradle.api.provider.Property<java.lang.String> postscript;
   }
 
+  public abstract class SourceSetsHandler implements org.gradle.api.Named {
+    ctor @javax.inject.Inject public SourceSetsHandler(String sourceSetName, String projectPath, org.gradle.api.model.ObjectFactory objects);
+    method public String getName();
+  }
+
   public abstract class TypeUsageHandler {
     ctor @javax.inject.Inject public TypeUsageHandler(org.gradle.api.model.ObjectFactory objects);
     method public final void excludePackages(java.lang.String... packages);
     method public final void excludeRegex(@org.intellij.lang.annotations.Language("RegExp") java.lang.String... patterns);
     method public final void excludeTypes(java.lang.String... types);
-  }
-
-  public abstract class SourceSetsHandler implements org.gradle.api.Named {
-    ctor @javax.inject.Inject public SourceSetsHandler(String sourceSetName, String projectPath, org.gradle.api.model.ObjectFactory objects);
-    method public String getName();
   }
 
   public final class Undefined extends com.autonomousapps.extension.Behavior {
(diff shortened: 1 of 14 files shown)


api/api.txt
```diff
@@ -4,6 +4,7 @@ package com.autonomousapps {
   public abstract class AbstractExtension {
     ctor @javax.inject.Inject public AbstractExtension(org.gradle.api.model.ObjectFactory objects, org.gradle.api.invocation.Gradle gradle);
     method public final org.gradle.api.file.RegularFileProperty adviceOutput();
+    method public final org.gradle.api.file.RegularFileProperty typeUsageOutput();
```
Laimiux: Please provide guidance here - I've added this, but I'm not sure what should actually be included here.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypeUsageProject.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.SourceType
+import com.autonomousapps.model.ProjectTypeUsage
+
+import static com.autonomousapps.kit.gradle.dependencies.Dependencies.*
+
+final class TypeUsageProject extends AbstractProject {
```
Laimiux, line 16: Added a functional test for this task (tried to follow project conventions)

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypeUsageWithFiltersProject.groovy
```diff
@@ -0,0 +1,80 @@
...
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.SourceType
+import com.autonomousapps.model.ProjectTypeUsage
+
+import static com.autonomousapps.kit.gradle.dependencies.Dependencies.*
+
+final class TypeUsageWithFiltersProject extends AbstractProject {
```
Laimiux, line 16: Another functional test for the new  task (tried to follow project conventions)

src/functionalTest/groovy/com/autonomousapps/jvm/TypeUsageSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.TypeUsageProject
+import com.autonomousapps.jvm.projects.TypeUsageWithFiltersProject
+
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertThat
+
+final class TypeUsageSpec extends AbstractJvmSpec {
```
Laimiux, line 12: Following project convention to test `computeTypeUsageMain` task

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+      "kotlin.jvm.internal",
+      "androidx.compose.runtime.internal",
+      "kotlin.jvm.functions",
+      "android"
+    ))
+
+  // Excluded specific types
+  internal val excludedTypes: SetProperty<String> = objects.setProperty(String::class.java)
+    .convention(setOf(
+      "kotlin.Unit",
```
Laimiux: Curious if we should provide certain defaults here for exclusion out of the report

src/main/kotlin/com/autonomousapps/internal/OutputPaths.kt
```diff
@@ -135,3 +136,4 @@ public fun getDuplicateDependenciesReport(): String = "$ROOT_DIR/duplicate-depen
 public fun getAllLibsVersionsTomlPath(): String = "$ROOT_DIR/allLibs.versions.toml"
 public fun getResolvedDependenciesReport(): String = "$ROOT_DIR/resolved-dependencies-report.txt"
 public fun getResolvedVersionsTomlPath(): String = "$ROOT_DIR/resolvedAllLibs.versions.toml"
+public fun getTypeUsagePath(variantName: String = "main"): String = "$ROOT_DIR/$variantName/intermediates/type-usage.json"
```
Laimiux: TODO: it should probably not live in `intermediates`

src/main/kotlin/com/autonomousapps/model/ProjectTypeUsage.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.squareup.moshi.JsonClass
+
+/**
+ * Type usage information for a single project.
+ *
+ * This report shows which types (classes/interfaces) are used from each dependency,
+ * enabling coupling analysis and complexity metrics.
+ */
+@JsonClass(generateAdapter = false)
+public data class ProjectTypeUsage(
```
Laimiux, line 14: Json serialized models for type usage task.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1081,6 +1080,21 @@ internal class ProjectPlugin(private val project: Project) {
       t.output.set(outputPaths.dependencyTraceReportPath)
     }
 
+    // Computes type-level usage statistics for complexity analysis.
+    val computeTypeUsageTask = tasks.register("computeTypeUsage$taskNameSuffix", ComputeTypeUsageTask::class.java) { t ->
```
Laimiux: New task registration

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2025. Tony Robalik.
```
Laimiux, line 1: Should we use 2025 or 2026?

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,202 @@
...
+) {
+  fun analyze(): ProjectTypeUsage {
+    val usageMap = mutableMapOf<String, MutableMap<String, Int>>()
+
+    // Get project's own class names for internal type detection
+    val projectClasses = project.classNames
+
+    // Count usage per class per dependency (both non-annotation and annotation classes)
+    project.codeSource.forEach { source ->
+      val allUsedClasses = source.usedNonAnnotationClasses + source.usedAnnotationClasses
```
Laimiux, line 158: Should we include `usedAnnotationClasses` ?

api/api.txt
autonomousapps: Whatever the abi generation task requires!

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypeUsageMultiModuleProject.groovy
```diff
@@ -0,0 +1,168 @@
...
+      """.stripIndent()
+    )
+  ]
+
+  ProjectTypeUsage actualTypeUsageFor(String projectPath) {
+    def typeUsage = gradleProject.singleArtifact(projectPath,
+      com.autonomousapps.internal.OutputPathsKt.getTypeUsagePath('main'))
+    def adapter = com.autonomousapps.internal.utils.MoshiUtils.MOSHI
+      .adapter(com.autonomousapps.model.ProjectTypeUsage)
+    return adapter.fromJson(typeUsage.asPath.text)
```
autonomousapps, line 167: Can we import these types?

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypeUsageProject.groovy
```diff
@@ -0,0 +1,72 @@
...
+    )
+  ]
+
+  ProjectTypeUsage actualTypeUsage() {
+    def typeUsage = gradleProject.singleArtifact(':proj',
+      com.autonomousapps.internal.OutputPathsKt.getTypeUsagePath('main'))
+    def adapter = com.autonomousapps.internal.utils.MoshiUtils.MOSHI
+      .adapter(com.autonomousapps.model.ProjectTypeUsage)
+    return adapter.fromJson(typeUsage.asPath.text)
+  }
```
autonomousapps, line 72: Same comment.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/TypeUsageWithFiltersProject.groovy
```diff
@@ -0,0 +1,80 @@
...
+    )
+  ]
+
+  ProjectTypeUsage actualTypeUsage() {
+    def typeUsage = gradleProject.singleArtifact(':proj',
+      com.autonomousapps.internal.OutputPathsKt.getTypeUsagePath('main'))
+    def adapter = com.autonomousapps.internal.utils.MoshiUtils.MOSHI
+      .adapter(com.autonomousapps.model.ProjectTypeUsage)
+    return adapter.fromJson(typeUsage.asPath.text)
+  }
```
autonomousapps, line 80: Same comment.

src/functionalTest/groovy/com/autonomousapps/jvm/TypeUsageSpec.groovy
```diff
@@ -0,0 +1,196 @@
...
+    def project = new TypeUsageProject()
+    gradleProject = project.gradleProject
+
+    when:
+    build(gradleVersion, gradleProject.rootDir, 'computeTypeUsageMain')
+
+    then: 'has correct summary'
+    def usage = project.actualTypeUsage()
+    assertThat(usage.projectPath).isEqualTo(':proj')
+    assertThat(usage.summary.totalTypes).isGreaterThan(0)
```
autonomousapps: why a fuzzy match here?

Also in general, I think it might be better to just create an instance of a type usage and check equality directly?

src/functionalTest/groovy/com/autonomousapps/jvm/TypeUsageSpec.groovy
autonomousapps, line 1: Solid specs! I'm curious why we can't make more straightforward equality checks.

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
autonomousapps: Don't worry about it. I can update that in a separate PR later. But thanks for pointing it out.

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+) {
+
+  // Excluded package prefixes
+  internal val excludedPackages: SetProperty<String> = objects.setProperty(String::class.java)
+    .convention(setOf(
+      "kotlin.jvm.internal",
+      "androidx.compose.runtime.internal",
+      "kotlin.jvm.functions",
+      "android"
+    ))
```
autonomousapps: This convention should be documented. I'd also like to know for my own benefit the rationale 😅 

Also, to answer your question below: I don't know! Should these be standard defaults that users _add to_, or should it be as you have it here, a convention that is overwritten? You're the author of the feature, and I don't really know the use-cases for it yet. Speaking of that, an update to the README would be super useful.

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Exclude types by package prefix.
+   *
+   * Example:
+   * ```
+   * excludePackages("com.example.internal", "kotlin.jvm.internal")
+   * ```
+   */
+  public fun excludePackages(vararg packages: String) {
+    excludedPackages.addAll(packages.toList())
```
autonomousapps, line 47: For this and the other functions, should we check that users provide at least one value?

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Exclude specific types by fully-qualified name.
+   *
+   * Example:
+   * ```
+   * excludeTypes("kotlin.Unit", "com.example.GeneratedClass")
+   * ```
+   */
+  public fun excludeTypes(vararg types: String) {
+    excludedTypes.addAll(types.toList())
```
autonomousapps, line 60: Are types a full match and packages a prefix match? For packages, do you add a `.` to the end of what the user supplies so that it really is treated like a package vs just a prefix?

src/main/kotlin/com/autonomousapps/internal/OutputPaths.kt
autonomousapps: Yes I agree.

src/main/kotlin/com/autonomousapps/model/ProjectTypeUsage.kt
```diff
@@ -0,0 +1,61 @@
...
+  val summary: TypeUsageSummary,
+
+  /** Types defined in this project that are used by itself. Map: className -> usageCount. */
+  val internal: Map<String, Int>,
+
+  /** Types used from project dependencies (other modules). Map: coordinates -> (className -> usageCount). */
+  val projectDependencies: Map<String, Map<String, Int>>,
+
+  /** Types used from external library dependencies. Map: coordinates -> (className -> usageCount). */
+  val libraryDependencies: Map<String, Map<String, Int>>,
```
autonomousapps: why use string for coordinates when Coordinates can be json-serialized?

src/main/kotlin/com/autonomousapps/model/ProjectTypeUsage.kt
```diff
@@ -0,0 +1,61 @@
...
+) : Comparable<ProjectTypeUsage> {
+
+  override fun compareTo(other: ProjectTypeUsage): Int = projectPath.compareTo(other.projectPath)
```
autonomousapps: this is a footgun. This compareTo _must_ be total. in fact if I had to guess I'd say this breaks equality checking in the Groovy Spock tests and that might be why you're not doing more straightforward equality checks?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1081,6 +1080,21 @@ internal class ProjectPlugin(private val project: Project) {
...
+      t.explodedJars.set(explodeJarTask.flatMap { it.output })
+
+      // Configuration from extension
+      t.excludedPackages.set(dagpExtension.typeUsageHandler.excludedPackages)
+      t.excludedTypes.set(dagpExtension.typeUsageHandler.excludedTypes)
+      t.excludedRegexPatterns.set(dagpExtension.typeUsageHandler.excludedRegexPatterns)
+
+      t.output.set(outputPaths.typeUsagePath)
+    }
+    storeTypeUsageOutput(computeTypeUsageTask.flatMap { it.output })
```
autonomousapps: This isn't quite right. Your task analyzes a specific variant or source set, meaning there are several tasks of the same type, so this function is basically called in a loop, with whatever is called last being the only one that gets stored. I think you want some kind of aggregation task that collects all variants, and you store _that_ here. There are several examples of such aggregation tasks.

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+      // 4. Write output
+      output.bufferWriteJson(typeUsage)
+    }
+
+    private fun buildClassIndex(explodedJarsFile: File): Map<String, String> {
+      if (!explodedJarsFile.exists()) return emptyMap()
+
+      val map = mutableMapOf<String, String>()
+
+      GZIPInputStream(explodedJarsFile.inputStream()).use { gzipStream ->
```
autonomousapps: You don't need this, there's a utility. Example:
```kotlin
val explodedJars = parameters.explodedJars.fromJsonSet<ExplodedJar>(compressed = true)
```

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+        val explodedJars = gzipStream.source().buffer().use { bufferedSource ->
+          getJsonListAdapter<ExplodedJar>().fromJson(bufferedSource)!!
+        }
+
+        explodedJars.forEach { jar ->
+          // Normalize identifier to handle included builds (convert "build:project" to ":project")
+          val identifier = jar.coordinates.normalizedIdentifier(":")
```
autonomousapps: Don't pass in a raw string. You need to use the actual `buildPath`. Example:
```kotlin
normalizedIdentifier(buildPath)
```

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+private class TypeFilter(
+  val excludedPackages: Set<String>,
+  val excludedTypes: Set<String>,
+  excludedRegexPatterns: List<String>
+) {
+  private val compiledPatterns = excludedRegexPatterns.map { it.toRegex() }
+
+  fun shouldExclude(className: String): Boolean {
+    if (excludedTypes.contains(className)) return true
+    if (excludedPackages.any { className.startsWith("$it.") }) return true
```
autonomousapps: Here we answer one of my questions from earlier about how the packages are treated. I wonder if it might be cleaner if, when accepting the user input, map it and affix a `.` to what they pass in. Also, we need to handle the case where a user might think they have to pass in a `.` themselves. I have some functions that do similar stuff:
```kotlin
internal fun String.ensurePrefix(prefix: String = ":"): String {
  return if (startsWith(prefix)) this else "$prefix$this"
}
```

That function is in another context. For your use-case, please add a similar function named `ensureSuffix` that ensures all packages end with exactly one `.`.

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
autonomousapps: Yeah I think so?

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    // Get project's own class names for internal type detection
+    val projectClasses = project.classNames
+
+    // Count usage per class per dependency (both non-annotation and annotation classes)
+    project.codeSource.forEach { source ->
+      val allUsedClasses = source.usedNonAnnotationClasses + source.usedAnnotationClasses
+
+      allUsedClasses.forEach { className ->
+        if (filter.shouldExclude(className)) return@forEach
```
autonomousapps: I generally prefer using `filter()` to this pattern.

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+
+        // Determine coordinates: check project classes first, then external
+        val coords = when {
+          projectClasses.contains(className) -> project.coordinates.identifier
+          else -> classToCoords[className] ?: "UNKNOWN"
```
autonomousapps: When would `UNKNOWN` happen? Would `!!` or `error("shouldn't happen")` be more appropriate?

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+private class TypeUsageAnalyzer(
+  private val project: ProjectVariant,
+  private val classToCoords: Map<String, String>,
+  private val filter: TypeFilter
+) {
+  fun analyze(): ProjectTypeUsage {
+    val usageMap = mutableMapOf<String, MutableMap<String, Int>>()
```
autonomousapps: You might be better off using `Coordinates` instead of `String` here for the key.

autonomousapps (changes requested):
Thanks for this! Left some comments.

api/api.txt
Laimiux: Is there a Gradle task to generate this?

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
Laimiux: To clarify, are you suggesting checking  that `vararg packages: String` contains at least one value?

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
Laimiux: I've seen cases with `R` classes. Instead of failure, I suggest exposing this via separate `unknownDependencies` property.

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
Laimiux: Good call

api/api.txt
autonomousapps: `./gradlew :updateApi`

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
autonomousapps: Yes.

api/api.txt
```diff
@@ -53,6 +53,13 @@ package com.autonomousapps {
     field public static final com.autonomousapps.Flags INSTANCE;
   }
 
+  @org.gradle.work.DisableCachingByDefault(because="Writes to console") public abstract class ListSourceFilesTask extends org.gradle.api.DefaultTask {
```
Laimiux, line 56: This is not related to this PR

src/functionalTest/groovy/com/autonomousapps/jvm/TypeUsageSpec.groovy
Laimiux: Fixed it

src/functionalTest/groovy/com/autonomousapps/jvm/TypeUsageSpec.groovy
Laimiux: Fixed

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
Laimiux: Good call, added `ensureSuffix`

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
Laimiux: The intention was to always exclude these rather than overwrite. However, for simplicity, I decided to remove these values for now.

src/main/kotlin/com/autonomousapps/extension/TypeUsageHandler.kt
Laimiux: Yeah, types are a full match and package is prefix match. The appending of `.` is now done via `ensureSuffix` that was discussed in another comment.

src/main/kotlin/com/autonomousapps/model/ProjectTypeUsage.kt
Laimiux: I could be wrong, but I think for map to serialize correctly, the key has to be a string value.

src/main/kotlin/com/autonomousapps/model/ProjectTypeUsage.kt
Laimiux: I removed the `compareTo` completely, this was not needed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
Laimiux: Good catch. For simplicity, I've updated this to use a `Map<Variant, RegularFile>` instead of single file. It probably makes sense to add an aggregate task, however, I wasn't sure about aggregate report given there will be a lot of overlap between debug and release.

src/main/kotlin/com/autonomousapps/tasks/ComputeTypeUsageTask.kt
```diff
@@ -0,0 +1,203 @@
...
+    return ProjectTypeUsage(
+      projectPath = project.coordinates.identifier,
+      summary = TypeUsageSummary(
+        totalTypes = usageMap.values.sumOf { it.size },
+        totalFiles = project.codeSource.size,
+        internalTypes = internal.size,
+        projectDependencies = projectDeps.size,
+        libraryDependencies = libraryDeps.size
+      ),
+      internal = internal,
```
autonomousapps: I think this should also be a sorted map, right?

autonomousapps (changes requested):
Sorry for the delay in re-reviewing this. I have one more substantive comment. If you address it, and then also update the api again, happy to approve and merge.

api/api.txt
autonomousapps: Understood. There's something broken in how this is working on CI, which I haven't made time to dig into yet.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1027,6 +1027,24 @@ internal class ProjectPlugin(private val project: Project) {
       duplicateClassesRuntime = duplicateClassesRuntime,
     )
 
+    // TODO(tsr): push this into DependencyAnalyzer similar to above
```
autonomousapps, line 1030: Follow-up.

autonomousapps (approved):

autonomousapps:
I pushed one commit to deeply sort the maps with map values. I'm very happy with this, thank you for all the work. Will merge when CI is green.

There is also one TODO that I added. I'll get around to it eventually but if you want to PR a small follow-up I would not complain.

autonomousapps:
This has now been published in [v3.13.0](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CHANGELOG.md#version-3130).

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
autonomousapps: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1711

PR #1642, opened 2026-02-19
feat: update console report to be more idiomatic for KMP.

autonomousapps:
New output for KMP projects:
```
  commonMain.dependencies {
    api(libs.okio) (was commonMainImplementation)
  }
  jvmMain.dependencies {
    api(project(":lib1")) (was jvmMainImplementation)
    api(project(":lib2")) (was jvmMainImplementation)
  }
```

### DIFF
--- a/api/api.txt
+++ b/api/api.txt
@@ -146,6 +146,8 @@ package com.autonomousapps.extension {
     method public final void ignoreGeneratedCode();
     method public final void ignoreInternalPackages();
     method public final void ignoreSubPackage(String packageFragment);
+    method public final void includeAnnotations(@org.intellij.lang.annotations.Language("RegExp") java.lang.String... annotationRegexes);
+    method public final void includeClasses(@org.intellij.lang.annotations.Language("RegExp") java.lang.String... classRegexes);
   }
 
   public final class Fail extends com.autonomousapps.extension.Behavior {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthConsoleReportBuilder.kt
```diff
@@ -216,6 +227,94 @@ internal class ProjectHealthConsoleReportBuilder(
...
+    // For unknown scopes
+    unknownScopes
+      .forEach { (_, advice) ->
+        advice.forEach { a ->
+          if (shouldPrintNewLine) {
+            builder.appendLine()
+          }
+          shouldPrintNewLine = true
+
+          val data = transform(a)
```
autonomousapps, line 310: There's some redundancy with this being called three times on the same data, but not sure it's worth worrying about. This is already very fast.

src/main/kotlin/com/autonomousapps/internal/artifacts/DagpArtifacts.kt
```diff
@@ -17,6 +17,7 @@ internal interface DagpArtifacts : Named {
   enum class Kind : ArtifactDescription<DagpArtifacts> {
     COMBINED_GRAPH,
     PROJECT_HEALTH,
+    PROJECT_METADATA,
```
autonomousapps, line 20: Adding a new artifact meant we don't have to modify `ProjectAdvice`, which means we're not breaking ABI compatibility.

api/api.txt
```diff
@@ -557,6 +561,28 @@ package com.autonomousapps.model.source {
...
+    method public String getName();
+    method public String getRuntimeClasspathName();
+    property public String compileClasspathName;
+    property public String kind;
+    property public String name;
+    property public String runtimeClasspathName;
+    field public static final String COMMON_MAIN_NAME = "commonMain";
+    field public static final String COMMON_TEST_NAME = "commonTest";
+    field public static final String JVM_MAIN_NAME = "jvmMain";
+    field public static final String JVM_TEST_NAME = "jvmTest";
```
autonomousapps, line 583: These are all in an `internal companion object`, so I'm not sure why they're showing up in the published api.

PR #1645, opened 2026-02-25
feat: support KMP projects with android targets.

autonomousapps:
This support is preliminary and likely is missing features.


src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -704,7 +735,10 @@ internal class ProjectPlugin(private val project: Project) {
 //    }
 //  }
```
autonomousapps: Cleanup

PR #1651, opened 2026-03-01
fix: enable build cache reuse across machines

### LINKED ISSUE #1083: Cache misses between CI and local
**Build scan link**

**Plugin version**
1.28.0

**Gradle version**
8,5

**JDK version**
17

**(Optional) Kotlin and Kotlin Gradle Plugin (KGP) version**
1.9.21

N/a

**Describe the bug**
The buildHealth task and all its deps are not being cached properly and we get cache misses more often than ideal.
The cause seems to be not using PathSenstivity.Relative in all cases. The reason we get cache misses is because the absolute paths of input dependencies of many of the tasks created by the plugin use absolute or none path sensitivity. See image for one such example

image

**To Reproduce**
1. ...

**Expected behavior**
If the code hasn't change I expect cache hits for all the tasks that this plugin creates in CI.

**Additional context**

NikolayMetchev:
## Summary

Fixes #1083

- Change `PathSensitivity.ABSOLUTE` to `PathSensitivity.NONE` on `ArtifactsReportTask.getClasspathArtifactFiles()`, so the build cache key is based on file content rather than filesystem path. This enables cache hits between CI and local builds (and across any machines with different Gradle home locations).
- Remove `ExplodedJar.jarFile` — set from `PhysicalArtifact.file` but never accessed after serialization. Eliminating it removes absolute paths from `ExplodeJarTask`'s output.
- Remove `Dependency.files` — explicitly marked in code as "unused and was only added speculatively". Eliminating it removes absolute paths from `SynthesizeDependenciesTask`'s output.

### Why this is safe

When `ArtifactsReportTask` gets a cache hit, its output (PhysicalArtifact JSON) is restored verbatim. All downstream tasks (`ExplodeJarTask`, `FindKotlinMagicTask`, `ComputeDominatorTreeTask`, `SynthesizeDependenciesTask`) see identical input content, so they also get cache hits. Nobody tries to use the absolute paths at runtime. The entire chain caches together coherently.

### Why `NONE` and not `RELATIVE`?

`PathSensitivity.NONE` means Gradle ignores the file's path entirely and only fingerprints its content. This is appropriate here because jars live in the Gradle user home (outside the project directory), so `RELATIVE` would not help.

## Test plan

- [x] `./gradlew compileKotlin` passes
- [ ] Full test suite passes
- [ ] Verify intermediate JSON outputs no longer contain `jarFile` or `files` fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/kotlin/com/autonomousapps/model/internal/intermediates/producer/ExplodedJar.kt
+++ b/src/main/kotlin/com/autonomousapps/model/internal/intermediates/producer/ExplodedJar.kt
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.autonomousapps.model.internal.intermediates.producer
 
+import com.autonomousapps.internal.utils.LexicographicIterableComparator
 import com.autonomousapps.internal.utils.ifNotEmpty
 import com.autonomousapps.model.Coordinates
 import com.autonomousapps.model.internal.*
@@ -67,7 +68,9 @@ internal data class ExplodedJar(
   )
 
   override fun compareTo(other: ExplodedJar): Int {
-    return coordinates.compareTo(other.coordinates)
+    return compareBy(ExplodedJar::coordinates)
+      .thenBy(LexicographicIterableComparator()) { it.binaryClasses }
+      .compare(this, other)
   }
 
   init {


autonomousapps:
As noted elsewhere, I have recently added a [code of conduct](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CODE_OF_CONDUCT.md), and now LLM-generated content is strictly forbidden for both issues and PRs.

This PR is superseded by https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1682. It leaves in place the `@PathSensitive(PathSensitivity.ABSOLUTE)` task input, and it was a mistake to remove it in the first place. It does however reduce the blast radius of the absolute file paths.

Issue #1653, opened 2026-03-05
NoSuchElementException on some android projects 


matejdro:
**Plugin version**
3.6.0+. 

3.5.1 works fine

**Gradle version**
9.4.0

**JDK version**
11

**(Optional) Kotlin version**
3.4.0

**(Optional) Android Gradle Plugin (AGP) version**
8.12.0

**Describe the bug**
When this plugin is used on a project with [Metro](https://github.com/ZacSweers/metro) `0.10.2`, it will crash with the following exception:

```
Caused by: java.util.NoSuchElementException
        at com.autonomousapps.internal.Bundles.maybeParent(Bundles.kt:320)
        at com.autonomousapps.tasks.DependencyAdviceBuilder.computeDependencyAdvice$lambda$3(ComputeAdviceTask.kt:348)
        at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:243)
        at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:203)
        at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:227)
        at kotlin.sequences.FlatteningSequence$iterator$1.ensureItemIterator(Sequences.kt:356)
        at kotlin.sequences.FlatteningSequence$iterator$1.hasNext(Sequences.kt:351)
        at kotlin.sequences.SequencesKt___SequencesKt.toCollection(_Sequences.kt:806)
        at kotlin.sequences.SequencesKt___SequencesJvmKt.toSortedSet(_SequencesJvm.kt:49)
        at com.autonomousapps.tasks.DependencyAdviceBuilder.<init>(ComputeAdviceTask.kt:295)
        at com.autonomousapps.tasks.ComputeAdviceTask$ComputeAdviceAction.execute(ComputeAdviceTask.kt:198)
```

**To Reproduce**
1. Download [DependencyAnalysisTestAndroid.zip](https://github.com/user-attachments/files/25769332/DependencyAnalysisTestAndroid.zip)
2. Run `./gradlew :app:computeAdvice`

**Expected behavior**
Analysis should not crash

From what I can tell, updating metro to a newer version fixes this. If you think that's enough, feel free to close the issue. But I thought reporting would still be beneficial in case this is some underlying bug that would be good to be solved anyway.

StefMa:
Yep, same here for `3.6.1`. 
`3.5.1` works fine.

[...]

PR #1660, opened 2026-03-27
Sarif reporting

### LINKED ISSUE #1094: Generate output for `project`/`build` health in `errorformat`
New output format for `projectHealth` and `buildHealth`.

In order to support features like [Github Checks](https://docs.github.com/en/rest/checks?apiVersion=2022-11-28), we need the ability to see line-numbers and other metadata associated with files specifically. 

In order to reduce the feature request, if we generated `errorformat` outputs, we could hook into tools like [@reviewdog/reviewdog](https://github.com/reviewdog/reviewdog). 

I wrote up a github action with a python script to parse the `buildHealth.txt` file for outputs: https://github.com/Nava2/dependency-analysis-reviewdog-action. This works OK, but it is brittle and prone to errors as the formats change.

matejdro:
PR adds an ability for the plugin to also export the report in the standard `.sarif` format, in addition to the txt.

Would appreciate feedback on some things:

* For the best experience, issues in sarif file should have line numbers attached, but no such information exists in the DAGP. I have attempted to infer those line numbers by reading through the `build.gradle(.kts)` file and just string matching of the advice's coordinates. Is that okay? Could we come up with a better way to do this?
* I have only enabled this in the `buildHealth` task, but not the `projectHealth`. The issue is that individual projects do not have access to the reporting config and thus could not enable this. If necessary, I could add `public fun reporting(action: Action<ReportingHandler>)` to the `DependencyAnalysisSubExtension` and add sarif output to the `projectHealth`
* Any hints on how can I add tests for this? From what I saw, there are no existing tests checking report output, so I don't have any template to follow.

(This PR is based on the `3.5.1` due to #1653, as it makes testing easier for me. When the code is final, I can resolve conflicts and merge latest version in)

this fixes #1094


src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthSarifReportBuilder.kt
```diff
@@ -0,0 +1,176 @@
...
+          tool = SARIF_TOOL,
+        )
+      )
+    )
+  }
+}
+
+private val SARIF_TOOL = Tool(
+  driver = ToolComponent(
+    guid = "f9137358-f4fb-44f2-8300-39ca0b85fb77",
```
autonomousapps, line 118: What is the `guid`?

src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
```diff
@@ -418,7 +417,8 @@ internal class StandardTransform(
           advice += Advice.ofChange(
             coordinates = theRemove.coordinates,
             fromConfiguration = theRemove.fromConfiguration!!,
-            toConfiguration = theAdd.toConfiguration!!
+            toConfiguration = theAdd.toConfiguration!!,
+            declarationLineNumber = theRemove.buildFileDeclarationLineNumber ?: theAdd.buildFileDeclarationLineNumber
```
autonomousapps: `theAdd` won't have a current declaration. The only things that would have a current declaration relate to remove advice and _change_ advice.

autonomousapps (commented):
Thanks for this PR. I have not done a thorough review because I don't have time at the moment, but I'm very open to seeing this progress and eventually get merged.

We ought to be able to resolve the issue with projectHealth not having access to the reporting handler. I think for this to be considered feature-complete, it would have to work for both buildHealth and projectHealth.

The build file scanning may have some issues, but I haven't thoroughly inspected it. As you know, the plugin's analysis uses Gradle's APIs to find dependencies—it doesn't actually scan build scripts unless it's trying to rewrite them. So that'll have to be done on a best-effort basis (which is what I think you're doing).

src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthSarifReportBuilder.kt
matejdro: From what I understand, it's just a static unique identifiable ID for every tool. I just generated a random UUID and plopped it there.

src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
matejdro: Ah, you are right. Do I remove this? I have added it to all types just for consistency's sake.

matejdro:
So you would want to generate yet another json next to the regular advice json just for an extra property and then consume that json in addition to the regular json? Sounds a pretty complicated thing to do.

Or did I miss the intention here?

autonomousapps:
> So you would want to generate yet another json next to the regular advice json just for an extra property and then consume that json in addition to the regular json? Sounds a pretty complicated thing to do.
> 
> Or did I miss the intention here?

Yes, that's correct. And also that feature should be opt-in, based on user preference (as expressed in the `reporting` DSL).

In my opinion, this is actually easier than what you've spiked here. json is cheap.

matejdro:
Updated. now dependency filtering task will also emit `final-sourced-advice.json` after filtering. If this is okay, I can continue with the implementation of the per-project report.

There is also an issue of reporting configuration: `reportingHandler` is a singleton in the `GlobalDslService`, which means that, currently, projects cannot each set their own reporting configuration (changing it in one project would override the config in all of them). Are there any negative side effects if I make this not-singleton?

autonomousapps:
> There is also an issue of reporting configuration: `reportingHandler` is a singleton in the `GlobalDslService`, which means that, currently, projects cannot each set their own reporting configuration (changing it in one project would override the config in all of them). Are there any negative side effects if I make this not-singleton?

I think it should remain a singleton. I don't see the benefit of allowing per-project variation here. Do you?

matejdro:
Yeah, then I should revert the reporting handler being accessible from project - it would only be configurable from the root project.

This is an unusual UX for the Gradle project, though. It is generally expected, that every project's configuration is only affecting that project, not anything else (especially after Isolated Projects). Configuration for root project affecting other project is a bit unexpected. But I guess DAGP has already set a precedent for this with bundles being set only on root, but applying everywhere.

src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthSarifReportBuilder.kt
```diff
@@ -0,0 +1,179 @@
...
+  projectAdvices: Collection<SourcedProjectAdvice>,
+  dslKind: DslKind,
+  /** Customize how dependencies are printed. */
+  dependencyMap: ((String) -> String?)? = null,
+  useTypesafeProjectAccessors: Boolean,
+) {
+
+  val sarif: SarifSchema210
+
+  private val advicePrinter = AdvicePrinter(dslKind, ProjectType.JVM, dependencyMap, useTypesafeProjectAccessors)
```
matejdro, line 19: ProjectType appears to be unused in AdvicePrinter? So I just hardcoded it to the JVM for now. Is a future use expected?

matejdro:
Updated to the latest main, reverted per-project sarif settings and added sarif reporting to the per-project tasks. I think it's ready for the next review now.

src/main/kotlin/com/autonomousapps/internal/advice/ProjectHealthSarifReportBuilder.kt
autonomousapps: Huh. I added that when I added the first burst of KMP support. I can't say now whether not using it is an oversight, or leaving it in as an unused constructor is the oversight.

src/main/kotlin/com/autonomousapps/internal/artifacts/DagpArtifacts.kt
```diff
@@ -18,6 +18,7 @@ internal interface DagpArtifacts : Named {
     COMBINED_GRAPH,
     PUBLIC_CLASSES,
     PROJECT_HEALTH,
+    SOURCED_PROJECT_HEALTH,
```
autonomousapps: Could you order this alphabetically?

src/main/kotlin/com/autonomousapps/model/Advice.kt
```diff
@@ -32,32 +32,45 @@ public data class Advice(
 
   public companion object {
     @JvmStatic
-    public fun ofAdd(coordinates: Coordinates, toConfiguration: String): Advice = Advice(
+    public fun ofAdd(
+      coordinates: Coordinates,
+      toConfiguration: String,
+      declarationLineNumber: Int? = null
```
autonomousapps: vestige from an earlier version of this PR, I think?

src/main/kotlin/com/autonomousapps/model/Advice.kt
autonomousapps, line 1: I think probably all the changes in this file could be reverted.

src/main/kotlin/com/autonomousapps/model/SourcedAdvice.kt
```diff
@@ -0,0 +1,16 @@
...
+
+
+/**
+ * Wrapper for advice class with source data attached
+ * (e.g. in which line of the build file does the source of this advice originate)
+ */
+@JsonClass(generateAdapter = false)
+public data class SourcedAdvice(
+  val advice: Advice,
+  val buildFileDeclarationLineNumber: Int? = null,
```
autonomousapps, line 15: Once we're creating an instance of this type, can the line number be null?

src/main/kotlin/com/autonomousapps/model/SourcedProjectAdvice.kt
```diff
@@ -0,0 +1,17 @@
...
+@JsonClass(generateAdapter = false)
+public data class SourcedProjectAdvice(
+  val projectPath: String,
+  val dependencyAdvice: Set<SourcedAdvice> = emptySet(),
+  val pluginAdvice: Set<PluginAdvice> = emptySet(),
+  val moduleAdvice: Set<ModuleAdvice> = emptySet(),
+  val warning: Warning = Warning.empty(),
+  /** True if there is any advice in a category for which the user has declared they want the build to fail. */
+  val shouldFail: Boolean = false,
+  val projectBuildFile: String? = null,
```
autonomousapps: should this be nullable?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -128,6 +129,10 @@ internal class ProjectPlugin(private val project: Project) {
     project = project,
     artifactDescription = DagpArtifacts.Kind.PROJECT_HEALTH,
   )
+  private val sourcedProjectHealthPublisher = interProjectPublisher(
```
autonomousapps: please sort alphabetically with the other publishers

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -891,9 +896,9 @@ internal class ProjectPlugin(private val project: Project) {
     }
 
     dependencyAnalyzer.registerDominatorTreeTasks(
-      artifactsReportCompile = artifactsReport,
-      artifactsReportRuntime = artifactsReportRuntime,
-      graphViewTask = graphViewTask,
+            artifactsReportCompile = artifactsReport,
+            artifactsReportRuntime = artifactsReportRuntime,
+            graphViewTask = graphViewTask,
```
autonomousapps: ?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1002,16 +1007,16 @@ internal class ProjectPlugin(private val project: Project) {
...
+        graphViewTask,
+        declaredProcsTask,
+        explodeBytecodeTask,
+        explodeCodeSourceTask,
+        usagesExclusionsProvider,
+        artifactsReport,
+        abiAnalysisTask,
+        explodeXmlSourceTask,
+        explodeAssetSourceTask,
+      )
```
autonomousapps: ?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1032,7 +1037,7 @@ internal class ProjectPlugin(private val project: Project) {
     val computeUsagesTask = dependencyAnalyzer.registerComputeUsagesTask(
       checkSuperClasses = dagpExtension.usageHandler.analysisHandler.checkSuperClasses,
       checkBinaryCompat = checkBinaryCompat(),
-      isKaptApplied = isKaptApplied().orElse(isLegacyKaptApplied()),
+isKaptApplied = isKaptApplied().orElse(isLegacyKaptApplied()),
```
autonomousapps: ?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1067,6 +1072,9 @@ internal class ProjectPlugin(private val project: Project) {
       androidScoreTask?.let { a -> t.androidScoreReports.add(a.flatMap { it.output }) }
     }
     filterAdviceTask.configure { t ->
+      if (dagpExtension.reportingHandler.sarifReport.get()) {
+        t.buildFile.set(project.buildFile)
```
autonomousapps: I think this could be set unconditionally.

I haven't checked, but this `buildFile` property should be `@Internal` as well. I don't think we want Gradle wasting time hashing its contents.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1143,24 +1151,46 @@ internal class ProjectPlugin(private val project: Project) {
...
       t.output.set(paths.filteredAdvicePath)
+      t.sourcedOutput.set(
+        dagpExtension.reportingHandler.sarifReport.flatMap { enableSarifReport ->
+          if (enableSarifReport) {
+            paths.filteredSourcedAdvicePath
+          } else {
+            provider<RegularFile> { null }
+          }
+        }
+      )
```
autonomousapps: I don't think I like this. I think let's just set the property unconditionally.

You might just want a `Property<Boolean>` that triggers sarif report behavior in the task.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1143,24 +1151,46 @@ internal class ProjectPlugin(private val project: Project) {
...
+      t.dependencyMap.set(
+        objects.newInstance(DependenciesHandler::class.java).apply {
+          withVersionCatalogs(project)
+        }.map
+      )
```
autonomousapps: I don't understand this. Elsewhere in this same file there's this line:
```kotlin
t.dependencyMap.set(dagpExtension.dependenciesHandler.map)
```
why is that not sufficient for this case?

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1143,24 +1151,46 @@ internal class ProjectPlugin(private val project: Project) {
...
-      t.projectPath.set(theProjectPath)
-      t.projectType.set(projectType)
-      t.output.set(paths.projectMetadataPath)
-    }
+    val writeProjectMetadata = tasks.register("writeProjectMetadata", WriteProjectMetadataTask::class.java) {
+        t ->
+        t.projectPath.set(theProjectPath)
+        t.projectType.set(projectType)
+        t.output.set(paths.projectMetadataPath)
+      }
```
autonomousapps: Please revert all the indentation changes in this file.

src/main/kotlin/com/autonomousapps/internal/utils/utils.kt
autonomousapps, line 1: This change can be reverted, per other comments.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1143,24 +1151,46 @@ internal class ProjectPlugin(private val project: Project) {
...
         t.dslKind.set(DslKind.from(buildFile))
         t.dependencyMap.set(dagpExtension.dependenciesHandler.map)
         t.useTypesafeProjectAccessors.set(dagpExtension.useTypesafeProjectAccessors)
         t.useParenthesesForGroovy.set(dagpExtension.dependenciesHandler.useParenthesesForGroovy)
         t.output.set(paths.consoleReportPath)
+        t.sarifOutput.set(
+          dagpExtension.reportingHandler.sarifReport.flatMap { sarifReportEnabled ->
+            if (sarifReportEnabled) paths.sarifReportPath else null
+          }
+        )
```
autonomousapps: Set the output unconditionally.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
```diff
@@ -1211,6 +1241,9 @@ internal class ProjectPlugin(private val project: Project) {
     // Publish our artifacts
     combinedGraphPublisher.publish(mergeProjectGraphsTask.flatMap { it.output })
     projectHealthPublisher.publish(filterAdviceTask.flatMap { it.output })
+    if (dagpExtension.reportingHandler.sarifReport.get()) {
+      sourcedProjectHealthPublisher.publish(filterAdviceTask.flatMap { it.sourcedOutput })
+    }
```
autonomousapps: Unconditional.

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
```diff
@@ -44,6 +44,10 @@ internal class RootPlugin(private val project: Project) {
     project = project,
     artifactDescription = DagpArtifacts.Kind.PROJECT_HEALTH,
   )
+  private val sourcedAdviceResolver = interProjectResolver(
```
autonomousapps: alphabetical

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
```diff
@@ -131,6 +136,11 @@ internal class RootPlugin(private val project: Project) {
       t.output.set(paths.buildHealthPath)
       t.consoleOutput.set(paths.consoleReportPath)
       t.outputFail.set(paths.shouldFailPath)
+      t.sarifOutput.set(
+        dagpExtension.reportingHandler.sarifReport.flatMap { sarifReportEnabled ->
+          if (sarifReportEnabled) paths.sarifReportPath else null
+        }
+      )
```
autonomousapps: Unconditional

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
```diff
@@ -153,7 +163,10 @@ internal class RootPlugin(private val project: Project) {
       t.consoleReport.set(generatePublicTypeUsages.flatMap { it.outputConsole })
     }
 
-    tasks.register("generateWorkPlan", GenerateWorkPlan::class.java) { t ->
+    tasks.register(
+      "generateWorkP      it.sarifReport.set(generateBuildHealthTask.flatMap { it.sarifOutput })\nlan",
+      GenerateWorkPlan::class.java
+    ) { t ->
```
autonomousapps: ??

autonomousapps (changes requested):
Didn't finish review, left a bunch of comments. Please double-check all your changes and let me know when you're resolved all the weird formatting inconsistencies and breakages.

src/main/kotlin/com/autonomousapps/model/SourcedAdvice.kt
matejdro: Yes, because line number is a guess at best. Sometimes, we will not know the line number at all (if nothing matches in the build file). In this case, it is null.

src/main/kotlin/com/autonomousapps/model/Advice.kt
matejdro: Reverted

src/main/kotlin/com/autonomousapps/model/Advice.kt
matejdro: Reverted

src/main/kotlin/com/autonomousapps/model/SourcedProjectAdvice.kt
matejdro: You are right. It was nullable in the previous version, but in the new separate-class version it's always filled.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Oops, fixed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Oops, fixed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Oops, fixed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Changed to unconditional.

The issue with marking this as `@Internal` is that changes from the build file will not re-run the advice task.

For example, if you reorder the dependencies in the build file, Gradle will not re-run the build, because output of those dependencies is the same. But we still need re-run, because the line numbers have now changed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: ~~I think this one needs to stay conditional. Publish output always needs to be present, otherwise the build will fail. And, when sarif report is disabled, `sourcedOutput` would not return any outputs. So, we must disable publishing in this case, to avoid the empty publish scenario.~~

Nevermind, we can just not-read the resolver side. Removed the condition.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Updated all tasks to gate the sarif exporting via a separate boolean instead of a missing file property

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Ah, I missed this easier way. Changed.

src/main/kotlin/com/autonomousapps/internal/utils/utils.kt
matejdro: Done

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Oops, fixed.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Sorted

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
matejdro: Done

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
matejdro: Done

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
matejdro: Looks like a copy paste error. Reverted.

src/main/kotlin/com/autonomousapps/internal/artifacts/DagpArtifacts.kt
matejdro: Done. I've also moved existing `PUBLIC_CLASSES` to comform with the alphabetic sorting.

src/main/kotlin/com/autonomousapps/subplugin/RootPlugin.kt
matejdro: Sorted

matejdro:
I think that should be all weird indentations fixed. Sorry about that.

src/main/kotlin/com/autonomousapps/subplugin/ProjectPlugin.kt
autonomousapps: > The issue with marking this as @Internal

Good point!

autonomousapps:
> I think that should be all weird indentations fixed. Sorry about that.

Thanks for addressing! I'll continue my review soon.

PR #1669, opened 2026-04-07
fix: don't attempt to suggest upgrading an undeclared 'parent' dependency of a KMP dependency.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1653

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/Bundles.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/Bundles.kt
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.autonomousapps.internal
 
-import com.autonomousapps.model.internal.ProjectType
 import com.autonomousapps.extension.DependenciesHandler.SerializableBundles
 import com.autonomousapps.graph.Graphs.children
 import com.autonomousapps.graph.Graphs.reachableNodes
@@ -10,6 +9,7 @@ import com.autonomousapps.internal.utils.filterToOrderedSet
 import com.autonomousapps.model.*
 import com.autonomousapps.model.Coordinates.Companion.copy
 import com.autonomousapps.model.internal.DependencyGraphView
+import com.autonomousapps.model.internal.ProjectType
 import com.autonomousapps.model.internal.declaration.Bucket
 import com.autonomousapps.model.internal.declaration.ConfigurationNames
 import com.autonomousapps.model.internal.declaration.Declaration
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -359,12 +359,12 @@ internal class DependencyAdviceBuilder(
           //  `parentAdvice` case.
           // This can transform add-advice to change-advice. Currently only for KMP projects where the "parent" KMP dep
           // is declared on a commonX configuration, and the "child" -jvm or -android dep needs to be upgraded.
-          advice.isAdd() && bundles.hasParentInBundle(originalCoordinates) -> {
+          advice.isAdd() && advice.isToApiLike() && bundles.hasParentInBundle(originalCoordinates) -> {
```
autonomousapps: This is the change that resolves the specific issue.

src/main/kotlin/com/autonomousapps/internal/Bundles.kt
```diff
@@ -92,6 +93,13 @@ internal class Bundles private constructor(
     // Find all declarations for this dependency. E.g., ["commonMainImplementation", "jvmMainImplementation"].
     val parentDeclarations = declarations.filterToOrderedSet { decl -> decl.identifier == parentCoordinates.identifier }
 
+    // This can happen when the parent dependency is added by a plugin (no declarations). We can't do anything here.
+    // See https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1653.
+    if (parentDeclarations.isEmpty()) {
+      // nb: the only caller of this method (at time of writing!) will throw away the advice if it is unchanged.
+      return addAdvice
+    }
+
```
autonomousapps, line 101: This is the fix.

src/main/kotlin/com/autonomousapps/tasks/ComputeAdviceTask.kt
```diff
@@ -359,12 +359,12 @@ internal class DependencyAdviceBuilder(
...
             val parent = bundles.findParentInBundle(originalCoordinates)!!
-            bundledTraces.add(BundleTrace.DeclaredParent(parent = parent, child = originalCoordinates))
 
             val parentAdvice = bundles.maybeParent(advice, originalCoordinates)
             if (parentAdvice != advice) {
+              bundledTraces.add(BundleTrace.DeclaredParent(parent = parent, child = originalCoordinates))
```
autonomousapps, line 365: We should only add a trace when we're making a change.

PR #1681, opened 2026-04-18
feat: support analysis of a Gradle version catalog dependency.

autonomousapps:
A declaration on that 'dependency' typically looks like this:
```
// kotlin DSL
// https://github.com/gradle/gradle/issues/15383
implementation(files(libs::class.java.superclass.protectionDomain.codeSource.location))
```

With this infrastructure in place, it may be possible to support things like 'gradleApi()' as well, in the future.

### DIFF
--- a/testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Dependency.kt
+++ b/testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Dependency.kt
@@ -14,6 +14,7 @@ public data class Dependency @JvmOverloads constructor(
   private val ext: String? = null,
   private val capability: String? = null,
   private val isVersionCatalog: Boolean = false,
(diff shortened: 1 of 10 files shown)


testkit/gradle-testkit-support/src/main/kotlin/com/autonomousapps/kit/gradle/Dependency.kt
```diff
@@ -7,6 +7,8 @@ import com.autonomousapps.kit.internal.ensurePrefix
 import com.autonomousapps.kit.render.Element
 import com.autonomousapps.kit.render.Scribe
 
+// TODO(tsr): need some way to represent raw or flat coordinates, which we don't want to wrap in quotation marks. E.g.,
+//  implementation(files(libs::class.java.superclass.protectionDomain.codeSource.location))
```
autonomousapps: Followup.

PR #1706, opened 2026-05-19
fix: don't suggest non-existent Android unit test variant in advice.

autonomousapps:
Instead of 'debugTestImplementation', suggest 'testImplementation'.

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
@@ -80,14 +80,14 @@ internal class BaseConventionPlugin(private val project: Project) {
       .map { it.toBoolean() }
 
     // Both these values are static and safe to get during configuration
-    val shouldSign = !isCi.get() && !isSnapshot.get()
+//    val shouldSign = !isCi.get() && !isSnapshot.get()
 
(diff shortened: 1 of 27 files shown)


build-logic/convention/src/main/kotlin/com/autonomousapps/convention/BaseConventionPlugin.kt
autonomousapps, line 1: Cleanup.

src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
```diff
@@ -196,8 +196,14 @@ internal class StandardTransform(
...
 
-  private fun MutableSet<Usage>.simplify(visibility: Bucket.Visibility, sourceSetName: String): MutableSet<Usage> {
-    return if (visibility.forCompile && !explicitFor(sourceSetName)) {
+  private fun MutableSet<Usage>.simplify(
+    visibility: Bucket.Visibility,
+    sourceSetName: String,
+  ): MutableSet<Usage> {
+    return if (projectType == ProjectType.ANDROID && sourceSetName == SourceKind.TEST_NAME) {
+      // Android unit test configuration is complicated, so just bail out I guess?
+      reduceUsages(this)
```
autonomousapps: The fix.

PR #1708, opened 2026-05-20
fix: exceptions, main->test visibility, reason for android lint jars.

autonomousapps:
Three separate bugfixes:
1. `reason` output said "implies implementation" when it should have been "implies runtimeOnly" in the android lint jar case.
2. The new "exceptions are special" handling was over-eager and flagging things that only _referenced_ exceptions. We only want to flag things that _provide_ the exception types.
3. `MutableSet<Usage>.simplify(visibility, bucket)` in `StandardTransform` wasn't properly handling the case when a usage was visible on _both_ compile and runtime classpaths. In such a case, we can filter out _all_ usages.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
@@ -209,22 +209,29 @@ internal class StandardTransform(
     sourceSetName: String,
   ): MutableSet<Usage> {
     fun MutableSet<Usage>.maybeReduceFurther(): MutableSet<Usage> {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/autonomousapps/internal/transform/StandardTransform.kt
autonomousapps, line 1: Cleanup

src/main/kotlin/com/autonomousapps/model/internal/intermediates/Reason.kt
autonomousapps, line 1: Separate bug discovered when using `reason`.

PR #1717, opened 2026-05-28
fix: don't report Android R classes as classpath duplicates

### LINKED ISSUE #1335: Duplicate class detection warning when using strings from one module in another
I have a library module that has a `strings.xml` file. I reference the generated ids from it (e.g. `template.destinations.welcome.R.strings.*`) from the `androidTest` source set in my app module.

When I run `buildHealth` the following is printed:

```
Advice for :app
Warnings
Some of your classpaths have duplicate classes, which means the compile and runtime behavior can be sensitive to the classpath order.

Source set: devDebug
\--- compile classpath
     +--- template/destinations/welcome/R is provided by multiple dependencies: [:destinations:welcome, :app]
     \--- template/destinations/welcome/R$string is provided by multiple dependencies: [:destinations:welcome, :app]

```

Not sure if this is a bug or WAI (I have a repro I can share if it is a bug).

rvandermeulen:
Fixes #1335.

Android generates an `R` class (and nested `R$*` classes) into every dependency that ships resources, as well as the consuming module, so the same `R` legitimately appears on the classpath of multiple artifacts and gets reported as a duplicate even though it's never user-actionable. This skips `R`/`R$*` class files in `DiscoverClasspathDuplicationTask`.

The filter is name-based and unconditional rather than gated on the Android plugin, since `ProjectVariant` has no clean "is Android" signal at this point. Happy to gate it on Android if you'd prefer. I scoped it to `R`/`R$*` only and didn't touch the CHANGELOG, but can add an entry if you'd like.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/DiscoverClasspathDuplicationTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/DiscoverClasspathDuplicationTask.kt
@@ -7,6 +7,7 @@ import com.autonomousapps.internal.utils.*
 import com.autonomousapps.model.Coordinates
 import com.autonomousapps.model.DuplicateClass
 import com.autonomousapps.model.internal.ProjectVariant
+import com.autonomousapps.model.source.AndroidSourceKind
 import org.gradle.api.DefaultTask
 import org.gradle.api.artifacts.ArtifactCollection
 import org.gradle.api.artifacts.result.ResolvedArtifactResult
@@ -75,9 +76,12 @@ public abstract class DiscoverClasspathDuplicationTask : DefaultTask() {
     }
 
     fun duplicates(): Set<DuplicateClass> {
+      val isAndroid = project.sourceKind is AndroidSourceKind
       return duplicatesMap.asMap()
         // Find all class files provided by more than one dependency
         .filterValues { it.size > 1 }
+        // R classes are duplicated by design; ignore them on Android.
+        .filterKeys { !isAndroid || !isAndroidRClass(it) }
         // only warn about duplicates if it's about a class that's actually used.
         .filterKeys {
           val fqcn = it.replace('/', '.').removeSuffix(".class")
@@ -111,4 +115,13 @@ public abstract class DiscoverClasspathDuplicationTask : DefaultTask() {
         .forEach { duplicatesMap.put(it, coordinates) }
     }
   }
+}
+
+/**
+ * Returns true if [classFile] (a `/`-delimited class file name, e.g. `androidx/appcompat/R$id.class`) is an Android
+ * `R` class. Android generates `R` and nested `R$*` classes into every module, so they're never actionable duplicates.
+ */
+internal fun isAndroidRClass(classFile: String): Boolean {
+  val simpleName = classFile.removeSuffix(".class").substringAfterLast('/')
+  return simpleName == "R" || simpleName.startsWith("R$")
 }
--- a/src/test/kotlin/com/autonomousapps/tasks/DiscoverClasspathDuplicationTaskTest.kt
+++ b/src/test/kotlin/com/autonomousapps/tasks/DiscoverClasspathDuplicationTaskTest.kt
@@ -1,0 +1,42 @@
+// Copyright (c) 2026. Tony Robalik.
+// SPDX-License-Identifier: Apache-2.0
+package com.autonomousapps.tasks
+
+import com.google.common.truth.Truth.assertWithMessage
+import org.junit.jupiter.api.Test
+
+internal class DiscoverClasspathDuplicationTaskTest {
+
+  @Test fun `recognizes Android R classes`() {
+    val rClasses = listOf(
+      "androidx/appcompat/R.class",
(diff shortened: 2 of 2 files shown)


autonomousapps:
Do you use namespaced resources?

rvandermeulen:
No, we don't use resource namespacing. Both reference-browser and Android Components (which provides most of the duplicated R classes here) build with AGP defaults. reference-browser is on AGP 9.2.1, and neither project sets `android.nonTransitiveRClass` or enables any resource-namespacing option, so we're on non-transitive R classes (the AGP 8+ default).

One data point in case it helps: the published AC AARs ship `R.txt` but no compiled `R.class` in `classes.jar`, so the duplicated R classes here are the ones AGP generates on the consuming side.

autonomousapps:
Apparently it is a decade-long project to drag people kicking and screaming into using namespaced resources, which are obviously better in every possible way.

Sorry, just venting, that isn't aimed at you.

I think I would in fact prefer this to be gated on an "is android" check. But it might be easier than you think. The `ProjectVariant` class has a `sourceKind` property. `SourceKind` is a sealed class. You can do an `is` check to see if it's an `AndroidSourceKind` and boom, you've got it.

That'll obviously mean you'll need to change your test, since it's in the `jvm` package. Put it in the `android` package instead, and then I'd strongly prefer if you recreate this same scenario you're describing. Create a dependency from an app module to one of those two dependencies you've mentioned. The test should fail without your change and pass with it.

rvandermeulen:
Glad to hear that since AFAIK we're just doing standard AGP things 😅. Thanks for the feedback, I'll get working on the requested changes ASAP!

rvandermeulen:
Made the gating change: `DiscoverClasspathDuplicationTask` now skips R classes only when `project.sourceKind is AndroidSourceKind`.

On the test, I could use a pointer. I've been trying to recreate the duplication in an `android`-package functional test and haven't managed to, and I think the harness is part of why: `minimalAndroidProperties()` sets `android.nonTransitiveRClass=true`, so the fixtures run non-transitive. A minimal app module depending on `appcompat` (and referencing `androidx.appcompat.R`) doesn't end up with the dependency's R provided by `:app`, so no duplicate is reported, with or without the fix. Forcing `android.nonTransitiveRClass=false` didn't change that, which I assume is because AGP 9 no longer honors transitive R.

In the reported (reference-browser) case, every duplicate is `<dep-package>/R` provided by `[:app, <that-dep>]`, which is the signature of `:app` generating each dependency's R into its own output. Do you have a minimal setup in mind that recreates that under non-transitive R?

Alternatively, would you be open to a focused unit test of the R-skip logic (the `ClasspathAnalyzer`/`isAndroidRClass` filtering) instead of a full functional repro? Happy to go whichever way you prefer.

autonomousapps:
Thanks!

Regarding the test, I would **prefer** a functional test if it's not too onerous to create, and for this situation I _feel_ like it should be doable.

You can overwrite the default gradle properties (sound like you've already discovered how to do that?). `root.gradleProperties` is a `var`, so you can just set it with whatever you want, in your specific test fixture.

And you don't have to abide by the standard minimal app module. You should add a dependency to reference-browser from app.

Regarding AGP 9 not respecting transitive R, you should also adjust the test to only run on AGP 8. That would look like this in the spec:
```groovy
where:
[gradleVersion, agpVersion] << [[GRADLE_LATEST], [AGP_MIN]]
```

All that said, I would be open to a unit test instead, but let's give this one more shot? I appreciate it.

rvandermeulen:
Pushed an update. The gating is in place (skips R only when `sourceKind is AndroidSourceKind`), but I ended up going with a unit test.

On the functional test: I gave it a real go but couldn't get a self-contained fixture to reproduce the duplication, even with transitive R set via `gradleProperties` and pinning to AGP 8 (thanks for those pointers, BTW!). A minimal app depending directly on appcompat doesn't end up with appcompat's R provided by `:app`, and neither does an app depending on a local library. The only thing that reproduces is reference-browser's full transitive graph, and the specific dependencies involved (the Android Components artifacts) aren't on Maven Central, so I couldn't pull them into a DAGP fixture without adding Mozilla's maven repo and pinning an ephemeral nightly version.

The reason seems to be that the duplicate only surfaces when `:app` itself is listed as a provider (e.g. `[:app, androidx.appcompat:appcompat]`), which means the app's own merged R output contains the dependency's R class alongside the dependency's copy. In a minimal one or two module fixture I couldn't get the app's R to include a dependency's R that way, so there was only ever a single provider and nothing to flag. I suspect it's tied to how AGP merges the app R across reference-browser's larger transitive graph, but I wasn't able to pin down the exact trigger. I'm afraid I'm out of ideas for what else to try here, and I'm a bit limited on time to keep iterating on it at this point.

So in the end, I added a focused unit test for the R detection instead (`DiscoverClasspathDuplicationTaskTest`), covering the `R`/`R$*` cases and the lookalikes. If you can think of a simpler functional trigger I'm missing I can try to take another look, otherwise hopefully the unit test will suffice.

autonomousapps (approved):
LGTM. Thanks for trying with the functional test. Unit tests look good 👍

PR #1718, opened 2026-05-29
fix: JarExploder should ignore some MRJar content

ljacomet:
Content of a MRJar that is for a Java version higher than the one running the analysis should be ignored as it should not influence analysis and could cause bundled ASM crashes if the bytecode is too recent.


autonomousapps (approved):
This looks great. Thanks!

Issue #1671, opened 2026-04-08
Gradle 8 support broken since 3.2.0


Vampire:
I've seen that you again raised the Kotlin dependency this time to `2.3.20`, but `gradle-version-min` is still set to `8.11.1`.
As Gradle 8 still only ships with Kotlin 2.0 you can at most use Kotlin 2.1, right?
That's also why you downgraded from 2.2 back to 2.0 in b7121e7b8e14ffe229ba3c3d0c038f2187ef0bd1.

ljacomet:
As an example, this is the error I get when attempting to update to `3.9.0`, on Gradle's `release8x` branch, on top of [this PR](https://github.com/gradle/gradle/pull/37664):

> Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0. 

Build scan: https://ge.gradle.org/s/lqlpo4whryzlo

Vampire:
Ah sorry for not answering, yes, that's the issue.
Same that was there before when then backing down the Kotlin version again.

autonomousapps:
@ljacomet would you consider contributing a change to DAGP to resolve this issue? It's really challenging for me as the sole maintainer to keep this plugin working across major Gradle versions that also involve Kotlin version changes. Note that I have people who want DAGP to use Kotlin 2.3 or even support 2.4 (currently beta).

I'd like to keep the Gradle min version at 8.11 but I would also seriously consider bumping it to 9.0.

autonomousapps:
Case in point https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1662

ljacomet:
From a quick search about `kotlin-metadata` and its usage inside the codebase:
* You already use `KotlinClassMetadata.readLenient` since this plugin has no need to rewrite the metadata.
* Those uses are already inside workers, however they do not use isolation.

I think the path forward is to leverage class loader isolation, to make sure those work actions, which will use a more recent Kotlin version (and its metadata), are isolated.
In short, the main plugin classes need to keep using a Kotlin version compatible with the supported Gradle versions, while the action code, because it runs in isolated workers, should be able to use a more recent version.

I don't think that the running code of the action needs to be compiled with the newer Kotlin version itself, meaning it can stay in the main source set.

I'll see if I can find time to take a stab at this.

autonomousapps:
[...]

PR #1719, opened 2026-05-29
Isolate usage of kotlin-metadata to workers and downgrade Kotlin version used by the plugin

ljacomet:
Turns out that while the [base idea for the fix](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1671#issuecomment-4345964746) was right, the devil is in the details.

I attempted to split the commit logically to help the review:
* First one changes the use of the cache, so that work action can serialize what they need. And then their additions to the cache can be loaded back after being done
* Second one effectively moves the usage of kotlin-metadata to be an additional worker dependency, it no longer is anywhere on the plugin classpath
* Third is about making sure the plugin sticks with Kotlin 2.1, which is the last version that Gradle 8.11 will fully support. It should be updated only when support for that version is abandoned.
* Fourth is a minor version catalog change following a rename in the previous commit

If those changes are too impactful - there is a slight risk of a performance regression with the changes to how caching works - the only alternative is to drop support for older Gradle versions. But this will be a never ending race because of the need for the plugin to understand the latest Kotlin metadata.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -32,9 +32,8 @@ kotlinRuntime = "2.1.21"
 kotlinForAndroidtests = "2.2.21"
 kotlinMetadata = "2.2.21"
 kotlinDokka = "2.2.0"
-# Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
-kotlineditor-core = "0.24"
-kotlineditor-relocated = "0.24.0"
+kotlin-editor-core = "0.24"
+kotlin-editor-relocated = "0.24.0"
 # 0.36.0 requires JDK 17
 mavenPublish = "0.35.0"
 metalava = "1.0.0-alpha14"
@@ -69,9 +68,9 @@ junit-launcher = { module = "org.junit.platform:junit-platform-launcher" }
 junit-params = { module = "org.junit.jupiter:junit-jupiter-params" }
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlinGradlePlugin" }
 kotlinDokkaGradlePlugin = { module = "org.jetbrains.dokka-javadoc:org.jetbrains.dokka-javadoc.gradle.plugin", version.ref = "kotlinDokka" }
-kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlineditor-core" }
-kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlineditor-core" }
-kotlin-editor-relocated = { module = "com.autonomousapps:kotlin-editor-relocated", version.ref = "kotlineditor-relocated" }
+kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlin-editor-core" }
+kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlin-editor-core" }
+kotlin-editor-relocated = { module = "com.autonomousapps:kotlin-editor-relocated", version.ref = "kotlin-editor-relocated" }
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlinGradlePlugin" }
 kotlin-multiplatform-gradle = { module = "org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin", version.ref = "kotlinGradlePlugin" }
 kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlinRuntime" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -22,11 +22,13 @@ javaTarget = "11"
 jdkVersion = "17"
 jspecify = "1.0.0"
 junit = "5.14.1"
-# IMPORTANT: keep this version in sync with build-logic/build.gradle.kts
-# See https://docs.gradle.org/9.4.1/userguide/compatibility.html.
-# Gradle 9.x uses Kotlin language version 2.2
-kotlinLanguageVersion = "2.2"
-kotlin = "2.2.21"
+kotlinGradlePlugin = "2.2.21"
+# Constraint: while Gradle 8.x is supported, every Kotlin artifact on DAGP's plugin classpath must have
+# binary-metadata version <= 2.1.x or Kotlin DSL script compilation breaks. Caps kotlinLanguageVersion (DAGP's classes)
+# and kotlinRuntime (bundled kotlin-stdlib*/kotlin-reflect). kotlinGradlePlugin (compiler/KGP) and kotlinMetadata
+# (isolated workers only) are exempt.
+kotlinLanguageVersion = "2.1"
+kotlinRuntime = "2.1.21"
 kotlinForAndroidtests = "2.2.21"
 kotlinMetadata = "2.2.21"
 kotlinDokka = "2.2.0"
@@ -39,7 +41,8 @@ metalava = "1.0.0-alpha14"
 moshi = "1.15.2"
 moshix = "0.30.0"
 #moshix = "0.31.0"
-okio = "3.16.4"
+# Keep on a release built with Kotlin <= 2.1 (3.15.0 is the latest; 3.16+ breaks Gradle 8.x). See kotlin* note above.
+okio = "3.15.0"
 shadow = "8.3.10"
 spock = "2.3-groovy-4.0"
 truth = "1.4.5"
@@ -64,16 +67,16 @@ junit-bom = { module = "org.junit:junit-bom", version.ref = "junit" }
 junit-engine = { module = "org.junit.jupiter:junit-jupiter-engine" }
 junit-launcher = { module = "org.junit.platform:junit-platform-launcher" }
 junit-params = { module = "org.junit.jupiter:junit-jupiter-params" }
-kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
+kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlinGradlePlugin" }
 kotlinDokkaGradlePlugin = { module = "org.jetbrains.dokka-javadoc:org.jetbrains.dokka-javadoc.gradle.plugin", version.ref = "kotlinDokka" }
 kotlin-editor-core = { module = "app.cash.kotlin-editor:core", version.ref = "kotlineditor-core" }
 kotlin-editor-grammar = { module = "app.cash.kotlin-editor:grammar", version.ref = "kotlineditor-core" }
 kotlin-editor-relocated = { module = "com.autonomousapps:kotlin-editor-relocated", version.ref = "kotlineditor-relocated" }
-kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
-kotlin-multiplatform-gradle = { module = "org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin", version.ref = "kotlin" }
-kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
-kotlin-stdlib-core = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
-kotlin-stdlib-jdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
+kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlinGradlePlugin" }
+kotlin-multiplatform-gradle = { module = "org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin", version.ref = "kotlinGradlePlugin" }
+kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlinRuntime" }
+kotlin-stdlib-core = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlinRuntime" }
+kotlin-stdlib-jdk8 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlinRuntime" }
 kotlin-metadata-jvm = { module = "org.jetbrains.kotlin:kotlin-metadata-jvm", version.ref = "kotlinMetadata" }
 mavenPublishPlugin = { module = "com.vanniktech:gradle-maven-publish-plugin", version.ref = "mavenPublish" }
 metalava = { module = "com.android.tools.metalava:metalava", version.ref = "metalava" }
@@ -94,6 +97,6 @@ buildconfig = { id = "com.github.gmazzo.buildconfig", version.ref = "buildconfig
 dependencyAnalysis = { id = "com.autonomousapps.dependency-analysis", version.ref = "dagp" }
 dokka = { id = "org.jetbrains.dokka", version.ref = "kotlinDokka" }
 gradlePublishPlugin = { id = "com.gradle.plugin-publish", version.ref = "gradle-publish-plugin" }
-kotlinJvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
+kotlinJvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlinGradlePlugin" }
 shadow = { id = "com.gradleup.shadow", version.ref = "shadow" }
 testkit = { id = "com.autonomousapps.testkit", version.ref = "gradleTestKitPlugin" }
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/projects/KotlinDslMetadataVersionProject.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/projects/KotlinDslMetadataVersionProject.groovy
@@ -1,0 +1,98 @@
+// Copyright (c) 2026. Tony Robalik.
+// SPDX-License-Identifier: Apache-2.0
+package com.autonomousapps.jvm.projects
+
+import com.autonomousapps.AbstractProject
+import com.autonomousapps.kit.GradleProject
+import com.autonomousapps.kit.Source
+import com.autonomousapps.kit.SourceType
+import com.autonomousapps.kit.gradle.Dependency
+import com.autonomousapps.kit.gradle.GradleProperties
+import com.autonomousapps.model.ProjectAdvice
+
+import static com.autonomousapps.AdviceHelper.actualProjectAdvice
+import static com.autonomousapps.AdviceHelper.emptyProjectAdviceFor
+
+/**
+ * Regression project for <a href="https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1671">
+ * issue 1671</a>: applying DAGP on Gradle 8.x while compiling <strong>Kotlin DSL</strong> build scripts with the Kotlin
+ * metadata version check <em>enabled</em> ({@code org.gradle.kotlin.dsl.skipMetadataVersionCheck=false}) used to fail
+ * script compilation with:
+ *
+ * <pre>
+ *   Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected
+ *   version is 2.0.0.
+ * </pre>
+ *
+ * because DAGP shipped Kotlin 2.2 metadata on its plugin classpath. This project uses Kotlin DSL build scripts (so the
+ * Kotlin DSL compiler scans the plugin classpath) and enables the check, and exercises the Kotlin-metadata-reading
+ * tasks ({@code FindKotlinMagic*}, {@code AbiAnalysis*}, {@code explodeJar*}) so the classloader-isolated workers that
+ * now carry {@code kotlin-metadata-jvm} are also covered.
+ */
+final class KotlinDslMetadataVersionProject extends AbstractProject {
+
+  final GradleProject gradleProject
+
+  KotlinDslMetadataVersionProject() {
+    this.gradleProject = build()
+  }
+
+  private GradleProject build() {
+    return newGradleProjectBuilder(GradleProject.DslKind.KOTLIN)
+      .withRootProject { r ->
+        // The whole point of the regression: opt IN to the metadata version check (default is to skip it), which is
+        // what trips DAGP's incompatible metadata on Gradle 8.x.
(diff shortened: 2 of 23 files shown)


ljacomet:
Moving back to draft as it looks like there is one [test failure](https://scans.gradle.com/s/tezelhdydvpvk/tests/task/:functionalTest/details/com.autonomousapps.jvm.ClassifiersSpec/transitive%20classifier%20dependencies%20do%20not%20lead%20to%20wrong%20advice%20(Gradle%209.4.1%20DEPENDENCY_WITH_CLASSIFIER_EXISTS)?top-execution=1).

autonomousapps:
Honestly just one test failure with all that is kind of impressive.

ljacomet:
I cannot reproduce the failure in isolation for now, but when running what runs on CI, I got a different error.

My concern is that whatever I did causes some data corruption somewhere ... which I hope is a test artifact rather than a deeper bug.

autonomousapps:
That test that failed on CI also passes locally for me.

Hm, if only Gradle had ever implemented my request for free hosted Develocity for some OSS projects, we could do a build comparison of local vs CI 🤔 

Edit: I wonder if this relates to the Java runtime? CI is pinned to 21 (zulu):
```
- name: Setup java
  uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # 5.2.0
  with:
    distribution: 'zulu'
    java-version: 21
```
whereas locally I am currently using 
```
java -version
openjdk version "23.0.2" 2025-01-21
OpenJDK Runtime Environment Temurin-23.0.2+7 (build 23.0.2+7)
OpenJDK 64-Bit Server VM Temurin-23.0.2+7 (build 23.0.2+7, mixed mode, sharing)
```
I might try using zulu-21 and see if that changes anything.

gradle/libs.versions.toml
```diff
@@ -22,24 +22,27 @@ javaTarget = "11"
...
+kotlinRuntime = "2.1.21"
 kotlinForAndroidtests = "2.2.21"
 kotlinMetadata = "2.2.21"
 kotlinDokka = "2.2.0"
-# Cannot be called kotlin-editor as it causes `libs.versions.kotlin.get()` to fail
-kotlineditor-core = "0.20"
-kotlineditor-relocated = "0.20.0"
+# Keep on a release built with Kotlin <= 2.1 (0.18 is the latest; 0.20 breaks Gradle 8.x). See kotlin* note above.
+kotlin-editor-core = "0.18"
+kotlin-editor-relocated = "0.18.0"
```
autonomousapps: This will probably break some tests. Why isn't the version-forcing mechanism from build-logic handling this issue?

gradle/libs.versions.toml
```diff
@@ -22,24 +22,27 @@ javaTarget = "11"
...
 # 0.36.0 requires JDK 17
 mavenPublish = "0.35.0"
 metalava = "1.0.0-alpha14"
 moshi = "1.15.2"
 moshix = "0.30.0"
 #moshix = "0.31.0"
-okio = "3.16.4"
+# Keep on a release built with Kotlin <= 2.1 (3.15.0 is the latest; 3.16+ breaks Gradle 8.x). See kotlin* note above.
+okio = "3.15.0"
```
autonomousapps, line 44: This downgrade is probably fine, but I have the same question about the version-forcing mechanism.

src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
```diff
@@ -11,18 +11,20 @@ import com.autonomousapps.model.internal.PhysicalArtifact.Mode
...
 import com.autonomousapps.model.internal.intermediates.producer.ExplodedJar
-import com.autonomousapps.services.InMemoryCache
 import com.autonomousapps.tasks.ExplodeJarTask
 import java.util.zip.ZipFile
 
 internal class JarExploder(
   private val artifacts: List<PhysicalArtifact>,
   private val androidLinters: Set<AndroidLinterDependency>,
-  private val inMemoryCache: InMemoryCache
+  private val seedCache: Map<String, ExplodedJar>,
```
autonomousapps, line 20: Note to self: check where this comes from.

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+internal object KotlinMetadataClasspath {
+
+  private const val CONFIGURATION_NAME = "dependencyAnalysisKotlinMetadata"
+
+  /** Returns the (idempotently created) resolvable configuration as a [FileCollection] suitable for a task input. */
+  fun of(project: Project): FileCollection {
+    project.configurations.findByName(CONFIGURATION_NAME)?.let { return it }
+
+    return project.configurations.create(CONFIGURATION_NAME) { c ->
```
autonomousapps: I would prefer we use lazy APIs if possible. `named` and `register`. There is a file, `com.autonomousapps.artifacts.configurations.kt`, that has some helpers for this case.

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
```diff
@@ -0,0 +1,44 @@
...
+      c.isCanBeResolved = true
+      c.isCanBeConsumed = false
```
autonomousapps: Probably want `isVisible = false` for pre-Gradle 9? The helpers I mention above handle this.

src/main/kotlin/com/autonomousapps/services/InMemoryCache.kt
autonomousapps, line 1: I actually really appreciate this. The whole exploding/exploded has been annoying me for a while. Too many intermediate model classes. Just haven't made time to simplify.

src/main/kotlin/com/autonomousapps/tasks/ExplodeJarTask.kt
```diff
@@ -53,36 +62,65 @@ public abstract class ExplodeJarTask @Inject constructor(
   public abstract val output: RegularFileProperty
 
   @TaskAction public fun action() {
-    workerExecutor.noIsolation().submit(ExplodeJarWorkAction::class.java) {
-      it.inMemoryCache.set(inMemoryCache)
+    // Pass the shared cache content to the work action, which requires serializable data only
```
autonomousapps, line 65: ah ok, that explains it. Thanks for the comment. Although it is kind of a confusing that a shared build service can't be used directly in a gradle-managed object like a worker action. Why is that?

src/main/kotlin/com/autonomousapps/tasks/ExplodeJarTask.kt
```diff
@@ -53,36 +62,65 @@ public abstract class ExplodeJarTask @Inject constructor(
...
         artifacts = parameters.physicalArtifacts.fromJsonList(),
         androidLinters = parameters.androidLinters.fromNullableJsonSet<AndroidLinterDependency>(),
-        inMemoryCache = parameters.inMemoryCache.get()
-      ).explodedJars()
+        seedCache = parameters.cacheSeed.fromJsonMap(),
+      )
+      val explodedJars = exploder.explodedJars()
 
       output.bufferWriteJsonSet(explodedJars, compress = true)
+      parameters.newCacheEntries.getAndDelete().bufferWriteJsonMap(exploder.newEntries)
```
autonomousapps: I would prefer the `getAndDelete()` part to happen at the top of the action. The point of that is to ensure that if something happens that causes the action to fail/crash, then there aren't stale/wrong outputs from a prior run.

src/main/kotlin/com/autonomousapps/tasks/FindKotlinMagicTask.kt
```diff
@@ -106,6 +137,8 @@ public abstract class FindKotlinMagicTask @Inject constructor(
       inlineUsageReportFile.bufferWriteJsonSet(inlineMembers)
       typealiasReportFile.bufferWriteJsonSet(typealiases)
 
+      parameters.newCacheEntries.getAndDelete().bufferWriteJsonMap(finder.newEntries)
```
autonomousapps: same comment

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
```diff
@@ -0,0 +1,44 @@
...
+      c.description = "kotlin-metadata-jvm, supplied to classloader-isolated DAGP workers."
+      c.defaultDependencies { deps ->
+        val dep = project.dependencies.create(
+          "org.jetbrains.kotlin:kotlin-metadata-jvm:${BuildConfig.KOTLIN_METADATA_VERSION}"
```
autonomousapps: I think I'm missing something. Can you explain how this can resolve to a higher version that the consumer might require? Does that happen automatically or would they have to explicitly configure via the `dependencyAnalysisKotlinMetadata` configuration?

autonomousapps (changes requested):
Thanks for this. Left some comments. Not yet sure why the test is failing.

autonomousapps:
I installed zulu-21 and ran the tests again, and they still all succeeded:
```
$ sdk i java 21.0.11-zulu

$ java -version
openjdk version "21.0.11" 2026-04-21 LTS
OpenJDK Runtime Environment Zulu21.50+19-CA (build 21.0.11+10-LTS)
OpenJDK 64-Bit Server VM Zulu21.50+19-CA (build 21.0.11+10-LTS, mixed mode, sharing)

$ ./gradlew :fT --tests "ClassifiersSpec.transitive classifier dependencies*"
```
https://scans.gradle.com/s/3zgahtrg4spn4

autonomousapps:
@ljacomet I've [added](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/commit/0220175ef10382367cf73dbd5c8d0bf3a5d04456) temporary uploading of the functional test build artifacts. With that, I can compare the build artifacts locally to see what's different.

If you rebase your PR onto that and let it fail again, let me know.

src/main/kotlin/com/autonomousapps/tasks/ExplodeJarTask.kt
ljacomet: Because the moment you need isolation (classloader or process), Gradle requires all parameters to be serializable. And `BuildService` are not, they have too much ties into Gradle internals.

src/main/kotlin/com/autonomousapps/tasks/ExplodeJarTask.kt
ljacomet: Done

src/main/kotlin/com/autonomousapps/tasks/FindKotlinMagicTask.kt
ljacomet: Done

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
ljacomet: Done

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
ljacomet: Done

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
ljacomet: This probably needs also documentation.

The idea is that without upgrading this plugin, a project that bumps its Kotlin version to a newer one could also update this dependency and get analysis working - assuming the `kotlin-metadata-jvm` API is stable enough. But from my quick search, it should be.

ljacomet:
I've addressed some comments, but not all.

However this is rebased, let's see what happens with tests.

autonomousapps:
Looks like the same failure as before: https://scans.gradle.com/s/obkd2ypbdmt74

Here's the artifact download URL: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/actions/runs/26915202104/artifacts/7397914732

It appears to have worked exactly as desired. That zip contains some directories, `TransitiveClassifierTestProject-*` (one for each combination in the spec).

You can optionally simplify the failing spec to run just the combination of things that is failing, then run:
```shell
./gradlew :fT --tests "ClassifiersSpec.transitive classifier dependencies do not lead to wrong advice*"
```
And then you'll see output in `build/functionalTest/TransitiveClassiferTestProject-*`. You can either compare the outputs directly with whatever tool you like (I often use `diff --brief --recursive` for this), or just zip up that output and send it to me, and I can take a look since I'm more used to inspecting this output.

ianbrandt:
I'm not sure how you all feel about AI-assisted contributions, but I had some Claude tokens to burn. If interested, it reproduced the test failure and came up with a fix: https://github.com/ljacomet/dependency-analysis-gradle-plugin/pull/1.

ljacomet:
Thanks @ianbrandt, I've merged your PR on my branch.

I still need to answer questions around the dependency version-forcing mechanism.

ianbrandt:
Thanks for all your work on this issue, @ljacomet. I'm glad the PR was helpful.

gradle/libs.versions.toml
ljacomet: Regarding this comment and the other one about version-forcing, [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1719#discussion_r3329604328), you are right, it works in the sense that the plugin will use and depend on the Kotlin version defined in the catalog.

But my understanding is that this could still be an issue if the plugin ships with dependencies that use a Kotlin version that Gradle 8.11 will not be able to run. [Gradle 8.11 bundles Kotlin 2.0.20](https://docs.gradle.org/current/userguide/compatibility.html#kotlin). 

If I understand [Kotlin binary compatibility](https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format) properly, anything built on Kotlin `2.2+` (assuming it does not target a lower language level) is a potential problem.

After all, this is the reason behind moving `kotlin-metadata` to a worker with class loader isolation.

So maybe both `kotlin-editor` and `okio` can stay in the versions they had before this PR, as I tried running Gradle 8.11 tests and could not see them fail. But the risk is that a later upgrade, without raising the plugin's minimum support Gradle version could re-introduce a similar issue.

I have to admit I do not understand why `okio` `3.16.4` does not causes that issue [since version `3.16.0` bumped Kotlin to `2.2.0`](https://square.github.io/okio/changelog/#version-3160).

Let me know how you would want to proceed.

rvandermeulen:
@ljacomet Just curious if you've had a chance to look at this again recently?

ianbrandt:
@rvandermeulen, It's easy to miss, but I believe the last status change is that Louis has an open question threaded above: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1719#discussion_r3391830943.

gradle/libs.versions.toml
autonomousapps: Let's take a step back. This plugin has a functional testing suite that (nominally) runs against Gradle 8.11. Why isn't that showing any issues? If we can get the tests to _fail_ in the scenario that led to this PR, then I'd have confidence with making changes relating to these dependencies.

ianbrandt: I had some Fable 5 usage to burn. 🤖

> The suite passes on 8.11.1 because the failure is opt-in. Per @3flex's [comment on #1671](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1671#issuecomment-4446096661), the Kotlin DSL script compiler skips the metadata version check unless `org.gradle.kotlin.dsl.skipMetadataVersionCheck=false` (or warnings-as-errors, which is how the gradle/gradle build hit it).
> 
> This PR adds `KotlinDslMetadataVersionSpec` in d244cec6, which enables that property in a Kotlin DSL fixture. Cherry-picked onto current `main`, it fails on Gradle 8.11.1 with "Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0", flagging the plugin jar itself plus graph-support, variant-artifacts, kotlin-metadata-jvm 2.2.21, okio-jvm 3.16.4, and kotlin-editor-relocated 0.23.0. On this branch it passes on both 8.11.1 and latest. That also answers the earlier okio question: the compiler checks every classpath jar's module metadata whether or not the script references it, so the okio and kotlin-editor downgrades are required.
> 
> (To verify the green leg from a direct checkout of this branch, cherry-pick 11743268 first; the branch predates that fix, and the functional-test publish otherwise fails on signing. CI is unaffected since it builds the PR merged with main.)

autonomousapps: I don't mind so much pinning okio (though I'm not happy about it), but the lowered version of kotlin-editor isn't acceptable. That library is what lets `fixDependencies` work for Kotlin DSL scripts, which is an area of active development. If keeping it at a higher level isn't breaking anything, then I'm fine rolling the dice on that. Maybe a follow-up PR could do something similar for kotlin-editor as is being done here for kotlin-metadata.

ljacomet: Given the information above, I create a simple project on which I applied the DAG plugin.
* Latest released version causes the build to fail if `org.gradle.kotlin.dsl.skipMetadataVersionCheck=false` is set in `gradle.properties`
* The version produced by this PR no longer causes the build to fail, with that property set.
* The version produced by this PR, with either `okio` or `kotlin-editor` back to their original version, causes the build to fail as well with the property set.

So effectively, in order to get full Gradle `[8.11,9.0[` support, `okio` must be pinned to a Kotlin 2.0.0 compatible version and so should `kotlin-editor` until it follows the same handling as here: move to worker execution with classloader isolation.

My recommendation: This PR goes in (if approved) with the `okio` pinning as that one will be required for ever AFAICT. But I undo the pinning of `kotlin-editor`. Then a later PR can attempt to move script rewrite to a worker action, effectively removing that dependency from the buildscript classpath. In the meantime, Gradle 8.x compatibility is only valid if metadata version check is turned off.
Alternatives: Neither `okio` nor `kotlin-editor` are pinned in this PR, or both are.

@autonomousapps The final decision is yours to make

ljacomet:
Hey,

I've added [a comment](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1719#discussion_r3564422083) based on the additional information.

Let me know how I should proceed. Once a decision is made, I can rebase and mark it ready for review again in case a positive outcome is agreed upon.

gradle/libs.versions.toml
autonomousapps: @ljacomet I agree with your recommendation. Let's pin okio, unpin kotin-editor, and merge this PR. And then a future PR can move kotlin-editor to isolated worker actions. I just checked the source code and the usage of kotlin-editor classes is fairly limited. It looks like `RewriteTask` currently isn't using the Worker API, but I don't see a reason it can't.

I'll also add `org.gradle.kotlin.dsl.skipMetadataVersionCheck=false` to all test fixtures.

ljacomet:
I have rebased this work on top of the latest state, I had to reintroduce some of the methods removed in #1776 as I was using them.

As agreed in [this comment](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1719#discussion_r3564749525), `okio` is now pinned to a Kotlin metadata 2.1 compatible version. However `kotlin-editor` still uses a more recent Kotlin version and might still cause issues in Gradle [8.11, 9.0[. But this will have to be handled in a separate pull request where its usage is also moved to an isolated worker.

I'll follow up if there are CI failures, but otherwise this should reflect what I intended to do initially.

ljacomet:
Note that I had to disable caching in `InMemoryCacheSpec` as it otherwise gets cache hits preventing the `doLast` execution.

src/functionalTest/groovy/com/autonomousapps/jvm/projects/KotlinDslMetadataVersionProject.groovy
```diff
@@ -0,0 +1,98 @@
...
+  KotlinDslMetadataVersionProject() {
+    this.gradleProject = build()
+  }
+
+  private GradleProject build() {
+    return newGradleProjectBuilder(GradleProject.DslKind.KOTLIN)
+      .withRootProject { r ->
+        // The whole point of the regression: opt IN to the metadata version check (default is to skip it), which is
+        // what trips DAGP's incompatible metadata on Gradle 8.x.
+        r.gradleProperties += GradleProperties.of('org.gradle.kotlin.dsl.skipMetadataVersionCheck=false')
```
trobalik-duolingo, line 45: Note to self: I want to enable this property for ALL tests once we have fixed the remaining issues (i.e. with kotlin-editor).

src/main/kotlin/com/autonomousapps/internal/KotlinMetadataClasspath.kt
autonomousapps: I did finally check for myself and the solution is for consumers to add this to their normal build scripts (not meddle with their _buildscript_ classpaths)
```kotlin
// build.gradle.kts
dependencies {
  "dependencyAnalysisKotlinMetadata"("org.jetbrains.kotlin:kotlin-metadata-jvm:2.4.0")
}
```
So that definitely needs to be documented. I also wonder if we can expose something on the `dependencyAnslysis` DSL. I might do one or both.

Also worth noting this needs to be added to _all modules_. So this PR unfortunately harms DX for users of this plugin. I think we need to consider ways to make this more automatic.

autonomousapps:
If anyone tracking this PR has ideas about [this](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1719#discussion_r3632691140), I'd like to hear them. I actually don't think I can merge this PR if Kotlin users are forced to add an explicit kotlin-metadata-jvm dependency in all of their build scripts.

Making it a DSL option helps because then users could configure things in just one place, but even having to configure **one thing** is a bad solution, since the failure mode is pretty obscure. I am also aware that having to configure one thing is already the situation, but I was hoping this PR would eliminate the need for this workaround:

```kotlin
// settings.gradle.kts
buildscript {
  // DAGP 3.14.1 bundles kotlin-metadata-jvm 2.2.21, which only understands @Metadata
  // up to Kotlin 2.3.0. Our classes are compiled with 2.4.0, so force the matching
  // metadata parser onto the settings/plugin classpath.
  configurations.classpath {
    resolutionStrategy { force("org.jetbrains.kotlin:kotlin-metadata-jvm:2.4.0") }
  }
}
```

Is there a way that the plugin can easily and quickly (at configuration time) infer which version of the `kotlin-metadata-jvm` artifact is needed for analysis so that it can automatically add that as project dependencies?

**Edit:** maybe it would be good enough to just check which version of Gradle is in use?

**Second edit**: I'm experimenting and this seems to be working.

autonomousapps:
While debugging this issue, I noticed that Kotlin's `KotlinModuleMetadata.read(bytes)` function doesn't have a `...lenient()` version, in contrast to the `KotlinClassMetadata.readLenient()` function. Inside its `read()` function, is this:
```kotlin
throwIfNotCompatible(it, lenient = false /* TODO */)
```
which is all to say, this issue would never occur if that TODO had been resolved. Lenient reads don't trigger the error that led to this PR. Certain kinds of reads are unfortunately only available in strict form.

autonomousapps (approved):
I think this needs documentation, but the commit I've pushed that automatically tracks the Gradle runtime for adding the right Kotlin metadata artifact has resolved my major concern.

PR #1721, opened 2026-05-30
Draft: fix: Use runtime classpath for ServiceLoader dependency advice.

Freeezzzi:
## Problem

`FindServiceLoadersTask` scanned the compile classpath for `META-INF/services/*` files.

This missed cases where ServiceLoader providers are only visible from runtime artifacts. In particular, project dependencies can appear on the compile classpath as class directories without resources, while their runtime jar contains the actual service provider file.

This meant a dependency like:

```groovy
dependencies {
  implementation(project(":service-loader"))
}
```
could be reported as an implementation dependency, even though its only relevant capability is a runtime ServiceLoader provider and it should be moved to:
```groovy
dependencies {
  runtimeOnly(project(":service-loader"))
}
```
It also meant runtime-only transitive providers could be missed when removing an otherwise unused dependency:
```
dependencies {
  implementation(project(":unused"))
}
```
where `:unused` brings in a runtime ServiceLoader dependency such as:
```groovy
runtimeOnly("com.mysql:mysql-connector-j:8.0.33")
```
## How I propose to fix
`FindServiceLoadersTask` now scans runtime classpath artifacts instead of compile classpath artifacts.

`ComputeUsagesTask` also consumes the ServiceLoader report directly and adds runtime ServiceLoader coordinates to the dependency metadata it visits. This makes runtime-only ServiceLoader providers visible to usage computation even when they are absent from the compile graph.

The legacy `setCompileClasspath()` and `getCompileClasspath()` methods are retained as deprecated delegating accessors for compatibility.


src/functionalTest/groovy/com/autonomousapps/jvm/projects/ProjectServiceLoaderProject.groovy
autonomousapps, line 1: What's the meaning of the name, "_Project_ ServiceLoaderProject"?

src/functionalTest/groovy/com/autonomousapps/jvm/RuntimeOnlySpec.groovy
autonomousapps, line 1: Did this spec fail before your other changes?

src/main/kotlin/com/autonomousapps/tasks/FindServiceLoadersTask.kt
```diff
@@ -43,17 +45,25 @@ public abstract class FindServiceLoadersTask : DefaultTask() {
...
-    //  compileClasspath contains a directory that includes only class files. It doesn't not contain any resources
-    //  files, which is where the service loader definition would be.
-    val serviceLoaders = compileClasspath
-      .filterNonGradle()
+    // Module jars are scanned from the compile classpath to preserve existing advice behavior. Project dependencies may
+    // appear there as resource-less class directories, so scan their runtime jars as well.
+    val serviceLoaderArtifacts = compileClasspath.filterNonGradle() +
+      runtimeClasspath.filterNonGradle().filter { it.id.componentIdentifier is ProjectComponentIdentifier }
+
+    val serviceLoaders = serviceLoaderArtifacts
```
autonomousapps, line 66: I'm not sure about this for a couple of reasons.
1. I think we could actually just stop scanning the compile classpath.
2. The map/filter could be more fluent
3. I don't understand why we'd only want to look at local projects and not _anything_ on the runtime classpath.
4. I don't think that comment is accurate.

src/main/kotlin/com/autonomousapps/tasks/FindServiceLoadersTask.kt
```diff
@@ -64,39 +74,39 @@ public abstract class FindServiceLoadersTask : DefaultTask() {
...
-        val providerClasses = zip.getInputStream(serviceFile)
-          .bufferedReader().use(BufferedReader::readLines).asSequence()
-          // remove whitespace
-          .map(String::trim)
-          // remove blank lines
-          .filterNot(String::isBlank)
-          // ignore comments
-          .filter { !it.startsWith("#") }
-          .toSortedSet()
+    return ZipFile(artifact.file).use { zip ->
```
autonomousapps, line 77: Good catch 👍

src/test/kotlin/com/autonomousapps/tasks/FindServiceLoadersTaskTest.kt
autonomousapps, line 1: We can delete this. I don't think we want to retain the "legacy" behavior.

autonomousapps:
@Freeezzzi please assert that you created this PR without the assistance of any LLM coding assistant. Note that the code of conduct for this repo forbids contributions that used LLMs in any way.

Issue #1725, opened 2026-06-08
`:aggregatePublicTypes` task has incorrect task metadata


jordanterry:
In the task's `init` block the task's group is set to a description: 

https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/b253f1f51b3a27ac98cf8713d1700ed169315659/src/main/kotlin/com/autonomousapps/tasks/AggregatePublicTypesTask.kt#L26-L28

A sibling task has the following definition: 

https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/b253f1f51b3a27ac98cf8713d1700ed169315659/src/main/kotlin/com/autonomousapps/tasks/AbiAnalysisTask.kt#L23-L31

### `:tasks` output

```
Writes report of all public classes across all variants or source sets in this project tasks
--------------------------------------------------------------------------------------------
aggregatePublicTypes
```

autonomousapps:
Whoops! Thank you for the report.

jordanterry:
Completed in: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1726

PR #1726, opened 2026-06-08
Update metadata of `:aggregatePublicTypesTask` to move `description` to `group`

jordanterry:
Tackles: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1725

### DIFF
--- a/src/main/kotlin/com/autonomousapps/tasks/AggregatePublicTypesTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/AggregatePublicTypesTask.kt
@@ -25,6 +25,7 @@ public abstract class AggregatePublicTypesTask @Inject constructor(
 
   init {
     group = "Writes report of all public classes across all variants or source sets in this project"
+    description = "Writes report of all public classes across all variants or source sets in this project"
   }
 
   @get:Input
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/autonomousapps/tasks/AggregatePublicTypesTask.kt
```diff
@@ -24,7 +25,8 @@ public abstract class AggregatePublicTypesTask @Inject constructor(
 ) : DefaultTask() {
 
   init {
-    group = "Writes report of all public classes across all variants or source sets in this project"
+    group = TASK_GROUP_DEP
```
jordanterry: todo @jordanterry: Double check this is necessary (I haven't dived into the source of all other tasks yet)

autonomousapps (changes requested):
Thanks for moving the group to the description, but this task shouldn't be in a group at all. It's internal and it just clutters output for something that isn't user-facing. Thanks.

jordanterry:
@autonomousapps thanks! I really wasn't sure about that from my quick look this morning. Will update shortly!

src/main/kotlin/com/autonomousapps/tasks/AggregatePublicTypesTask.kt
jordanterry: https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1726#pullrequestreview-4451229688

autonomousapps (changes requested):
Oh haha, you should _delete_ the `group`. Otherwise this PR looks good.

jordanterry:
@autonomousapps oh my, I dropped the commit but didn't think about what that does! 🤦

autonomousapps (approved):
LGTM!

PR #1737, opened 2026-06-18
Escape reserved keywords when using project accessors

### LINKED ISSUE #1715: Modules with names equivalent to Kotlin keywords should be escaped with `fixDependencies` task
**Plugin version**

3.13.0

**Gradle version**

9.4.1

**JDK version**

17

**Describe the bug**

When `./gradlew fixDependencies` updates Gradle files, modules whose name happens to be the same as a Kotlin keyword is unescaped, so in a Kotlin Gradle script, a compile time error occurs. This only happens when Gradle's typesafe project accessors are enabled.

**To Reproduce**

1. Checkout https://github.com/veyndan/dependency-analysis-gradle-plugin-playground/tree/1715.
2. Run `./gradlew fixDependencies`.
3. The root `build.gradle.kts` file is updated with `api(projects.interface)`.

**Expected behavior**

In step (3) from above, the `build.gradle.kts` file is updated with ``` api(projects.`interface`) ```.

veyndan:
Fixes #1715.

I kept both commits just to show a train of thought. Initially, I thought that every keyword in Kotlin needs to be escaped, so that's what my implementation in the first commit does. When running the checks before making a PR, the `use typesafe project accessors syntax when dsl is kotlin and useTypesafeProjectAccessors is true is true` test was failing, because it references a coordinate like `:sad-robot:internal`, but interestingly the conversion to `projects.sadRobot.internal` is truly, actually valid.

In the second commit, I'm doing the simpler thing of just listing some known keywords that need to be escaped. I don't know what makes something needing escape or not. I tried a few random keywords, and `interface` and `class` aren't fine but `internal`, `abstract`, `public` are fine.

The downside with the second approach is that this list is most likely incomplete. At least if there are further issues, it's trivial to update.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -27,9 +27,10 @@ internal class AdvicePrinter(
       }
     }
 
-    val kotlinReservedKeywords = List(KotlinParser.VOCABULARY.maxTokenType) { KotlinParser.VOCABULARY.getLiteralName(it) }
-      .filterNotNull()
-      .map { it.removeSurrounding("'") }
+    val kotlinReservedKeywords = listOf(
+      "class",
+      "interface",
+    )
   }
 
   fun line(configuration: String, printableIdentifier: String, was: String = ""): String {
--- a/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.autonomousapps.internal.advice
 
+import com.autonomousapps.internal.squareup.cash.grammar.KotlinParser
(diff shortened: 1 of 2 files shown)


autonomousapps:
@veyndan I think it must have to do with "hard" vs "soft" keywords. See [here](https://kotlinlang.org/docs/keyword-reference.html). If you want to update this PR so that all of the "hard" keywords are covered, that would make sense I think.

src/main/kotlin/com/autonomousapps/internal/advice/AdvicePrinter.kt
```diff
@@ -25,6 +26,11 @@ internal class AdvicePrinter(
         it.value.last().uppercaseChar().toString()
       }
     }
+
+    val kotlinReservedKeywords = listOf(
+      "class",
+      "interface",
+    )
```
autonomousapps: could you move this up with the other `val`?

autonomousapps (commented):
This looks good. LMK if you want to update the list to include more hard keywords.

veyndan:
That's it! I'd totally forgot about hard vs soft keywords. I just updated the PR with your suggestions.

autonomousapps (approved):
LGTM

PR #1768, opened 2026-07-10
feat!: add DSL option for binary-compatibility check. It now defaults to 'false'.

autonomousapps:
Doing this reduces memory usage when *not* opting-in to the binary-compatibility check.


src/main/kotlin/com/autonomousapps/extension/UsageAnalysisHandler.kt
```diff
@@ -12,6 +12,9 @@ import javax.inject.Inject
  * dependencyAnalysis {
  *   usage {
  *     analysis {
+ *       // When true, does binary compatibility analysis, which can be very expensive in memory terms.
+ *       checkBinaryCompatibility(true) // defaults to false
```
autonomousapps, line 16: Defaulting to false is a breaking (behavioral) change, needs to be clearly documented.

PR #1772, opened 2026-07-14
refactor: make all json outputs compressable (false by default).

autonomousapps:
A follow-up will flip the default to compressed by default.


autonomousapps:
Need to revisit this later. Having the file extensions change to be json or json.gz depending on whether they're compressed (and making compression configurable) is a much bigger change than I anticipated.

PR #1779, opened 2026-07-16
fix: make GraphViewTask sensitive to new transitive project edges

### LINKED ISSUE #1775: Plugin can reuse a stale graph and suppress dependency advice when an upstream project dependency changes
**Plugin version**
3.16.1

**Gradle version**
9.6.0

**JDK version**
21

**Kotlin and Kotlin Gradle Plugin (KGP) version**
2.3.21

**Describe the bug**
I’m seeing a failure where an application depends on a library, and that library adds an API dependency on another project. The application begins using a type from that new transitive project, but its own dependency declarations remain unchanged.

I expect the plugin to advise the application to declare that dependency directly with `implementation project(':transitive')`, but instead, GraphViewTask can restore the graph produced before the new project edge existed. The stale graph-compile.json omits the new project, and I get empty dependency advice which then causes `buildHealth` to fail later when someone busts the cache.

**To Reproduce**

On a fork I added two failing tests modeled on the existing GraphViewCacheSpec:

- [`graphViewTask is sensitive to new transitive project edges`](https://github.com/andrewparmet/dependency-analysis-gradle-plugin/blob/d2227e74/src/functionalTest/groovy/com/autonomousapps/jvm/GraphViewProjectEdgeCacheSpec.groovy)
- [`stale project graph does not suppress dependency advice`](https://github.com/andrewparmet/dependency-analysis-gradle-plugin/blob/d2227e74/src/functionalTest/groovy/com/autonomousapps/jvm/GraphViewProjectEdgeCacheSpec.groovy)

Both tests fail against `main` and I threw together a [fix on top](https://github.com/andrewparmet/dependency-analysis-gradle-plugin/commit/e86db335d192cfcba5dc7ddec71ecd5c03489566) as an MVP - not sure what direction you might want to go.

**Expected behavior**

Adding transitive usage like this should produce advice to declare the used transitive.

**Additional context**

 I found a few past issues and changes that look related:

  - #612 describes stale graph output after an external dependency version changes. #618 addressed that case by changing how the compile classpath is tracked.

  - #728 added dependency declarations as an input, but it does not appear to cover changes to the resolved project graph when the current project’s declarations stay the same.

  - #1220 also involves cached graph data, although that issue concerns included builds and appears to be a different failure mode.

This case differs because only an upstream project’s dependencies change.

Thanks!

andrewparmet:
Fixes #1775.

Adds regression tests demonstrating stale GraphViewTask outputs when an upstream project dependency changes. Fixes the issue by tracking resolved component IDs as task inputs, ensuring those changes invalidate the task’s build cache entry.

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/GraphViewProjectEdgeCacheSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/GraphViewProjectEdgeCacheSpec.groovy
@@ -1,0 +1,63 @@
+// Copyright (c) 2026. Tony Robalik.
+// SPDX-License-Identifier: Apache-2.0
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.GraphViewProjectEdgeCacheProject
+import com.autonomousapps.model.Advice
+import org.gradle.util.GradleVersion
+
+import static com.autonomousapps.AdviceHelper.projectCoordinates
+import static com.autonomousapps.kit.truth.BuildTaskSubject.buildTasks
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertAbout
+import static com.google.common.truth.Truth.assertThat
+
+final class GraphViewProjectEdgeCacheSpec extends AbstractJvmSpec {
+
+  def "graphViewTask is sensitive to new transitive project edges"() {
+    given:
+    def project = new GraphViewProjectEdgeCacheProject()
+    gradleProject = project.gradleProject
+    def task = ':consumer:graphViewMain'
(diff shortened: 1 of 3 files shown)


autonomousapps:
Thanks for the PR. Either I made a mistake (very possible) or there's something else blocking me from pushing to your branch. I've created a fresh PR at https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1780 with a commit that fixes some issues.

autonomousapps:
I've merged the other PR (which include your two original commits, with my third on top), so I'm going to close this one. Thanks for the contribution!

andrewparmet:
It’s possible I forgot to check a box that allows you to commit. In any case, thanks!

PR #1780, opened 2026-07-16
fix: make GraphViewTask sensitive to new transitive project edges.

autonomousapps:
Fixup of https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1779.


src/main/kotlin/com/autonomousapps/tasks/GraphViewTask.kt
```diff
@@ -152,4 +164,22 @@ public abstract class GraphViewTask : DefaultTask() {
     outputRuntime.bufferWriteJson(runtimeGraphView)
     outputRuntimeDot.writeText(graphWriter.toDot(runtimeGraph))
   }
+
+  private fun ResolvedComponentResult.allComponentIds(): Set<String> {
+    val visited = mutableSetOf<String>()
```
autonomousapps, line 169: I've made this a `Set<String>` instead of a `Set<ResolvedComponentResult>`. AFAIK, `ResolvedComponentResult` has no real `hashCode()` function, and so has no legitimate set semantics.

PR #1790, opened 2026-07-26
fix: invalidate artifact reports when resolved versions change

### LINKED ISSUE #1426: False positive unused dependency on debugImplementation for Compose ui-tooling
**Plugin version**
2.16.0

**Gradle version**
8.13

**JDK version**
23

**Kotlin and Kotlin Gradle Plugin (KGP) version**
2.1.20

**Android Gradle Plugin (AGP) version**
8.9.2

**Describe the bug**
All of my UI modules have a `debugImplementation` on `androidx.compose.ui:ui-tooling` which was fine until one of the versions of DAGP released in the past few weeks, when I started noticing that after some time I got false positives for each UI module saying:

```
    debugImplementation(libs.compose.ui.tooling)
```

Running `buildHealth` with `--no-configuration-cache --no-build-cache --rerun-tasks` fixes the problem temporarily. I've added that artifact to my excludes for now.

I'm trying to repro in a shareable project, and will update if I can.

mvanhorn:
Add a normalized, order-independent input to `ArtifactsReportTask` that captures the identities of the resolved artifacts whose coordinates are serialized, including both ordinary and opaque artifact collections where applicable. `ArtifactsReportTask` writes resolved coordinates into each `PhysicalArtifact`, but its declared cache inputs cover artifact files, build path, and exclusions rather than the resolved artifact identities used to produce that output.

Happy path: build health with BOM version A, clean, then build with BOM version B whose selected artifact bytes are identical; both compile and runtime artifact-report tasks execute with the new identities, their reports contain version B, and no false unused advice is emitted; Edge cases: exercise the fixture across the repository's supported Gradle-version matrix and verify both ordinary compile/runtime artifacts and any opaque artifact input remain deterministic and order-independent.

Fixes #1426

### DIFF
--- a/src/functionalTest/groovy/com/autonomousapps/jvm/ArtifactsReportCacheSpec.groovy
+++ b/src/functionalTest/groovy/com/autonomousapps/jvm/ArtifactsReportCacheSpec.groovy
@@ -1,0 +1,58 @@
+// Copyright (c) 2026. Tony Robalik.
+// SPDX-License-Identifier: Apache-2.0
+package com.autonomousapps.jvm
+
+import com.autonomousapps.jvm.projects.ArtifactsReportCacheProject
+
+import static com.autonomousapps.kit.truth.BuildTaskSubject.buildTasks
+import static com.autonomousapps.utils.Runner.build
+import static com.google.common.truth.Truth.assertAbout
+import static com.google.common.truth.Truth.assertThat
+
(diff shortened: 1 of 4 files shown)


autonomousapps:
I have a simpler alternative fix for this. I would also like to refer you to the [code of conduct](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CODE_OF_CONDUCT.md), which explicitly forbids LLM contributions.

PR #1793, opened 2026-07-27
fix: GraphViewTask's inputs are inputs, not internal details.

autonomousapps:
This fixes an issue with the build cache that was causing issues with 'CustomSourceSetSpec'.


src/main/kotlin/com/autonomousapps/tasks/GraphViewTask.kt
autonomousapps, line 1: The fix.

PR #1797, opened 2026-07-28
Update to AGP 9.3.1

MGaetan89:
I've added AGP 9.3.1 as the latest stable version, and bumped the other versions to the latest bug fix version.


autonomousapps (approved):
Thanks!

PR #1800, opened 2026-07-28
refactor: add ExpensiveJar type to hold BinaryClasses, removing them from ExplodedJar.

autonomousapps:


src/main/kotlin/com/autonomousapps/internal/JarExploder.kt
```diff
@@ -4,39 +4,50 @@ package com.autonomousapps.internal
...
-      }
-      .toExplodedJars()
+    return expensiveJars.mapToOrderedSet { it.explodedJar }
   }
 
-  private fun Sequence<PhysicalArtifact>.toExplodedJars(): Set<ExplodedJar> =
+  private fun Sequence<PhysicalArtifact>.toExpensiveJars(): Set<ExpensiveJar> =
     map { artifact ->
-      val key = artifact.file.absolutePath
+      val key = artifact.coordinates.toString()
```
autonomousapps: Unrelated change: reduce usage of `PhysicalArtifact.file`, which is an absolute file.

src/main/kotlin/com/autonomousapps/model/internal/intermediates/producer/ExplodedJar.kt
autonomousapps, line 1: The point of the PR: remove `binaryClasses` from `ExplodedJar`, eliminating a potential footgun when dealing with this class.

ianbrandt:
I think the exploded-jar cache never hits with this change. `ExplodeJarTask` seeds it keyed by `artifact.file.absolutePath`, but `JarExploder.toExpensiveJars()` now looks up `seedCache[artifact.coordinates.toString()]`. The write-back path then stores new entries under the coordinates key, so neither side finds the other.

Running `buildHealth` on six projects that share one classpath, sequentially so the cache has a chance to fill:

```
3.18.0                      this PR
lib1  seed=2     new=14MB   lib1  seed=2  new=14MB
lib2  seed=14MB  new=2      lib2  seed=2  new=14MB
lib3  seed=14MB  new=2      lib3  seed=2  new=14MB
```

Every project re-explodes the full classpath. Changing that one key back to `artifact.file.absolutePath` restores the hits, and the build stays green. The KDoc on both sides still says "keyed by artifact path", so I believe that's what was intended.

autonomousapps:
> I think the exploded-jar cache never hits with this change. `ExplodeJarTask` seeds it keyed by `artifact.file.absolutePath`, but `JarExploder.toExpensiveJars()` now looks up `seedCache[artifact.coordinates.toString()]`. The write-back path then stores new entries under the coordinates key, so neither side finds the other.
> 
> Running `buildHealth` on six projects that share one classpath, sequentially so the cache has a chance to fill:
> 
> ```
> 3.18.0                      this PR
> lib1  seed=2     new=14MB   lib1  seed=2  new=14MB
> lib2  seed=14MB  new=2      lib2  seed=2  new=14MB
> lib3  seed=14MB  new=2      lib3  seed=2  new=14MB
> ```
> 
> Every project re-explodes the full classpath. Changing that one key back to `artifact.file.absolutePath` restores the hits, and the build stays green. The KDoc on both sides still says "keyed by artifact path", so I believe that's what was intended.

Thanks for taking a look. I missed the artifact.file usage in the task class, and updated that as well as KotlinMagicFinder to use the coordinates string instead. And I updated the KDoc.

I am trying to limit usage of the absolute file, since it infects lots of things with its absolute path, breaking build cache relocatability. I'm trying to slowly get rid of it over time.

PR updated.

ianbrandt:
The cache does hit now, but I think the coordinates key introduces a correctness problem. Two different jars can share one key, and whichever project runs second gets the first one's analysis.

`Coordinates.toString()` doesn't carry the classifier, and `GradleVariantIdentification` only carries capabilities, since `toGradleVariantIdentification()` sets `attributes = emptyMap()`. So `com.example:probe:1.0` and `com.example:probe:1.0:linux` are one key.

Minimal repro, where `probe-1.0.jar` contains only `com.example.Alpha` and `probe-1.0-linux.jar` only `com.example.Beta`:

```
// lib1
implementation("com.example:probe:1.0")        // doesn't use it
// lib2
implementation("com.example:probe:1.0:linux")  // calls com.example.Beta.greet()
```

On 3.18.0, lib2's `exploded-jars.json` reports `com.example.Beta` and lib2 gets no advice. On this branch it reports `com.example.Alpha`, and `buildHealth` emits:

```
Advice for :lib2
  implementation("com.example:probe:1.0")
```

**lib2 does use that dependency, so applying that advice breaks compilation.** The same happens for `implementation(files("libs/local.jar"))` in two projects, since `FlatCoordinates` is just the file name.

The analysis is derived from file content, so I think the key has to identify content. Perhaps the file name alongside the coordinates, or a content hash.

Two smaller things. `seedCache[key]?.copy(coordinates = artifact.coordinates)` can no longer change anything now that the key is the coordinates, and the comment above it still explains the path rationale. And this key never reaches the build cache, since it only lives in the BuildService and two `temporaryDir` files, so relocatability doesn't have to constrain the choice here.

autonomousapps:
What I'll do is revert that change so the rest of the PR can move forward.

My point about cache relocatability has to do with the existence of that `PhysicalArtifact.file` property in the first place. It's an absolute path, and if there's some way I can fully remove it from this project, I want to do that.

Your comment exposes a need for an additional regression test. I don't like that CI is green but apparently has this gap. I'll see about adding such a test and reproducing the failure to ensure a regression isn't introduced later.

Part of the gap must have to do with Gradle's lack of a public API for determining an artifact's classifier. Of course Gradle itself encourages the use of variants over classifiers, but Maven still exists and we just have to accept that. Anyway, missing Gradle support means the `Coordinates` class doesn't model the classifier at all, and I think that's the root of this problem. There's a TODO somewhere in the codebase about this.

Tangential to the main point, but I find the `seedCache[key]?.copy(coordinates = artifact.coordinates)` code annoying and a bandaid over the problem.

autonomousapps:
[Unit test](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1803) to prevent the issue identified [here](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/pull/1800#issuecomment-5112124960).

PR #1812, opened 2026-08-11
fix: handle mixed source sets better.

### LINKED ISSUE #1811: Mixed Java/Kotlin test fixtures are attributed to the main dependency
### Contributor Checklist
- [x] I have read the [Contributing guide](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CONTRIBUTING.asciidoc).
- [x] I have read the
  [Code of Conduct](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CODE_OF_CONDUCT.md).
- [x] No part of this bug report was created with an LLM/AI.

### Plugin version
3.18.0

### Gradle version
9.7.0

### JDK version
17

### (Optional) Kotlin and Kotlin Gradle Plugin (KGP) version
2.4.10

### (Optional) Android Gradle Plugin (AGP) version
9.3.1

### Describe the bug

I have a project with mixed Kotlin and Java test fixtures across JVM/Android modules. First I get weird incorrect advice, and with a workaround I get a crash.

### To Reproduce

I created a minimal reproducer here: https://github.com/andrewparmet/dagp-reproducers.

I'm seeing one issue and then my workaround exposes another issue.

First, if you run `./gradlew :consumer:projectHealth`, it succeeds, but `projectHealth` reports:

```
  testImplementation(project(":producer"))
```

Which the consumer doesn't reference - it uses only the test fixtures. It looks like this is because the classes root dir is widened to `build/classes`.

To work around that, I can relocate the test fixtures output: 

```
kotlin {
    sourceSets.named("testFixtures") {
        kotlin.destinationDirectory.set(
            layout.buildDirectory.dir("testFixtures-classes/kotlin")
        )
    }
}

sourceSets.named("testFixtures") {
    java.destinationDirectory.set(
        layout.buildDirectory.dir("testFixtures-classes/java")
    )
}
```

In my reproducer, run Gradle with `-PrelocateTestFixtures=true`: `./gradlew clean :consumer:projectHealth -PrelocateTestFixtures=true`

Then `projectHealth` fails with `Collection contains no element matching the predicate.`, which I thought was caused within JarExploder's dependency coordinate map colliding on different directories corresponding to the same dependency coordinate.

Thanks!

autonomousapps:
Resolve https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1811.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/internal/artifactViews.kt
+++ b/src/main/kotlin/com/autonomousapps/internal/artifactViews.kt
@@ -7,12 +7,13 @@ import org.gradle.api.artifacts.ArtifactView
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier
 import org.gradle.api.artifacts.result.ResolvedDependencyResult
+import org.gradle.api.artifacts.type.ArtifactTypeDefinition
 import org.gradle.api.attributes.Attribute
 import org.gradle.api.attributes.Category
 import org.gradle.internal.component.local.model.OpaqueComponentArtifactIdentifier
 
 /**
- * This is different than [org.gradle.api.attributes.Category.CATEGORY_ATTRIBUTE], which has type
+ * This is different from [org.gradle.api.attributes.Category.CATEGORY_ATTRIBUTE], which has type
  * `Category` (cf `String`).
  */
 internal val CATEGORY = Attribute.of("org.gradle.category", String::class.java)
--- a/src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
+++ b/src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
@@ -4,6 +4,7 @@
 
 package com.autonomousapps.tasks
 
+import com.autonomousapps.internal.ArtifactsExpander
 import com.autonomousapps.internal.utils.bufferWriteJsonSet
 import com.autonomousapps.internal.utils.filterNonGradle
 import com.autonomousapps.internal.utils.getAndDelete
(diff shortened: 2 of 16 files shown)


src/main/kotlin/com/autonomousapps/internal/artifactViews.kt
```diff
@@ -51,7 +52,7 @@ internal fun ResolvedDependencyResult.isJavaPlatform(): Boolean = selected.varia
 }
 
 internal object ArtifactAttributes {
-  const val ANDROID_CLASSES_JAR = "android-classes"
+  const val ANDROID_CLASSES_JAR = ArtifactTypeDefinition.JAR_TYPE//"android-classes"
```
autonomousapps: The proposed fix.

src/main/kotlin/com/autonomousapps/tasks/ArtifactsReportTask.kt
```diff
@@ -127,6 +127,9 @@ public abstract class ArtifactsReportTask : DefaultTask() {
       .filterNonGradle()
       .mapNotNull {
         try {
+          // TODO(tsr): this is broken, but I'm also not sure if it's hit anymore, since moving to consuming Android
+          //  "JAR" artifacts instead of "android-classes". Actually, `JarExploder.explode()` has a note about this. We
+          //  need to test this same problem against java-library and kotlin-jvm producers and consumers.
```
autonomousapps: TODO: 
- [x] Codify the test from [the reproducer](https://github.com/andrewparmet/dagp-reproducers)
- [ ] Add test for java-library
- [ ] Add test for kotlin-library (maybe the same test)

src/main/kotlin/com/autonomousapps/internal/artifactViews.kt
autonomousapps: Sadly, this doesn't actually work. Leads to some bizarre outcomes. Will have to find another solution that continues to use the `android-classes` artifact type.

PR #1839, opened 2026-08-21
chore: build with Gradle 9.7.1 and KGP 2.4.0.

autonomousapps:


autonomousapps:
Hmm, in `AnnotationsImplementationSpec#class-retained annotations used by delegates do not need to be declared`, the class file for `MyService.kt` has changed from KGP 2.2.21 to 2.4.0.

In the `RuntimeVisibleAnnotations` section of the file, the `d2` array contains a new element, `"Lcom/google/errorprone/annotations/CanIgnoreReturnValue;"`. (There are some other differences in the class file, too.) For this reason, the `@CanIgnoreReturnValue` annotation has been upgraded from `Uses (as or in an annotation)` to `Uses 1 class` (according to `reason`). I think that happens due to this in `asm.kt`:

```kotlin
private fun addClass(className: String?, enclosingAnnotation: String? = null) {
  classes.addClass(className, ClassRef.Kind.ANNOTATION, enclosingAnnotation)

  if (arrayName == "d2") {
    // Use "Kind.NOT_ANNOTATION" always so that our "is this typealias used?" algorithm works.
    arrayElements.addClass(className, ClassRef.Kind.NOT_ANNOTATION)
  }
}
```
I'm not entirely clear on why things end up in that `d2` array. Here's the full contents:
```
d2=["Lcom/example/consumer/MyService;","Lcom/google/common/util/concurrent/Service;","delegate","<init>","(Lcom/google/common/util/concurrent/Service;)V","addListener","","listener","Lcom/google/common/util/concurrent/Service$Listener;","kotlin.jvm.PlatformType","executor","Ljava/util/concurrent/Executor;","awaitRunning","timeout","","unit","Ljava/util/concurrent/TimeUnit;","awaitTerminated","failureCause","","isRunning","","startAsync","Lcom/google/errorprone/annotations/CanIgnoreReturnValue;","state","Lcom/google/common/util/concurrent/Service$State;","stopAsync","the-project:consumer"]
```
vs the *before* case:
```
d2=["Lcom/example/consumer/MyService;","Lcom/google/common/util/concurrent/Service;","delegate","<init>","(Lcom/google/common/util/concurrent/Service;)V","addListener","","listener","Lcom/google/common/util/concurrent/Service$Listener;","kotlin.jvm.PlatformType","executor","Ljava/util/concurrent/Executor;","awaitRunning","timeout","","unit","Ljava/util/concurrent/TimeUnit;","awaitTerminated","failureCause","","isRunning","","startAsync","state","Lcom/google/common/util/concurrent/Service$State;","stopAsync","consumer"]
```
Or, probably easier to see in this diff view, with before on the left and after on the right:
Screenshot 2026-08-21 at 6 01 24 PM

Issue #1828, opened 2026-08-20
buildHealth fails for needed compile time dependency


Mahoney:
### Contributor Checklist
- [x] I have read the [Contributing guide](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CONTRIBUTING.asciidoc).
- [x] I have read the [Code of Conduct](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CODE_OF_CONDUCT.md).
- [x] No part of this bug report was created with an LLM/AI.

### (Optional) Build scan link
https://scans.gradle.com/s/4rf5mqnnbarqa

### Plugin version
3.18.0

### Gradle version
9.7.0

### JDK version
26

### (Optional) Kotlin and Kotlin Gradle Plugin (KGP) version
2.4.10

### (Optional) Android Gradle Plugin (AGP) version
N/A

### (Optional) `reason` output for bugs relating to incorrect advice
```
$ ./gradlew :reason --id org.apache.kafka:kafka-server
Calculating task graph as no cached configuration is available for tasks: :reason --id org.apache.kafka:kafka-server

> Task :reason

------------------------------------------------------------
You asked about the dependency 'org.apache.kafka:kafka-server:4.3.1'.
You have been advised to remove this dependency from 'implementation'.
------------------------------------------------------------

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for grpcOtherCompileClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for grpcOtherRuntimeClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for jmhCompileClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for jmhRuntimeClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for compileClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

Shortest path from root project to org.apache.kafka:kafka-server:4.3.1 for runtimeClasspath:
:
\--- org.apache.kafka:kafka-server:4.3.1

[...]

PR #1855, opened 2026-08-26
docs: clarify that checkSuperClasses() is only for source code, not for external deps.

autonomousapps:
Resolves (as works-as-intended) https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1828

As the new KDOC in `ComputeUsagesTask` says:
```
  /**
   * This checks if [coordinates] provides a symbol for a superclass missing from THIS module's source code. It does NOT
   * check if [coordinates] (or any dependency) has superclasses missing from the compile classpath.
   */
```

So, the issue in question is not for a supported use-case.

### DIFF
--- a/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/artifact/JarSubject.kt
+++ b/testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/artifact/JarSubject.kt
@@ -36,7 +36,7 @@ public class JarSubject private constructor(
     val actual = assertNonNull(actual) { "jar was null" }
 
     // Open zip, copy entry to temp dir, close zip.
-    val tempResource = FileSystems.newFileSystem(actual, null).use { fs ->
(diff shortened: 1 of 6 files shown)


testkit/gradle-testkit-truth/src/main/kotlin/com/autonomousapps/kit/truth/artifact/JarSubject.kt
```diff
@@ -36,7 +36,7 @@ public class JarSubject private constructor(
     val actual = assertNonNull(actual) { "jar was null" }
 
     // Open zip, copy entry to temp dir, close zip.
-    val tempResource = FileSystems.newFileSystem(actual, null).use { fs ->
+    val tempResource = FileSystems.newFileSystem(actual, null as ClassLoader?).use { fs ->
```
autonomousapps, line 39: Java 17 has multiple `newFileSystem(path, ??)` overloads, so this clarifies for if/when I move the build target from 11 to 17. Discovered by accident due to IDE shenanigans. (I can't move to JDK 17 till I change the min supported Gradle fro 8.11 to 9+).

PR #1869, opened 2026-09-03
fix: don't crash when analyzing KMP project with 'jvm("foo")' target.

### LINKED ISSUE #1865: Dependency analysis crashes with a JVM target named desktop
### Contributor Checklist
- [x] I have read the [Contributing guide](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CONTRIBUTING.asciidoc).
- [x] I have read the [Code of Conduct](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CODE_OF_CONDUCT.md).
- [x] No part of this bug report was created with an LLM/AI.

### Plugin version

3.19.1

### Gradle version

9.7.0

### JDK version

25

### (Optional) Kotlin and Kotlin Gradle Plugin (KGP) version

2.4.10

### (Optional) Android Gradle Plugin (AGP) version

9.4.0

### Describe the bug

I have a KMP project with `jvm("desktop")`. If I declare `com.ibm.icu:icu4j:77.1` (ICU4J) in the corresponding `desktopMainImplementation` and used by desktop code, the `computeAdvice` task fails with:

```
java.lang.IllegalArgumentException: Change advice for com.ibm.icu:icu4j cannot be from and to the same configuration (desktopMainImplementation in this case)
    at com.autonomousapps.model.Advice$Companion.ofChange(Advice.kt:53)
    at com.autonomousapps.internal.transform.KmpTransform.simplify(KmpTransform.kt:299)
    at com.autonomousapps.internal.transform.KmpTransform.reduce(KmpTransform.kt:112)
```

I think the cause is inconsistent classpath names when constructing `KmpSourceKind`. The recorded usage has:

```json
{
  "name": "desktopMain",
  "kind": "CUSTOM_JVM",
  "compileClasspathName": "desktopCompileClasspath",
  "runtimeClasspathName": "desktopRuntimeClasspath"
}
```

However, `KmpSourceKind.of("desktopMain")` constructs `desktopMainCompileClasspath` and `desktopMainRuntimeClasspath`. Only jvmMain and androidMain receive special handling (basically - the `Main` suffix). Maybe there's just some inconsistency here due to the special casing of the default names?

### To Reproduce

1. Create an empty Gradle project with this build file

```kotlin
plugins {
    kotlin("multiplatform") version "2.4.10"
    id("com.autonomousapps.dependency-analysis") version "3.19.1"
}

// ...

kotlin {
    jvm("desktop")
    jvmToolchain(25)

    sourceSets {
        named("desktopMain") {
            dependencies {
                implementation("com.ibm.icu:icu4j:77.1")
            }
        }
    }
}
```

2. Use something from ICU4J in there, like this

```kotlin
package example

import com.ibm.icu.text.ListFormatter

// ICU is used internally
fun formatItems(items: List<String>): String =
    ListFormatter.getInstance().format(items)
```

3. Run `./gradlew computeAdvice`

4. Observe the failure

### Expected behavior

I’d expect analysis to complete and keep the dependency in its current configuration.

autonomousapps:
Resolves https://github.com/autonomousapps/dependency-analysis-gradle-plugin/issues/1865.

### DIFF
--- a/src/main/kotlin/com/autonomousapps/model/source/SourceKind.kt
+++ b/src/main/kotlin/com/autonomousapps/model/source/SourceKind.kt
@@ -318,20 +318,46 @@ public data class KmpSourceKind(
         compileClasspathName = when (sourceSetName) {
           // jvmMain => jvmCompileClasspath (jvmMain is special)
           JVM_MAIN_NAME -> "jvmCompileClasspath"
+
           // androidMain => androidCompileClasspath (androidMain is special)
           ANDROID_MAIN_NAME -> "androidCompileClasspath"
-          // jvmTest => jvmTestCompileClasspath
-          else -> "${sourceSetName}CompileClasspath"
+
+          else -> {
+            if (isProbablyJvmMain(sourceSetName)) {
+              // desktopMaim => desktopCompileClasspath
+              "${sourceSetName.substringBefore("Main")}CompileClasspath"
+            } else {
+              // jvmTest => jvmTestCompileClasspath
+              "${sourceSetName}CompileClasspath"
+            }
+          }
         },
         runtimeClasspathName = when (sourceSetName) {
           // jvmMain => jvmRuntimeClasspath (jvmMain is special)
           JVM_MAIN_NAME -> "jvmRuntimeClasspath"
+
           // androidMain => androidRuntimeClasspath (androidMain is special)
           ANDROID_MAIN_NAME -> "androidRuntimeClasspath"
-          // jvmTest => jvmTestRuntimeClasspath
-          else -> "${sourceSetName}RuntimeClasspath"
+
+          else -> {
+            if (isProbablyJvmMain(sourceSetName)) {
+              // desktopMaim => desktopRuntimeClasspath
+              "${sourceSetName.substringBefore("Main")}RuntimeClasspath"
+            } else {
+              // jvmTest => jvmTestRuntimeClasspath
+              "${sourceSetName}RuntimeClasspath"
+            }
+          }
         },
       )
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/autonomousapps/model/source/SourceKind.kt
```diff
@@ -318,20 +318,46 @@ public data class KmpSourceKind(
...
         },
       )
     }
+
+    /**
+     * Returns true if [sourceSetName] is probably from `jvm("sourceSetName")`. Explicitly ignores the `commonMain`
+     * source set because that's special.
+     */
+    private fun isProbablyJvmMain(sourceSetName: String): Boolean {
+      return sourceSetName != COMMON_MAIN_NAME && sourceSetName.endsWith("Main")
```
ZacSweers, line 360: I think this would match all non-commonmain source sets?

autonomousapps: non-commonMain and endsWith("Main")

ZacSweers: Yes. `js()` ends with `jsMain` and would match this no?

autonomousapps: Yes. I feel like you're being too Socratic. What's your point? Note that DAGP doesn't support js yet, so obviously there are no js-specific regression tests that would show whether this succeeds or fails in such a situation (and also, it doesn't really matter since: no support for js). Are you saying that js classpaths have a different naming convention? If so, what are they?

PR #1890, opened 2026-09-14
feat(testkit-support): remoeved deprecated factory methods from AndroidBlock. test: removed deprecated usages from functional tests.

autonomousapps:


src/functionalTest/groovy/com/autonomousapps/android/projects/BundleProject.groovy
```diff
@@ -24,13 +24,13 @@ final class BundleProject extends AbstractAndroidProject {
   }
 
   private GradleProject build() {
-    return newAndroidGradleProjectBuilder(agpVersion)
+    return newAndroidGradleProjectBuilder()
       .withAndroidSubproject('lib') { a ->
         a.sources = sources
-        a.manifest = libraryManifest()
+        a.manifest = null
```
autonomousapps, line 30: Need to explicitly set to null anywhere that I'm using `withAndroidSubproject()` instead of `withAndroidLibProject()`.

